PR #122, opened 2023-05-10
Refactor out cipher components from aead

skmcgrail:
```
Removed items from the public API
=================================
-pub const aws_lc_rs::aead::AES_128_KEY_LEN: usize = 16usize
-pub const aws_lc_rs::aead::AES_256_KEY_LEN: usize = 32usize
-pub const aws_lc_rs::aead::BLOCK_LEN: usize = 16usize

Changed items in the public API
===============================
(none)

Added items to the public API
=============================
(none)
```

the ISC license.


aws-lc-rs/src/aead.rs
```diff
@@ -72,21 +72,17 @@ use key_inner::KeyInner;
...
 pub mod chacha20_poly1305_openssh;
-mod cipher;
 mod key_inner;
 mod nonce;
 pub mod nonce_sequence;
 mod poly1305;
 pub mod quic;
 
 pub use self::{
-    aes_gcm::{AES_128_GCM, AES_128_KEY_LEN, AES_256_GCM, AES_256_KEY_LEN},
```
justsmth, line 88: There is value to exposing `AES_128_KEY_LEN` and `AES_256_KEY_LEN` and possibly a `AES_BLOCK_LEN` to consumers, but they shouldn't be exposed from the `aead` module.

justsmth: We could potentially move these constants to `cipher::aes`.

justsmth (approved):

PR #129, opened 2023-05-18
add support and tests for ECDH P-521

samuel40791765:
### Issues:
Addresses `V907517296`

### Description of changes: 
This extends our existing ECDH implementation to support P-521. I've also extended the existing test file to include P-521 test vectors from the NIST files specified in the comments.
I've also added a function in the `SecureRandom` trait to indicate if the rng is used for testing. This is needed to differentiate if we're setting a key from our KATs or if we need to generate a random value.

### Call-outs:
P-521 is not a Montgomery curve, so we can't directly generate a random key and assign it to the EC_KEY. Directly assigning the private key a value will induce an "INVALID_SCALER" error from AWS-LC. We use `EC_KEY_generate_key(_fips)` to delegate the random key generation from AWS-LC instead. 

**Overall question: Should we delegate the random key generation to AWS-LC for all curves?**

### Testing:
KATs from NIST. New modified tests like the existing modified tests from ring.

the ISC license.


aws-lc-rs/src/agreement.rs
```diff
@@ -125,6 +129,11 @@ pub static ECDH_P384: Algorithm = Algorithm {
     id: AlgorithmID::ECDH_P384,
 };
 
+/// ECDH using the NSA Suite B P-521 (secp384r1) curve.
```
dkostic, line 132:
```suggestion
/// ECDH using the NSA Suite B P-521 (secp521r1) curve.
```

aws-lc-rs/src/agreement.rs
```diff
@@ -218,6 +230,21 @@ impl EphemeralPrivateKey {
                 rng.fill(&mut priv_key)?;
                 Self::from_p384_private_key(&priv_key)
             }
+            AlgorithmID::ECDH_P521 => {
+                let mut priv_key = [0u8; ECDH_P521_PRIVATE_KEY_LEN];
+                // If the rng is only used for testing purposes, we set the hard-coded value. This is the case
+                // for our KATs. Otherwise, we delegate the random key generation to AWS-LC.
+                //
+                // P-521 is not a Montgomery curve, so we can't directly generate a random key and assign it to
```
dkostic, line 238: This is not the issue. The problem is that you are generating a private key that's too large. The order of the P-521 group `n` is a number that's 521-bits long and the private key has to be smaller than `n`. By generating 66 random bytes you are generating a random 528-bit number. With very high probability this number will be larger than `n`, in which case `EC_KEY_set_private_key` will fail. Exactly the same applies to the other curves, but for P-521 it happens with much larger probability.

I see two proper ways to handle this:
1. Relegate the key generation to AWS-LC for all curves.
2. Generate a valid random private key for all curves (`0 < priv_key < n`).

aws-lc-rs/src/agreement.rs
```diff
@@ -218,6 +230,14 @@ impl EphemeralPrivateKey {
                 rng.fill(&mut priv_key)?;
                 Self::from_p384_private_key(&priv_key)
             }
+            AlgorithmID::ECDH_P521 => {
+                let mut priv_key = [0u8; ECDH_P521_PRIVATE_KEY_LEN];
+                // The private key is stored in 66 bytes = 528-bits, which is too large.
+                rng.fill(&mut priv_key)?;
+                // We mask the top 7-bits of the first byte, so at most 521-bits are set.
+                priv_key[0] &= 0x01;
```
dkostic, line 238: I guess this means the key bytes are stored in big endian form?

aws-lc-rs/src/agreement.rs
```diff
@@ -218,6 +230,14 @@ impl EphemeralPrivateKey {
                 rng.fill(&mut priv_key)?;
                 Self::from_p384_private_key(&priv_key)
             }
+            AlgorithmID::ECDH_P521 => {
+                let mut priv_key = [0u8; ECDH_P521_PRIVATE_KEY_LEN];
+                // The private key is stored in 66 bytes = 528-bits, which is too large.
```
dkostic, line 235: ok, this change makes the P-521 case the same as the others, but the general problem is still there (for all curves). For example, `priv_key` with 256 ones is not a valid P-256 private key because it's larger than the order of P-256, same for other curves.

Is this function used only for testing?

samuel40791765: This function is used for both production and testing.
The difference between testing and production is the type of `struct` of the `rng` passed in. In production, `rng.fill` generates a random number. In our tests, `rng.fill` is configured to pass in a known value for our KATs.
Your concern with the `priv_key` generation for all the other curves is valid though, which is why I think we should consider using `EC_KEY_generate_key` for all curves.

aws-lc-rs/src/agreement.rs
```diff
@@ -137,14 +151,18 @@ pub static X25519: Algorithm = Algorithm {
     id: AlgorithmID::X25519,
 };
 const X25519_PRIVATE_KEY_LEN: usize = aws_lc::X25519_PRIVATE_KEY_LEN as usize;
+#[cfg(test)]
 const ECDH_P256_PRIVATE_KEY_LEN: usize = 32;
+#[cfg(test)]
 const ECDH_P384_PRIVATE_KEY_LEN: usize = 48;
+const ECDH_P521_PRIVATE_KEY_LEN: usize = 66;
```
dkostic, line 158: I don't even know what `#[cfg(test)]` means but should we add it above the P521 line as well? :)

aws-lc-rs/src/ec.rs
```diff
@@ -259,7 +264,25 @@ pub(crate) unsafe fn ec_key_from_private(
...
+#[inline]
+pub(crate) unsafe fn ec_key_generate(
+    ec_group: &ConstPointer<EC_GROUP>,
+) -> Result<DetachableLcPtr<*mut EC_KEY>, Unspecified> {
+    let ec_key = DetachableLcPtr::new(EC_KEY_new())?;
+    if 1 != EC_KEY_set_group(*ec_key, **ec_group) {
+        return Err(Unspecified);
+    }
 
+    if 1 != aws_lc::EC_KEY_generate_key(*ec_key) {
```
dkostic, line 283: should we have FIPS and non-FIPS call here? There is `EC_KEY_generate_key_fips` in AWS-LC.

justsmth: Yes. Good catch.

justsmth (dismissed):

samuel40791765 (commented):
PR looks good to me if we're going this route.
One thing we should note here is that we're ignoring the `SecureRandom` parameter once this is done. If there were users importing external keys and setting them with the original `test::rand::FixedSliceRandom` API that was available, this would break them.
But this isn't a documented API in ring and was only intended for testing, so..

justsmth (approved):

aws-lc-rs/src/agreement.rs
dkostic: interesting, thanks for explaining

aws-lc-rs/src/ec.rs
```diff
@@ -259,7 +264,29 @@ pub(crate) unsafe fn ec_key_from_private(
...
+}
 
+#[inline]
+pub(crate) unsafe fn ec_key_generate(
+    ec_group: &ConstPointer<EC_GROUP>,
+) -> Result<DetachableLcPtr<*mut EC_KEY>, Unspecified> {
+    let ec_key = DetachableLcPtr::new(EC_KEY_new())?;
+    if 1 != EC_KEY_set_group(*ec_key, **ec_group) {
+        return Err(Unspecified);
+    }
```
dkostic, line 281: I suggest using `EC_KEY_new_by_curve_name` instead of `EC_KEY_new()` + `EC_KEY_set_group`.

justsmth: Fixed in the next commit.

dkostic (approved):

PR #141, opened 2023-06-06
Minor cipher refactoring; Add integ tests

justsmth:
### Issues:
N/A

### Description of changes: 
* Renamed `DecryptionContext` to `CipherContext`
* Moved `SymmetricCipherKey` to a `key` sub-module.
* `impl AsRef<[u8]> for CipherContext`
* `#[derive(Debug, PartialEq, Eq, Clone, Copy)]` on simple enums.
* Add integ tests.

### Call-outs:
N/A

### Testing:
N/A

the ISC license.

### DIFF
--- a/aws-lc-rs/src/cipher.rs
+++ b/aws-lc-rs/src/cipher.rs
@@ -180,17 +180,6 @@ impl<'a> TryFrom<&'a CipherContext> for &'a [u8] {
     }
 }
 
-impl<'a> TryFrom<&'a mut CipherContext> for &'a mut [u8] {
-    type Error = Unspecified;
-
-    fn try_from(value: &'a mut CipherContext) -> Result<Self, Unspecified> {
-        match value {
-            CipherContext::Iv128(iv) => Ok(iv.as_mut()),
-            CipherContext::None => Err(Unspecified),
-        }
-    }
-}
-
 #[non_exhaustive]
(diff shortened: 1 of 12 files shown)


aws-lc-rs/src/cipher.rs
```diff
@@ -208,15 +161,27 @@ pub enum OperatingMode {
...
-    Iv128(iv::FixedLength<16>),
+    Iv128(FixedLength<16>),
 
     /// No input to the cipher mode.
     None,
 }
 
+const EMPTY_BYTE_ARRAY: [u8; 0] = [];
+
+impl AsRef<[u8]> for CipherContext {
```
skmcgrail, line 174: My only concern with this is if there is a mode whose context can't be defined from a single `&[u8]`. For example are there modes where it would make sense to model this as a tuple of input values etc.

If we want to support such a potential use case in the future then implementing `AsRef<[u8]>` might be problematic.

skmcgrail: Suggestion:
```rust
impl TryFrom<CipherContext> for &[u8] {
    type Error;

    fn try_from(value: CipherContext) -> Result<Self, Self::Error> {
        todo!()
    }
}
```

skmcgrail (approved):

PR #150, opened 2023-06-08
Add Support for CBC, CTR Cipher modes with AES 128 & 256 bit keys.

skmcgrail:
### Description of changes: 
This pull request adds support for CBC and CTR cipher mode encryption for 128-bit and 256-bit AES keys.

* Introduces `aws_lc_rs::cipher` module which supports unauthenticated ciphers for encryption/decryption.
* `PaddedBlockEncryptingKey`, `PaddedBlockDecryptingKey` provide encryption/decryption of plaintext/ciphertext for block cipher modes using padding. `PaddedBlockEncryptingKey::cbc_pkcs7(...)` `PaddedBlockDecryptingKey::cbc_pkcs7(...)` and are the constructor methods provided for using CBC mode with PKCS#7 padding. 
  * Users who need to explicitly set an initialization vector can use `PaddedBlockEncryptingKey::less_safe_cbc_pkcs7`.
* `EncryptingKey`, `DecryptingKey` provide encryption/decryption of plaintext/ciphertext for unpadded block cipher modes. `PaddedBlockEncryptingKey::ctr(...)` `PaddedBlockDecryptingKey::ctr(...)` and are the constructor methods provided for using CTR mode (unpadded). 
  * Users who need to explicitly set an initialization vector can use `EncryptingKey::less_safe_ctr`.

#### AES-128 CBC
```rust
use aws_lc_rs::cipher::{
    PaddedBlockDecryptingKey, PaddedBlockEncryptingKey, UnboundCipherKey, AES_128,
};

let original_message = "This is a secret message!".as_bytes();
let mut in_out_buffer = Vec::from(original_message);

let key_bytes: &[u8] = &[
    0xff, 0x0b, 0xe5, 0x84, 0x64, 0x0b, 0x00, 0xc8, 0x90, 0x7a, 0x4b, 0xbf, 0x82, 0x7c, 0xb6,
    0xd1,
];

let key = UnboundCipherKey::new(&AES_128, key_bytes)?;
let encrypting_key = PaddedBlockEncryptingKey::cbc_pkcs7(key)?;
let context = encrypting_key.encrypt(&mut in_out_buffer)?;

let key = UnboundCipherKey::new(&AES_128, key_bytes)?;
let decrypting_key = PaddedBlockDecryptingKey::cbc_pkcs7(key)?;
let plaintext = decrypting_key.decrypt(&mut in_out_buffer, context)?;
assert_eq!(original_message, plaintext);
```

#### AES-128 CTR
```rust
use aws_lc_rs::cipher::{DecryptingKey, EncryptingKey, UnboundCipherKey, AES_128};

let original_message = "This is a secret message!".as_bytes();
let mut in_out_buffer = Vec::from(original_message);

let key_bytes: &[u8] = &[
    0xff, 0x0b, 0xe5, 0x84, 0x64, 0x0b, 0x00, 0xc8, 0x90, 0x7a, 0x4b, 0xbf, 0x82, 0x7c, 0xb6,
    0xd1,
];

let key = UnboundCipherKey::new(&AES_128, key_bytes)?;
let encrypting_key = EncryptingKey::ctr(key)?;
let context = encrypting_key.encrypt(&mut in_out_buffer)?;

let key = UnboundCipherKey::new(&AES_128, key_bytes)?;
let decrypting_key = DecryptingKey::ctr(key)?;
let plaintext = decrypting_key.decrypt(&mut in_out_buffer, context)?;
assert_eq!(original_message, plaintext);
```


samuel40791765 (commented):
First pass; I'm a bit curious what the benchmark numbers look like against `openssl_bench` :)

aws-lc-rs/src/cipher.rs
```diff
@@ -3,158 +3,1268 @@
...
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> Result<(), core::fmt::Error> {
+        f.debug_struct("UnboundCipherKey")
+            .field("algorithm", &self.algorithm)
+            .finish()
+    }
+}
+
+impl From<hkdf::Okm<'_, &'static Algorithm>> for UnboundCipherKey {
+    fn from(okm: hkdf::Okm<&'static Algorithm>) -> Self {
+        let mut key_bytes = [0; AES_256_KEY_LEN];
```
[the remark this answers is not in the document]
justsmth: I agree. Latest commit uses a `MAX_CIPHER_KEY_LEN`.

aws-lc-rs/src/cipher.rs
```diff
@@ -3,158 +3,1268 @@
...
+                // This implements PKCS#7 padding scheme, used by aws-lc if we were using EVP_CIPHER API's
+                let remainder = in_out_len % block_len;
+                if remainder == 0 {
+                    let block_size: u8 = block_len.try_into().map_err(|_| Unspecified)?;
+                    in_out.extend(vec![block_size; block_len].iter());
+                } else {
+                    let padding_size = block_len - remainder;
+                    let v: u8 = padding_size.try_into().map_err(|_| Unspecified)?;
+                    // Heap allocation :(
+                    in_out.extend(vec![v; padding_size].iter());
```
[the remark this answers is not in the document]
justsmth: It would apply to both branches. 

I updated the implementation in the latest commit to use a 16-byte `padding_buffer` instead of the `vec!` macro. With this change it's possibly not a "heap allocation" depending on the "capacity" of the collection and how it implements the [`Extend`](https://dev-doc.rust-lang.org/beta/std/iter/trait.Extend.html) trait.

aws-lc-rs/src/cipher/key.rs
```diff
@@ -0,0 +1,200 @@
...
+            SymmetricCipherKey::Aes128 { enc_key, .. }
+            | SymmetricCipherKey::Aes256 { enc_key, .. } => encrypt_block_aes(enc_key, block),
+            SymmetricCipherKey::ChaCha20 { .. } => panic!("Unsupported algorithm!"),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::cipher::block::{Block, BLOCK_LEN};
```
samuel40791765, line 170: Is `BLOCK_LEN` redundant with `AES_BLOCK_LEN`, or do they imply different things?

aws-lc-rs/examples/cipher.rs
```diff
@@ -0,0 +1,233 @@
...
+
+fn aes_ctr_encrypt(key: &[u8], iv: Option<String>, plaintext: String) -> Result<(), &'static str> {
+    let hex_key = hex::encode(key);
+    let key = new_unbound_key(key)?;
+
+    let key = match iv {
+        Some(iv) => {
+            let iv = {
+                let v = hex::decode(iv).map_err(|_| "invalid iv")?;
+                let v: FixedLength<16> = v.as_slice().try_into().map_err(|_| "invalid iv")?;
```
samuel40791765, line 107: Can we use `AES_IV_LEN` here? Same for the other instances below, it might be best to show in our examples that the user can depend on the predefined values we provide.
```suggestion
                let v: FixedLength<AES_IV_LEN> = v.as_slice().try_into().map_err(|_| "invalid iv")?;
```

aws-lc-rs/src/cipher/key.rs
```diff
@@ -0,0 +1,200 @@
...
+            Ok(SymmetricCipherKey::Aes128 {
+                raw_key: Aes128Key(kb.assume_init()),
+                enc_key,
+                dec_key,
+            })
+        }
+    }
+
+    pub(crate) fn aes256(key_bytes: &[u8]) -> Result<Self, Unspecified> {
+        if key_bytes.len() != 32 {
```
samuel40791765, line 99:
```suggestion
        if key_bytes.len() != AES_256_KEY_LEN {
```
Maybe we can depend on `AES_128_KEY_LEN` and `AES_256_KEY_LEN` for all the instances of `16` and `32` in this file (except for `chacha20`).

samuel40791765 (commented):
initialization in initialization vector seems to be misspelled throughout the PR, so we might want to adjust that.

aws-lc-rs/src/cipher.rs
```diff
@@ -3,158 +3,1270 @@
...
-                value.zeroize();
+            Self::Iv128(_) => write!(f, "Iv128"),
+            Self::None => write!(f, "None"),
+        }
+    }
+}
+
+#[non_exhaustive]
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+/// Cipher algorithm identifier.
```
torben-hansen, line 283: should we call this "block cipher", which is the standard nomenclature. But I guess all those names are just a subset of "algorithm" anyway.

aws-lc-rs/src/cipher.rs
```diff
@@ -3,158 +3,1270 @@
...
+    PKCS7,
+}
+
+impl PaddingStrategy {
+    fn add_padding<InOut>(self, block_len: usize, in_out: &mut InOut) -> Result<(), Unspecified>
+    where
+        InOut: AsMut<[u8]> + for<'in_out> Extend<&'in_out u8>,
+    {
+        match self {
+            PaddingStrategy::PKCS7 => {
```
torben-hansen, line 116: one restriction is that the padding length must be strictly less than 256 (otherwise it doesn't fit in an octet).

aws-lc-rs/src/cipher.rs
```diff
@@ -3,158 +3,1270 @@
...
+        EncryptingKey::new(key, OperatingMode::CTR, None)
+    }
+
+    /// Constructs an `EncryptingKey` operating in counter (CTR) mode using the provided key.
+    /// The users provided context will be used for the CTR mode initalization-vector.
+    ///
+    /// # Errors
+    /// * [`Unspecified`]: Returned if there is an error creating the `EncryptingKey`.
+    ///
+    pub fn less_safe_ctr(
```
torben-hansen, line 621: Still pondering on the name. It's not inherently less safe than the other function call....

aws-lc-rs/src/cipher.rs
```diff
@@ -3,158 +3,1270 @@
...
+    pub fn mode(&self) -> OperatingMode {
+        self.mode
+    }
+
+    /// Decrypts the data provided in `in_out` in-place.
+    /// Returns a references to the decrypted data.
+    ///
+    /// # Errors
+    /// * [`Unspecified`]: Returned if cipher mode requires input to be a multiple of the block length,
+    /// and `in_out.len()` is not. Otherwise returned if decryption fails.
```
torben-hansen, line 751:
```suggestion
    /// and `in_out.len()` is not. Also returned if decryption fails.
```
?

aws-lc-rs/src/cipher/key.rs
justsmth: Both the type `Block` and the value `BLOCK_LEN` are only internally visible (`pub(crate)`). The type is primarily used for [`aead::HeaderProtectionKey::new_mask`](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/aead/quic/struct.HeaderProtectionKey.html#method.new_mask) which uses a "sample_len" of 16 (even for ChaCha20), which also happens to align with `BLOCK_LEN`.

aws-lc-rs/src/cipher.rs
```diff
@@ -3,158 +3,1270 @@
...
 use zeroize::Zeroize;
 
-pub(crate) enum SymmetricCipherKey {
-    Aes128(Aes128Key, AES_KEY),
-    Aes256(Aes256Key, AES_KEY),
-    ChaCha20(ChaCha20Key),
+/// The cipher block padding strategy.
+#[non_exhaustive]
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum PaddingStrategy {
```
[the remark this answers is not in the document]
justsmth: * Having a method that only provides a specific *mode* and *padding* combination is intentional. The goal is guide the user toward the most common (or the "best practice" for) mode/padding combinations. We anticipate adding methods for other combinations, or even opening it up to arbitrary combinations in the future.
* I prefer to expose this type to consumers, but I agree that an enum with only one variant is not very useful.

aws-lc-rs/src/cipher.rs
```diff
@@ -3,158 +3,1270 @@
...
+        }
+
+        let cipher_key = UnboundCipherKey::new(alg, key).unwrap();
+        let encrypting_key = EncryptingKey::new(cipher_key, mode, None).unwrap();
+
+        let mut in_out = input.clone();
+        let decrypt_iv = encrypting_key.encrypt(&mut in_out).unwrap();
+
+        if n > 5 {
+            // There's no more than a 1 in 2^48 chance that this will fail randomly
```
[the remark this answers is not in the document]
justsmth: This is a test where we originally didn't have the `n > 5` condition, so with 1 byte there was about a (1/256) probability that the ciphertext would be the same as the plaintext and the next test condition would fail.  For n bytes, the probability of a match is around 1/(2^(n*8)).

aws-lc-rs/src/cipher.rs
justsmth: We intend for this API to potentially support Stream ciphers (like ChaCha) as well.

skmcgrail:
https://github.com/awslabs/aws-lc-rs/pull/155 has some API improvements which will allow reuse of the Encrypt/Decrypt keys so that they aren't consumed by value after calling encrypt/decrypt. Will allow us a path for future optimizations if/when we decide to migrate to the EVP_CIPHER based APIs.

aws-lc-rs/src/cipher/key.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+}
+
+impl SymmetricCipherKey {
+    pub(crate) fn aes128(key_bytes: &[u8]) -> Result<Self, Unspecified> {
+        if key_bytes.len() != AES_128_KEY_LEN {
+            return Err(Unspecified);
+        }
+
+        unsafe {
```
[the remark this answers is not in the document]
camshaft: If I'm not mistaken, this is the same for the [AEAD types](https://docs.rs/aws-lc-rs/1.1.0/aws_lc_rs/aead/index.html) as well. I wish it wasn't like this, though, since we have to pay both the latency and memory costs. This is especially wasteful for things like TLS and QUIC where we have separate derived sealer and opener keys for both directions so the other halves never get used.

aws-lc-rs/src/hkdf.rs
```diff
@@ -92,6 +92,7 @@ pub struct Salt {
     key_len: usize,
 }
 
+#[allow(clippy::missing_fields_in_debug)]
```
camshaft, line 95: nit: This seems unrelated to this PR. In the future, it would be easier to review with more focused PRs.

aws-lc-rs/src/iv.rs
```diff
@@ -0,0 +1,98 @@
+// Copyright 2018 Brian Smith.
+// SPDX-License-Identifier: ISC
+// Modifications copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+#![allow(dead_code)]
+
+//! Initialization Vector (IV) cryptographic primitives
```
camshaft, line 7: This module could also be a separate PR, along with using it everywhere that isn't the CBC/CTR stuff

aws-lc-rs/tests/cipher_test.rs
```diff
@@ -0,0 +1,360 @@
...
+            assert_eq!(expected_ciphertext.as_slice(), in_out.as_slice());
+
+            let unbound_key2 = UnboundCipherKey::new($alg, &key).unwrap();
+            let decrypting_key =
+                PaddedBlockDecryptingKey::$constructor(unbound_key2, context).unwrap();
+            assert_eq!($mode, decrypting_key.mode());
+            assert_eq!($padding, decrypting_key.padding());
+            assert_eq!($alg, decrypting_key.algorithm());
+            let plaintext = decrypting_key.decrypt(&mut in_out).unwrap();
+            assert_eq!(input.as_slice(), plaintext);
```
camshaft, line 41: nit: it might be worth extracting some of this out into an actual function rather than a macro. It's especially annoying to work with `paste!` and other tooling like rustfmt or rust-analyzer.

camshaft (commented):
I got through everything outside of `cipher.rs` and `cipher/key.rs`. I'll need to do the rest tomorrow.

aws-lc-rs/src/cipher/key.rs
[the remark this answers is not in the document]
torben-hansen: All uni-directional protocols will have to pay this cost. Probably somewhat minor(?), but still.

aws-lc-rs/benches/cipher_benchmark.rs
```diff
@@ -0,0 +1,165 @@
+use aws_lc_rs::cipher::{
```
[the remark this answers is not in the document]
justsmth: PR: https://github.com/awslabs/aws-lc-rs/pull/158

aws-lc-rs/src/cipher/key.rs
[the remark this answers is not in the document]
justsmth: We could lazily initialize these values at the cost of some code complexity and the additional latency of a branch or two in the logic. These are optimizations we should make, but in a separate PR.

On a related note, my [PR here](https://github.com/awslabs/aws-lc-rs/pull/156) eliminates one of the fields of this struct.

aws-lc-rs/src/cipher.rs
```diff
@@ -3,158 +3,1270 @@
...
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum OperatingMode {
+    /// Cipher block chaining (CBC) mode.
+    CBC,
+
+    /// Counter (CTR) mode.
+    CTR,
+}
+
+/// The contextual data used to encrypted/decrypt data.
```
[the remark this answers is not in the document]
skmcgrail: Examples in the repository aren't immediately discoverable unless the user looks for them, or you direct them to a hard-coded link to them which might change later. This allows the code to be directly inlined into the documentation, and it is tested as part of `cargo test` ensuring it doesn't drift in the future.

aws-lc-rs/src/cipher.rs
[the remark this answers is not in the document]
samuel40791765: Yeah I like the concept of exposing it once we have support for additional variants.

aws-lc-rs/src/hkdf.rs
justsmth: The latest commit merged changes (like this one) which were already on main, so some of the "noise" like this should be gone from the diff.

torben-hansen (approved):
CI looks unhappy. But otherwise, OK from me.

justsmth:
> CI looks unhappy. But otherwise, OK from me.

CI is happy now.

samuel40791765 (approved):

PR #156, opened 2023-06-14
Cleanup of `quic` module implementation

justsmth:
### Issues:
N/A

### Description of changes:
* Cleanup of the `quic` modules implementation to simplify upcoming integration of AES CBC/CTR:
   * `aead::quic::HeaderProtectionKey` now contains a `SymmetricCipherKey` (instead of an `aead::KeyInner`).   
   * Renames internal `aead::KeyInner` to `AeadCtx`, removes its `SymmetricCipherKey` field.
   * Removes the "raw" AES keys from `SymmetricCipherKey`.

### Call-outs:
N/A

### Testing:
N/A


aws-lc-rs/src/aead/aead_ctx.rs
```diff
@@ -0,0 +1,100 @@
...
+        }
+        unsafe {
+            let aead = ConstPointer::new(EVP_aead_chacha20_poly1305())?;
+            Ok(AeadCtx::CHACHA20_POLY1305(AeadCtx::build_context(
+                &aead, key_bytes,
+            )?))
+        }
+    }
+
+    unsafe fn build_context(
```
skmcgrail, line 68: If you isolate the `unsafe` into the body, you could remove needing to wrap all the functions above.

justsmth: The call to get the `*const EVP_AEAD` (e.g., `EVP_aead_aes_128_gcm()`) also requires an unsafe block. I don't see a good way of eliminating the unsafe block from the calling function.

skmcgrail (approved):

samuel40791765 (dismissed):

aws-lc-rs/src/aead/aead_ctx.rs
justsmth: I was able to avoid several unsafe blocks by passing a function pointer to `build_context`.

PR #165, opened 2023-06-20
Smart-pointers to seperate out encryption & decryption contexts

skmcgrail:


aws-lc-rs/src/cipher.rs
```diff
@@ -243,41 +351,54 @@ pub enum OperatingMode {
...
-///
-///
-#[non_exhaustive]
-pub enum CipherContext {
-    /// A 128-bit Initialization Vector.
-    Iv128(FixedLength<IV_LEN_128_BIT>),
+pub struct DecryptionContext(CipherContext);
 
-    /// No input to the cipher mode.
-    None,
```
skmcgrail, line 261: I removed `None`, since we don't support any modes currently that would return this value.

PR #174, opened 2023-06-29
Introduce KEM API with support for Kyber-512

bencivjan:
### Description of changes:
This pull request introduces a KEM API into the library with support for Kyber-512

### Call-outs:
- Creates `aws_lc_rs::key_transport` module which provides KEM API
- `KemAlgorithm` contains the offered KEM algorithms
- `KemPrivateKey` and `KemPublicKey` manage the public and private keys (respectively). `KemPrivateKey` offers a method to generate a corresponding `KemPublicKey` and a method to decapsulate a ciphertext. `KemPublicKey` offers a method an encapsulation method, generating a shared secret and ciphertext.

### Testing:
- Integration test performing a key transfer using Kyber-512
- Integration test performing a key transfer using Kyber-512 with serialization/deserialization of keys
- Unit testing of key serialization/deserialization functionality


aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,389 @@
...
+    /// from the operation and then returns what `kdf` returns.
+    /// 
+    /// # Errors
+    /// `error::Unspecified` when operation fails due to internal error.
+    /// 
+    pub fn decapsulate<F, R>(&self, ciphertext: &mut [u8], kdf: F) -> Result<R, Unspecified>
+    where
+        F: FnOnce(&[u8]) -> Result<R, Unspecified> {
+        unsafe {
+            let ctx = DetachableLcPtr::new(EVP_PKEY_CTX_new(*self.context, null_mut()))?;
```
justsmth, line 178: I think all the places using `DetachableLcPtr` could just be `LcPtr`.

justsmth (changes requested):
Run `make format` before the next commit.

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,389 @@
...
+        }
+    }
+}
+/// A serializable private key usable with KEMs. This can be randomly generated
+/// or constructed from raw bytes.
+#[derive(Debug)]
+pub struct KemPrivateKey {
+    algorithm: KemAlgorithm,
+    context: LcPtr<*mut EVP_PKEY>,
+    priv_key: Vec<u8>
```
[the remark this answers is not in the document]
bencivjan: It's necessary for us to do this if we want to return an `&[u8]` in the `get_raw_bytes` method. I think a Vec makes the implementation easier because we don't always know the length of what we are allocating at compile time. Why would a Box<[u8]> be better?

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,389 @@
...
+                panic!("Error getting private key from context");
+            }
+            if 1 != EVP_PKEY_get_raw_private_key(*other.context, other_privkey_bytes_from_ctx.as_mut_ptr(), &mut other.algorithm.get_privkey_len()) {
+                panic!("Error getting private key from context");
+            }
+        }
+        self.algorithm == other.algorithm &&
+        self.priv_key == other.priv_key &&
+        self_privkey_bytes_from_ctx == other_privkey_bytes_from_ctx
+    }
```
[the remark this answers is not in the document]
bencivjan: I suppose, yes. The reason I added this was to essentially check the validity of the EVP_PKEY pointer. Since when reading from raw bytes technically these aren't exactly the same, as the raw bytes are written directly to the buffer while the EVP_PKEY is constructed from them.
```rust
let pkey = DetachableLcPtr::new(EVP_PKEY_kem_new_raw_secret_key(
    alg.nid(),
    bytes.as_ptr(),
    bytes.len(),
))?;
Ok(KemPrivateKey {
    algorithm: alg,
    context: LcPtr::from(pkey),
    priv_key: bytes.to_owned(),
})
``` 

If this is unnecessary I can remove this extra check.

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,389 @@
...
+
+    /// Gets the raw bytes of the current `KemPrivateKey`
+    pub fn get_raw_bytes(&self) -> &[u8] {
+        &self.priv_key
+    }
+}
+
+#[cfg(test)]
+impl PartialEq for KemPrivateKey {
```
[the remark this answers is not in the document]
bencivjan: Ok. If we are getting rid of the EVP_PKEY pointer assertion then we can get rid of the PartialEq implementation.

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,389 @@
...
+                context: LcPtr::from(pkey),
+                priv_key: bytes.to_owned()
+            })
+        }
+    }
+
+    /// Gets the raw bytes of the current `KemPrivateKey`
+    pub fn get_raw_bytes(&self) -> &[u8] {
+        &self.priv_key
+    }
```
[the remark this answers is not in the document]
bencivjan: Ok

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,389 @@
...
+    /// Creates a new KEM private key from raw bytes
+    /// 
+    /// `alg` is the `KemAlgorithm` to be associated with the generated `KemPrivateKey`
+    /// 
+    /// `bytes` is a slice of raw bytes representing a `KemPrivateKey`
+    /// 
+    /// # Errors
+    /// `error::KeyRejected` when operation fails during key creation.
+    /// 
+    pub fn from_raw_bytes(alg: KemAlgorithm, bytes: &[u8]) -> Result<Self, KeyRejected> {
```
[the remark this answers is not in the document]
bencivjan: Yeah we can rename it

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,389 @@
...
+        unsafe {
+            let pkey = DetachableLcPtr::new(EVP_PKEY_kem_new_raw_secret_key(alg.nid(), bytes.as_ptr(), bytes.len())).map_err(|_e| KeyRejected::unexpected_error())?;
```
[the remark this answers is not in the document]
bencivjan: Ack on LcPtr. That's interesting with the map_err, is it because the unit type can be considered a KeyRejected type?

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,389 @@
...
+    /// `alg` is the `KemAlgorithm` to be associated with the generated `KemPublicKey`
+    /// 
+    /// `bytes` is a slice of raw bytes representing a `KemPublicKey`
+    /// 
+    /// # Errors
+    /// `error::KeyRejected` when operation fails during key creation.
+    /// 
+    pub fn from_raw_bytes(alg: KemAlgorithm, bytes: &[u8]) -> Result<Self, KeyRejected> {
+        unsafe {
+            let pubkey_ctx = EVP_PKEY_kem_new_raw_public_key(alg.nid(), bytes.as_ptr(), alg.get_pubkey_len());
```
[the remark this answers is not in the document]
bencivjan: Ack

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,389 @@
...
+    EVP_PKEY_KEM, EVP_PKEY_kem_new_raw_secret_key, EVP_PKEY_CTX_kem_set_params, EVP_PKEY_CTX_new,
+    EVP_PKEY_encapsulate, EVP_PKEY_decapsulate, EVP_PKEY_get_raw_public_key, EVP_PKEY_kem_new_raw_public_key, EVP_PKEY_get_raw_private_key
+};
+use zeroize::Zeroize;
+
+
+const KYBER512_SECRETKEYBYTES: usize = 1632;
+const KYBER512_PUBLICKEYBYTES: usize = 800;
+const KYBER512_CIPHERTEXTBYTES: usize = 768;
+const KYBER512_BYTES: usize = 32;
```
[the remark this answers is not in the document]
bencivjan: I stole these from the aws-lc source code for Kyber. I agree with the name changes. I'll double check the numbers

aws-lc-rs/src/key_transport.rs
bencivjan: Ack

aws-lc-rs/src/key_transport.rs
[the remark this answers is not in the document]
andrewhop: Also you don't need to duplicate these, if you call keygen/encapsulate/decapsulate with null buffers we set the expected size for you. This could make it easier to add more KEMs in the future. https://github.com/aws/aws-lc/blob/main/include/openssl/evp.h#L723-L726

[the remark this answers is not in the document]
bencivjan: Yeah, I thought maybe it would be better for performance if we're making less C FFI calls. Probably negligible though and worth getting rid of the constants.

aws-lc-rs/src/key_transport.rs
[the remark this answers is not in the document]
bencivjan: After removing macros [per Andrew's comment](https://github.com/aws/aws-lc-rs/pull/174#discussion_r1248167658) we can no longer verify that bytes.len() matches the provided algorithm. This is because in order to get the public key length we need to already have an EVP_PKEY, which is what we are creating here.

aws-lc-rs/src/key_transport.rs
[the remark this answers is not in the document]
justsmth: We use constants for fixed key size everywhere else in the package. I think they're fine as long as the logic encapsulate/decapsulate/etc is obtaining these sizes via a call to the `KemAlgorithm` and not directly.

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,394 @@
...
+use std::ptr::null_mut;
+use zeroize::Zeroize;
+
+#[allow(non_camel_case_types)]
+#[derive(Clone, Debug, PartialEq)]
+/// A selection of algorithms to be used as KEMs.
+pub enum KemAlgorithm {
+    /// NIST Round 3 iteration of the Kyber-512 algorithm
+    KYBER512_R3,
+}
```
justsmth, line 70: * Any public `enum` like should also be `#[non_exhaustive]`.
* It might be better for `KemAlgorithm` to be a pub struct (with fields for those constants) for which we provide `static` references. This is how we expose algorithms in the other modules: [aead](https://docs.rs/aws-lc-rs/1.2.0/aws_lc_rs/aead/index.html#statics), [hkdf](https://docs.rs/aws-lc-rs/1.2.0/aws_lc_rs/hkdf/index.html#statics), [cipher](https://docs.rs/aws-lc-rs/1.2.0/aws_lc_rs/cipher/index.html#statics), ...

aws-lc-rs/src/key_transport.rs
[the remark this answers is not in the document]
bencivjan: Lengths are as specified on the crystals website: [https://pq-crystals.org/kyber/](https://pq-crystals.org/kyber/)

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// Creates a new KEM private key from raw bytes
+    ///
+    /// `alg` is the `KemAlgorithm` to be associated with the generated `KemPrivateKey`
+    ///
+    /// `bytes` is a slice of raw bytes representing a `KemPrivateKey`
+    ///
+    /// # Errors
+    /// `error::KeyRejected` when operation fails during key creation.
+    ///
+    pub fn new(alg: KemAlgorithm, bytes: &[u8]) -> Result<Self, KeyRejected> {
```
samuel40791765, line 219: We should make it clear in the documentation that `KemPrivateKey` should generally be created from `KemPrivateKey::generate`. This is only used to recreate private keys that had been previously generated and stored.

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,394 @@
...
+            if EVP_PKEY_encapsulate(
+                *ctx,
+                ciphertext.as_mut_ptr(),
+                &mut ciphertext_len,
+                shared_secret.as_mut_ptr(),
+                &mut shared_secret_len,
+            ) != 1
+            {
+                ciphertext.zeroize();
+                shared_secret.zeroize();
```
samuel40791765, line 299: * Not quite sure if these need to be zerioized either. Same for above.
* Zeroization is great in that we don't leak any potential private data. But I think it should be more valuable when we already have secret values capable of being leaked. If we do want to implement Zeroization, something like what we're doing for HKDF below would be more valuable. This helps ensure that any available secret data we have is zeroized before it's being freed/dropped. 
* https://github.com/aws/aws-lc-rs/blob/832e839e9013d96775df726df32f566e4558b3bd/aws-lc-rs/src/hkdf.rs#L104-L108

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// Return the algorithm associated with the given KEM public key.
+    #[must_use]
+    pub fn algorithm(&self) -> &KemAlgorithm {
+        &self.algorithm
+    }
+
+    /// Performs the encapsulate operation using the current KEM public key, generating a ciphertext
+    /// and associated shared secret.
+    ///
+    /// After the decapsulation is finished, `decapsulate` calls `kdf` with the ciphertext and raw shared secret
```
samuel40791765, line 260:
```suggestion
    /// After the encapsulation is finished, `encapsulate` calls `kdf` with the ciphertext and raw shared secret
```

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,394 @@
...
+//! // protocol message.
+//! let pub_key_bytes = pub_key.as_ref();
+//!
+//! let mut ciphertext: Vec<u8> = vec![];
+//!
+//! let retrieved_pub_key = KemPublicKey::new(KemAlgorithm::KYBER512_R3, pub_key_bytes)?;
+//! let bob_result = retrieved_pub_key.encapsulate(|ct, ss| {
+//!     ciphertext.extend_from_slice(ct);
+//!     // In a real application, we'd apply a KDF to the shared secret and the
+//!     // public keys (as recommended in RFC 7748) and then derive session
```
samuel40791765, line 32: RFC 7748 is for Elliptic Curves Security. Not sure if there's a standardization for KEMs or Kyber yet, but the recommendation doesn't seem quite applicable here.
Maybe we could just mention this was the norm in real applications for ECs?

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,394 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+//! Key Transport: KEMs, including support for Kyber.
+//!
+//! # Example
+//!
+//! Note that this example uses Kyber-512, but other algorithms can be used
+//! exactly the same way, just substituting
```
samuel40791765, line 9:
```suggestion
//! in the exact same way by just substituting
```

aws-lc-rs/tests/key_transport_tests.rs
```diff
@@ -0,0 +1,69 @@
...
+    let priv_key = KemPrivateKey::generate(KemAlgorithm::KYBER512_R3).unwrap();
+    assert_eq!(priv_key.algorithm(), &KemAlgorithm::KYBER512_R3);
+
+    // Generate private key bytes to possibly save for later
+    let privkey_raw_bytes = priv_key.as_ref();
+
+    let pub_key = priv_key.compute_public_key().unwrap();
+
+    // Generate public key bytes to send to bob
+    let pub_key_bytes = pub_key.as_ref();
```
samuel40791765, line 44: Minor nit: 
```suggestion
    // Generate public key bytes to send to bob
    let pub_key_bytes = priv_key.compute_public_key().unwrap().as_ref();
```

bencivjan: Can't do this because of lifetimes

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// # Errors
+    /// `error::Unspecified` when operation fails due to internal error.
+    ///
+    pub fn decapsulate<F, R>(&self, ciphertext: &mut [u8], kdf: F) -> Result<R, Unspecified>
+    where
+        F: FnOnce(&[u8]) -> Result<R, Unspecified>,
+    {
+        unsafe {
+            let ctx = LcPtr::new(EVP_PKEY_CTX_new(*self.pkey, null_mut()))?;
+            let mut shared_secret: Vec<u8> = vec![0u8; self.algorithm.shared_secret_size()];
```
samuel40791765, line 213: Does this still need to be allocated on the heap, now that we know the size beforehand? Same question for all the other instances where we've abstracted the size values into `KemAlgorithm`

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,404 @@
...
+        match self {
+            KemAlgorithmID::Kyber512_R3 => NID_KYBER512_R3,
+        }
+    }
+}
+
+impl KemAlgorithm {
+    #[inline]
+    fn secret_key_size(&self) -> usize {
+        self.secret_key_size
```
samuel40791765, line 105: Can't we directly access these values without the get functions?

bencivjan: Yeah but sometimes we have to pass the length of a key by pointer to C functions and it can possibly be overwritten. The getter returns a copy of the integer so the static struct's values are never accidentally changed. Also it shows the programmer that these values shouldn't be modified.

aws-lc-rs/src/key_transport.rs
bencivjan: They don't but the key lengths can be really big (Kyber-512 SK for example is 1632 bytes and Kyber-1024 SK is 3168 bytes). Also aws-lc expects these dynamically allocated

aws-lc-rs/src/key_transport.rs
samuel40791765: We could copy the value into another mutable pointer to be passed into the C function. The static struct's values aren't accidentally changed either way.
```
let mut secret_key_size = alg.secret_key_size;
let mut priv_key_bytes = vec![0u8; secret_key_size];
if 1 != EVP_PKEY_get_raw_private_key(
        *kyber_key,
         priv_key_bytes.as_mut_ptr(),
         &mut secret_key_size,
) {
        return Err(Unspecified);
}
```
I think the immutable variable type in the static structure should already show the programmer that the value can't be modified. The way these structure values are immutable and Rust disallows directly passing an immutable reference to the mutable pointer, sort of implies they can't be accidentally changed.

samuel40791765: I'm fine either way though. I was just curious about why this was needed, and how the rest of the codebase was taking care of this circumstance.

aws-lc-rs/src/key_transport.rs
samuel40791765: I'm not sure if AWS-LC has an expectation for how the memory is allocated. It should be on the user to allocate enough memory either way.
I do agree with the concern for larger key lengths. But I don't have enough experience with how large an array can be before it becomes too "big" to maintain on the stack in this case. It's not a one way door though, we can always look into it later during benchmarking :)

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// Creates a new KEM private key from raw bytes
+    ///
+    /// `alg` is the `KemAlgorithm` to be associated with the generated `KemPrivateKey`
+    ///
+    /// `bytes` is a slice of raw bytes representing a `KemPrivateKey`
+    ///
+    /// # Errors
+    /// `error::KeyRejected` when operation fails during key creation.
+    ///
+    pub fn new(alg: &'static KemAlgorithm, bytes: &[u8]) -> Result<Self, KeyRejected> {
```
samuel40791765, line 237: This comment I had seems to have gotten lost: 
https://github.com/aws/aws-lc-rs/pull/174#discussion_r1248317298

aws-lc-rs/src/key_transport.rs
bencivjan: Yeah I see what you're saying. Definitely could do it either way. I guess it just depends what it more idiomatic to Rust, which I'm not really sure of

bencivjan: You can see that in [cipher.rs](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-rs/src/cipher.rs#L307-L320) they create getters for the relevant struct fields. So seems like a common pattern in the library to me.

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+        self.ciphertext_size
+    }
+
+    #[inline]
+    fn shared_secret_size(&self) -> usize {
+        self.shared_secret_size
+    }
+}
+
+/// A serializable private key usable with KEMs. This can be randomly generated
```
samuel40791765, line 126:
```suggestion
/// A serializable private key usable with KEMs. This can be randomly generated with KemPrivateKey::generate
```

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+    ///
+    /// # Errors
+    /// `error::KeyRejected` when operation fails during key creation.
+    ///
+    pub fn new(alg: &'static KemAlgorithm, bytes: &[u8]) -> Result<Self, KeyRejected> {
+        if alg.secret_key_size() != bytes.len() {
+            return Err(KeyRejected::unexpected_error());
+        }
+        unsafe {
+            let privkey_ctx = LcPtr::new(EVP_PKEY_kem_new_raw_secret_key(
```
samuel40791765, line 244: Nit: 
```suggestion
            let privkey = LcPtr::new(EVP_PKEY_kem_new_raw_secret_key(
```

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+        unsafe {
+            if 1 != EVP_PKEY_get_raw_public_key(
+                *self.pkey,
+                pubkey_bytes.as_mut_ptr(),
+                &mut self.algorithm.public_key_size(),
+            ) {
+                return Err(Unspecified);
+            }
+
+            let pubkey_ctx_copy = LcPtr::new(EVP_PKEY_kem_new_raw_public_key(
```
samuel40791765, line 182: Minor nit:
```suggestion
            let pubkey = LcPtr::new(EVP_PKEY_kem_new_raw_public_key(
```

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+    ///
+    /// # Errors
+    /// `error::KeyRejected` when operation fails during key creation.
+    ///
+    pub fn new(alg: &'static KemAlgorithm, bytes: &[u8]) -> Result<Self, KeyRejected> {
+        if alg.public_key_size() != bytes.len() {
+            return Err(KeyRejected::unexpected_error());
+        }
+        unsafe {
+            let pubkey_ctx = LcPtr::new(EVP_PKEY_kem_new_raw_public_key(
```
samuel40791765, line 336: Minor nit: 
```suggestion
            let pubkey = LcPtr::new(EVP_PKEY_kem_new_raw_public_key(
```

samuel40791765 (approved):

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+
+            let pubkey_ctx_copy = LcPtr::new(EVP_PKEY_kem_new_raw_public_key(
+                self.algorithm.id.nid(),
+                pubkey_bytes.as_mut_ptr(),
+                self.algorithm.public_key_size(),
+            ))?;
+
+            Ok(KemPublicKey {
+                algorithm: self.algorithm,
+                pkey: pubkey_ctx_copy,
```
bencivjan, line 190:
```suggestion
                pkey: pubkey,
```

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+        }
+        unsafe {
+            let privkey_ctx = LcPtr::new(EVP_PKEY_kem_new_raw_secret_key(
+                alg.id.nid(),
+                bytes.as_ptr(),
+                bytes.len(),
+            ))?;
+            Ok(KemPrivateKey {
+                algorithm: alg,
+                pkey: privkey_ctx,
```
bencivjan, line 251:
```suggestion
                pkey: privkey,
```

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+        }
+        unsafe {
+            let pubkey_ctx = LcPtr::new(EVP_PKEY_kem_new_raw_public_key(
+                alg.id.nid(),
+                bytes.as_ptr(),
+                bytes.len(),
+            ))?;
+            Ok(KemPublicKey {
+                algorithm: alg,
+                pkey: pubkey_ctx,
```
bencivjan, line 343:
```suggestion
                pkey: pubkey,
```

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// Computes the KEM public key from the KEM private key
+    ///
+    /// # Errors
+    /// `error::Unspecified` when operation fails due to internal error.
+    ///
+    pub fn compute_public_key(&self) -> Result<KemPublicKey, Unspecified> {
+        let mut pubkey_bytes = vec![0u8; self.algorithm.public_key_size()];
+        unsafe {
+            if 1 != EVP_PKEY_get_raw_public_key(
+                *self.pkey,
```
justsmth, line 175: Since it's `const`, you should use `self.pkey.as_const_ptr()`.

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+        unsafe {
+            let privkey = LcPtr::new(EVP_PKEY_kem_new_raw_secret_key(
+                alg.id.nid(),
+                bytes.as_ptr(),
+                bytes.len(),
+            ))?;
+            Ok(KemPrivateKey {
+                algorithm: alg,
+                pkey: privkey,
+                priv_key: bytes.to_owned().into(),
```
justsmth, line 252: The `to_owned()` shouldn't be needed: `bytes.into()`

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+        unsafe {
+            let pubkey = LcPtr::new(EVP_PKEY_kem_new_raw_public_key(
+                alg.id.nid(),
+                bytes.as_ptr(),
+                bytes.len(),
+            ))?;
+            Ok(KemPublicKey {
+                algorithm: alg,
+                pkey: pubkey,
+                pub_key: bytes.to_owned().into(),
```
justsmth, line 344: The `to_owned()` shouldn't be needed: `bytes.into()`

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+    ///
+    /// # Errors
+    /// `error::Unspecified` when operation fails due to internal error.
+    ///
+    pub fn generate(alg: &'static KemAlgorithm) -> Result<Self, Unspecified> {
+        unsafe {
+            let kyber_key = kem_key_generate(alg.id.nid())?;
+            let mut priv_key_bytes = vec![0u8; alg.secret_key_size()];
+            if 1 != EVP_PKEY_get_raw_private_key(
+                *kyber_key,
```
justsmth, line 146: `kyber_key.as_const_ptr()`

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+                priv_key_bytes.as_mut_ptr(),
+                &mut alg.secret_key_size(),
```
justsmth, line 148: This value needs to be stored in a stack variable:
```
    let mut secret_key_size = alg.secret_key_size();
    let mut priv_key_bytes = vec![0u8; secret_key_size];
    if 1 != EVP_PKEY_get_raw_private_key(
        kyber_key.as_const_ptr(),
        priv_key_bytes.as_mut_ptr(),
        &mut secret_key_size,
    ) {
        return Err(Unspecified);
    }
```
Do we need to verify that `secret_key_size` was not changed prior to returning?

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+
+            let mut ciphertext: Vec<u8> = vec![0u8; self.algorithm.cipher_text_size()];
+            let mut shared_secret: Vec<u8> = vec![0u8; self.algorithm.shared_secret_size()];
+
+            if EVP_PKEY_encapsulate(
+                *ctx,
+                ciphertext.as_mut_ptr(),
+                &mut self.algorithm.cipher_text_size(),
+                shared_secret.as_mut_ptr(),
+                &mut self.algorithm.shared_secret_size(),
```
justsmth, line 310: Don't pass mutable pointers to return values:
```
    let mut cipher_text_size = self.algorithm.cipher_text_size();
    let mut shared_secret_size = self.algorithm.shared_secret_size();
    if EVP_PKEY_encapsulate(
        *ctx,
        ciphertext.as_mut_ptr(),
        &mut cipher_text_size,
        shared_secret.as_mut_ptr(),
        &mut shared_secret_size,
    ) != 1
    {
        return Err(Unspecified);
    }
```

As above, do we need to check that these values were not changed?

justsmth (changes requested):

aws-lc-rs/src/key_transport.rs
```diff
@@ -0,0 +1,408 @@
...
+unsafe fn kem_key_generate(nid: c_int) -> Result<LcPtr<*mut EVP_PKEY>, Unspecified> {
+    let ctx = LcPtr::new(EVP_PKEY_CTX_new_id(EVP_PKEY_KEM, null_mut()))?;
+    let mut key_raw: *mut EVP_PKEY = null_mut();
+    if 1 != EVP_PKEY_keygen_init(*ctx)
+        || 1 != EVP_PKEY_CTX_kem_set_params(*ctx, nid)
+        || 1 != EVP_PKEY_keygen(*ctx, &mut key_raw)
+    {
+        // We don't have the key wrapped with LcPtr yet, so explicitly free it
+        key_raw.free();
+        return Err(Unspecified);
```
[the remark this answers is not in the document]
bencivjan: > * Does `EVP_PKEY_CTX_kem_set_params` need to be called prior to `EVP_PKEY_keygen_init`?

The order of these doesn't matter. The important thing is that these are both called before `EVP_PKEY_keygen`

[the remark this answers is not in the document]
bencivjan: > If EVP_PKEY_keygen returns an error, nothing was allocated and key_raw is still null -- there's nothing for key_raw.free() to free.

Yeah it seems that way. The reason I did this was to follow what aws-lc provides in the KEM [ReadMe](https://github.com/aws/aws-lc/blob/main/crypto/kem/README.md?plain=1#L101-L108). I looked at the source though and it seems like everything is taken care of in the function. Not sure why they made the example like this...

aws-lc-rs/src/key_transport.rs
bencivjan: > Do we need to verify that secret_key_size was not changed prior to returning?

The key size would never be changed since we are passing in a defined constant length for the algorithm we are using.

aws-lc-rs/src/key_transport.rs
bencivjan: We shouldn't need to check that these are unchanged because we know that our getters will return the correct length

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,468 @@
...
+        unsafe {
+            let ctx = LcPtr::new(EVP_PKEY_CTX_new(*self.pkey, null_mut()));
+            if ctx.is_err() {
+                return Err(error_value);
+            }
+            let mut shared_secret_len = self.algorithm.shared_secret_size();
+            let mut shared_secret: Vec<u8> = vec![0u8; shared_secret_len];
+
+            if 1 != EVP_PKEY_decapsulate(
+                *ctx.unwrap(),
```
skmcgrail, line 229: Avoid the unwrap panic here. Instead prefer something like `.map_err(|_|Unspecified)?`

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,468 @@
...
+    pub fn decapsulate<F, R, E>(
+        &self,
+        ciphertext: &mut [u8],
+        error_value: E,
+        kdf: F,
+    ) -> Result<R, E>
+    where
+        F: FnOnce(&[u8]) -> Result<R, E>,
+    {
+        unsafe {
```
skmcgrail, line 220: Try to reduce the scopes of your unsafe blocks to the smallest portions as possible. Most of the code here isn't unsafe.

aws-lc-rs/src/kem.rs
bencivjan: This will never fail because we check if ctx is an error. `map_err` just maps the error, it doesn't actually get the value. We can't use `?` because the error type is caller defined.

justsmth: Ben - I see what you mean, but stylistically, we want to avoid using `unwrap` (as it's a "code smell"). The problem with `map_err` is that we need to return `error_value` but the type `E` is not `Copy` so it would get "moved" into the closure, and the borrow checker wouldn't let it be used later in the function.

The best way might be to use a let-else for the assignment so that `ctx` gets the `LcPtr`, the borrow-checker is happy, and we don't have to call `unwrap`:
```
    let Ok(ctx) =
        LcPtr::new(EVP_PKEY_CTX_new(*self.pkey, null_mut())) else {
        return Err(error_value);
    };
```

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,468 @@
...
+    #[allow(clippy::missing_panics_doc)]
+    pub fn encapsulate<F, R, E>(&self, error_value: E, kdf: F) -> Result<R, E>
+    where
+        F: FnOnce(&[u8], &[u8]) -> Result<R, E>,
+    {
+        unsafe {
+            let ctx = LcPtr::new(EVP_PKEY_CTX_new(*self.pkey, null_mut()));
+            if ctx.is_err() {
+                return Err(error_value);
+            }
```
justsmth, line 319: Use a let-else here as well to avoid `unwrap`:
```
    let Ok(ctx) =
        LcPtr::new(EVP_PKEY_CTX_new(*self.pkey, null_mut())) else {
        return Err(error_value);
    };
```

aws-lc-rs/tests/key_transport_tests.rs
```diff
@@ -0,0 +1,71 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
justsmth, line 1: Since the module was renamed to `kem`, the test file should be renamed as well.

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,468 @@
...
+    /// Generate a new KEM private key for the given algorithm.
+    ///
+    /// # Errors
+    /// `error::Unspecified` when operation fails due to internal error.
+    ///
+    pub fn generate(alg: &'static KemAlgorithm) -> Result<Self, Unspecified> {
+        unsafe {
+            let kyber_key = kem_key_generate(alg.id.nid())?;
+            let mut secret_key_size = alg.secret_key_size();
+            let mut priv_key_bytes = vec![0u8; secret_key_size];
```
justsmth, line 146: Nit -- `priv_key_bytes` initialization could be done before the `unsafe` block:
```
        let mut secret_key_size = alg.secret_key_size();
        let mut priv_key_bytes = vec![0u8; secret_key_size];
        unsafe { ...
```

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,468 @@
...
+    where
+        F: FnOnce(&[u8]) -> Result<R, E>,
+    {
+        unsafe {
+            let ctx = LcPtr::new(EVP_PKEY_CTX_new(*self.pkey, null_mut()));
+            if ctx.is_err() {
+                return Err(error_value);
+            }
+            let mut shared_secret_len = self.algorithm.shared_secret_size();
+            let mut shared_secret: Vec<u8> = vec![0u8; shared_secret_len];
```
justsmth, line 226: Nit -- `shared_secret` could be initialized before the `unsafe` block.

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,468 @@
...
+        unsafe {
+            let ctx = LcPtr::new(EVP_PKEY_CTX_new(*self.pkey, null_mut()));
+            if ctx.is_err() {
+                return Err(error_value);
+            }
+
+            let mut ciphertext_len = self.algorithm.ciphertext_size();
+            let mut shared_secret_len = self.algorithm.shared_secret_size();
+            let mut ciphertext: Vec<u8> = vec![0u8; ciphertext_len];
+            let mut shared_secret: Vec<u8> = vec![0u8; shared_secret_len];
```
justsmth, line 324: Nit -- This initialization could be done before the `unsafe` block.

justsmth (changes requested):

aws-lc-rs/src/kem.rs
bencivjan: Yeah the let else looks clean. [agreement.rs](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-rs/src/agreement.rs#L534) uses unwrap in the exact same way so we should probably refactor there then too.

bencivjan: > The problem with map_err is that we need to return error_value but the type E is not Copy so it would get "moved" into the closure, and the borrow checker wouldn't let it be used later in the function.

Right....I see what you mean about the "move".

justsmth: > Yeah the `let else` looks clean. [Agreement.rs](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-rs/src/agreement.rs#L534) uses unwrap in the exact same way, so we should probably refactor there then too.

https://github.com/aws/aws-lc-rs/pull/177

justsmth (approved):

Issue #160, opened 2023-06-15
Support building for aarch64-apple-ios-sim


tomleavy:
### Problem:

Targeting the iOS device (aarch64-apple-ios) works properly but building for the simulator doesn't. I can generally build aws-lc for iOS simulator, so I suspect that this is a problem between the cmake cargo crate and aws-lc cmake.

Building aws-lc for iOS simulator works with the following cmake command

```
cmake -DBUILD_TESTING=OFF -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DCMAKE_SYSTEM_NAME="iOS" -DCMAKE_OSX_ARCHITECTURES="arm64" -DCMAKE_OSX_SYSROOT="iphonesimulator" -DBUILD_LIBSSL=OFF ..
```

#### Relevant details
AWS-LC for Rust versions or commit: latest main

System information: for linux, below info can be collected by running `uname -srvmp`
 * CPU architecture: aarch64
 * Host OS: MacOS 13.4 (arm) + XCode 14.3.1
 * Compiler: Rust 1.70
 * Target OS: iOS Simulator (aarch64-apple-ios-sim)
 
Build log:
 
```text
❯ cargo build --release -p aws-lc-sys --target aarch64-apple-ios-sim
   Compiling libc v0.2.146
   Compiling memchr v2.5.0
   Compiling regex-syntax v0.7.2
   Compiling syn v2.0.18
   Compiling nom v7.1.3
   Compiling which v4.4.0
   Compiling clang-sys v1.6.1
   Compiling regex v1.8.4
   Compiling cexpr v0.6.0
   Compiling prettyplease v0.2.6
   Compiling bindgen v0.65.1
   Compiling aws-lc-sys v0.7.0 (/Users/tomleavy/git/aws-lc-rs/aws-lc-sys)
error: failed to run custom build command for `aws-lc-sys v0.7.0 (/Users/tomleavy/git/aws-lc-rs/aws-lc-sys)`

Caused by:
  process didn't exit successfully: `/Users/tomleavy/git/aws-lc-rs/target/release/build/aws-lc-sys-e1e4be26c4b092ee/build-script-main` (exit status: 101)
  --- stdout
  cargo:rustc-cfg=use_bindgen_generated
  CMAKE_TOOLCHAIN_FILE_aarch64-apple-ios-sim = None
  CMAKE_TOOLCHAIN_FILE_aarch64_apple_ios_sim = None
[...]

PR #199, opened 2023-07-21
Better support of iOS

justsmth:
### Issues:
Non-FIPS build is supported: #160

### Description of changes:
* Update "target" logic in sys-crate build scripts
* Better support of building/testing for iOS.
```
cargo build -p aws-lc-rs --features bindgen --target aarch64-apple-ios
```

### Call-outs:
* The FIPS build for iOS does not yet work.
* The CI performs a build iOS, but does not run the test.

### Testing:
* Tests can be run on an M1 Mac via the iphone simulator. The `DYLD_ROOT_PATH` must be set to an appropriate value for tests to run.
```
cargo test -p aws-lc-rs --target aarch64-apple-ios-sim
```


aws-lc-sys/builder/main.rs
```diff
@@ -178,6 +178,18 @@ fn prepare_cmake_build(manifest_dir: &PathBuf, build_prefix: Option<&str>) -> cm
...
 
+    if target_vendor() == "apple" {
+        if target_os().trim() == "ios" {
+            cmake_cfg.define("CMAKE_SYSTEM_NAME", "iOS");
+            if target().trim().ends_with("-ios-sim") {
+                cmake_cfg.define("CMAKE_OSX_SYSROOT", "iphonesimulator");
+            }
+        }
+        if target_arch().trim() == "aarch64" {
+            cmake_cfg.define("CMAKE_OSX_ARCHITECTURES", "arm64");
```
samuel40791765, line 189: Is the default x86_64 if this isn't set? Just curious if there could be differences in behavior, depending on the platform we cross-compiling from.

samuel40791765 (approved):

aws-lc-sys/builder/main.rs
justsmth: It must not be needed. We're not setting it currently and I have no trouble build on a x86 Mac.

skmcgrail (approved):

PR #202, opened 2023-07-21
Reduce rand test failure rate

justsmth:
### Issues:
N/A

### Description of changes: 
* The `test_secure_random_fill` test fails the CI more often than intended. This change reduces the sensitivity.

### Call-outs:
N/A

### Testing:
N/A

### DIFF
--- a/aws-lc-rs/src/rand.rs
+++ b/aws-lc-rs/src/rand.rs
@@ -166,29 +166,34 @@ pub fn fill(dest: &mut [u8]) -> Result<(), error::Unspecified> {
 mod tests {
     use crate::rand;
     use std::array::IntoIter;
+    use std::ops::Range;
 
     use crate::rand::{generate, SecureRandom, SystemRandom};
 
+    const RAND_ARY_SIZE: usize = 191;
+    const EXPECTED_MEAN_RANGE: Range<f64> = 101f64..155f64;
+    const EXPECTED_VARIANCE_LB: f64 = 8f64;
+
     #[test]
     fn test_secure_random_fill() {
-        let mut random_array = [0u8; 191];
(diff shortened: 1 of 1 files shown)


skmcgrail (approved):

skmcgrail (dismissed):

aws-lc-rs/src/rand.rs
```diff
@@ -166,40 +166,45 @@ pub fn fill(dest: &mut [u8]) -> Result<(), error::Unspecified> {
 mod tests {
     use crate::rand;
     use std::array::IntoIter;
+    use std::ops::Range;
 
     use crate::rand::{generate, SecureRandom, SystemRandom};
 
+    const RAND_ARY_SIZE: usize = 191;
```
dkostic, line 173: Nit: why not the whole word -- `ARRAY`?

aws-lc-rs/src/rand.rs
```diff
@@ -166,40 +166,45 @@ pub fn fill(dest: &mut [u8]) -> Result<(), error::Unspecified> {
...
+    const EXPECTED_MEAN_RANGE: Range<f64> = 101f64..155f64;
+    const EXPECTED_VARIANCE_LB: f64 = 8f64;
```
dkostic, line 175: where do these values come from?

justsmth: I can't really defend these specific numbers, but the range checked for is ~5 standard deviations above/below the mean.

Expected Mean: E(X) = (255-0)/2 = 127.5
Expected Variance: Var(X) = ((256^2)-1)/12 = 5461.25
Standard Deviation of n-bytes: (Var(X)/n)^(1/2) = (5461.25/191)^(1/2) = ~5.34

(What's called variance here is actually the standard deviation.)

PR #206, opened 2023-07-28
feat: add AEAD seal_in_place_scatter method

camshaft:
### Description of changes: 

The current AEAD encrypt API takes a `in_out: &mut [u8]`, which forces callers to copy all of the cleartext into a single slice. This shouldn't be needed since encrypting a payload _is_ copying the data since it has to XOR it with the cipher stream. 

Luckily, AWS-LC already provides a EVP API to account for this: [EVP_AEAD_CTX_seal_scatter](https://docs.rs/aws-lc-sys/latest/aws_lc_sys/fn.EVP_AEAD_CTX_seal_scatter.html). This allows the caller to pass a `extra_in`, which is encrypted into the `tag_out` buffer.

In transport protocols like TLS and QUIC, where the bulk of the data being encrypted is separate from the framing header, using this API can yield large wins. In the case of s2n-quic, we see a reduction in 8% of CPU by avoiding the copy.

#### API ergonomics

This new API isn't the most straightforward interface to use. You have to make sure to split the slices up in the correct order and account for the tag size. If we want to limit its use, we can add a feature flag stating that it's somewhat advanced.

#### API naming

I'm not sure if this is the best name, but it's what I came up with. Let me know if there's something better.

### Testing:

I've updated the AEAD tests to check the new API matches the existing ones. I'm using part of the key bytes to pick an arbitrary split point in the input for the `extra_in` parameter. This will make the tests deterministic for a given key but still be somewhat "random" in how the `in_out` is split from the `extra_in`.

#### s2n-quic without this change

Throughput is ~20Gbps. You can see the `memcpy` taking about 8% of CPU cycles in the flamegraph:

flamegraph

#### s2n-quic with this change

Throughput is ~21Gbps. You can see in the flamegraph that the `memcpy` is gone:

flamegraph

### DIFF
--- a/aws-lc-rs/src/aead.rs
+++ b/aws-lc-rs/src/aead.rs
@@ -811,13 +811,7 @@ where
     InOut: AsMut<[u8]> + for<'in_out> Extend<&'in_out u8>,
 {
     unsafe {
-        let aead_ctx = match key {
-            AeadCtx::AES_128_GCM(aead_ctx)
-            | AeadCtx::AES_256_GCM(aead_ctx)
-            | AeadCtx::AES_128_GCM_SIV(aead_ctx)
-            | AeadCtx::AES_256_GCM_SIV(aead_ctx)
-            | AeadCtx::CHACHA20_POLY1305(aead_ctx) => aead_ctx,
-        };
+        let aead_ctx = key.as_ref();
         let nonce = nonce.as_ref();
 
         let plaintext_len = in_out.as_mut().len();
@@ -856,13 +850,7 @@ pub(crate) fn aead_open_combined(
     in_out: &mut [u8],
 ) -> Result<(), Unspecified> {
     unsafe {
-        let aead_ctx = match key {
-            AeadCtx::AES_128_GCM(aead_ctx)
-            | AeadCtx::AES_256_GCM(aead_ctx)
-            | AeadCtx::AES_128_GCM_SIV(aead_ctx)
-            | AeadCtx::AES_256_GCM_SIV(aead_ctx)
-            | AeadCtx::CHACHA20_POLY1305(aead_ctx) => aead_ctx,
-        };
+        let aead_ctx = key.as_ref();
         let nonce = nonce.as_ref();
 
         let plaintext_len = in_out.len() - TAG_LEN;
--- a/aws-lc-rs/src/aead.rs
+++ b/aws-lc-rs/src/aead.rs
@@ -396,43 +396,6 @@ impl<N: NonceSequence> SealingKey<N> {
             in_out,
         )
     }
-
-    /// Encrypts and signs (“seals”) data in place with extra plaintext.
-    ///
-    /// `aad` is the additional authenticated data (AAD), if any. This is
-    /// authenticated but not encrypted. The type `A` could be a byte slice
-    /// `&[u8]`, a byte array `[u8; N]` for some constant `N`, `Vec<u8>`, etc.
-    /// If there is no AAD then use `Aad::empty()`.
-    ///
-    /// The plaintext is given as the input value of `in_out` and `extra_in`. `seal_in_place()`
-    /// will overwrite the plaintext contained in `in_out` with the ciphertext. The `extra_in` will
-    /// be encrypted into the `extra_out_and_tag`, along with the tag.
-    /// The `extra_out_and_tag` length must be equal to the `extra_len` and `self.algorithm.tag_len()`.
-    ///
-    /// # Errors
-    /// `error::Unspecified` when `nonce_sequence` cannot be advanced.
-    ///
-    #[inline]
-    #[allow(clippy::needless_pass_by_value)]
-    pub fn seal_in_place_scatter<A>(
-        &mut self,
-        aad: Aad<A>,
-        in_out: &mut [u8],
-        extra_in: &[u8],
-        extra_out_and_tag: &mut [u8],
-    ) -> Result<(), Unspecified>
-    where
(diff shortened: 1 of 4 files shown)


aws-lc-rs/src/aead.rs
```diff
@@ -396,6 +396,43 @@ impl<N: NonceSequence> SealingKey<N> {
...
+    ///
+    /// `aad` is the additional authenticated data (AAD), if any. This is
+    /// authenticated but not encrypted. The type `A` could be a byte slice
+    /// `&[u8]`, a byte array `[u8; N]` for some constant `N`, `Vec<u8>`, etc.
+    /// If there is no AAD then use `Aad::empty()`.
+    ///
+    /// The plaintext is given as the input value of `in_out` and `extra_in`. `seal_in_place()`
+    /// will overwrite the plaintext contained in `in_out` with the ciphertext. The `extra_in` will
+    /// be encrypted into the `extra_out_and_tag`, along with the tag.
+    /// The `extra_out_and_tag` length must be equal to the `extra_len` and `self.algorithm.tag_len()`.
```
justsmth, line 410: Technically, `extra_out_and_tag` can be longer than `extra_len + self.algorithm.tag_len()`, but only the bytes up to that length will be modified. So, the caller should ignore subsequent bytes.

aws-lc-rs/src/aead/aes_gcm.rs
```diff
@@ -51,6 +51,48 @@ pub(crate) fn aead_seal_separate(
...
+        };
+
+        let aad_slice = aad.as_ref();
+        let nonce = nonce.as_ref();
+        let mut out_tag_len = extra_out_and_tag.len();
+
+        if 1 != EVP_AEAD_CTX_seal_scatter(
+            aead_ctx,
+            in_out.as_mut_ptr(),
+            extra_out_and_tag.as_mut_ptr().cast(),
```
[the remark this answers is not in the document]
camshaft: Makes sense

aws-lc-rs/src/aead.rs
camshaft: Good question... Would it be better to just match what the existing API and allow for the larger messages? Doesn't matter to me.

justsmth: I'd prefer that we enforce the length constraint as documented to avoid potential client bugs/confusion/misuse. Such a constraint could potentially be removed in the future, but if we don't enforce it now, we'll never be able to enforce it.

camshaft: Sounds good. I'll make the change

justsmth (approved):

aws-lc-rs/src/aead.rs
```diff
@@ -631,6 +689,37 @@ impl LessSafeKey {
...
+    ///
+    #[inline]
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn seal_in_place_scatter<A>(
+        &self,
+        nonce: Nonce,
+        aad: Aad<A>,
+        in_out: &mut [u8],
+        extra_in: &[u8],
+        extra_out_and_tag: &mut [u8],
```
skmcgrail, line 708: I'm trying to think how we would best support this in the "FIPS" API for AES-GCM, see [`TLSRecordSealingKey`](https://github.com/aws/aws-lc-rs/pull/207). The AWS-LC API writes the randomized nonce to the end of the tag (thus the output tag is 12-bytes longer). This to me is an implementation detail of AWS-LC (likely having been done by the BoringSSL project to add the random nonce support within the existing API constraints), and not one I wish to force have aws-lc-rs users to have to think about. I'd like for our users to see those tags as standard sizes, and we provide the generated nonce as part of the output of the function.

I raise this issue because the `extra_in` API would require us to ask the user to size the `extra_out_and_tag` slice appropriately with the additional padding to write a nonce, which would leak this aspect of how the underlying API works. If we wanted to accept leaking this detail for this API, we could document the behavior, and require customers provided a slice that is `extra_in.len() + tag_len + 12 bytes`, but then I think the API should return a proper reference to the slice in the output result that is only `extra_in.len() + tag_len`. So something like `Result<&mut [u8], Unspecified>`. Otherwise if we absolutely don't want to expose this detail to the user then we are going to need to size an internal buffer and perform a memcpy, which may negate the benefits at that point, even if the buffer would be located on the stack.

aws-lc-rs/src/aead.rs
```diff
@@ -396,6 +396,43 @@ impl<N: NonceSequence> SealingKey<N> {
...
+    /// will overwrite the plaintext contained in `in_out` with the ciphertext. The `extra_in` will
+    /// be encrypted into the `extra_out_and_tag`, along with the tag.
+    /// The `extra_out_and_tag` length must be equal to the `extra_len` and `self.algorithm.tag_len()`.
+    ///
+    /// # Errors
+    /// `error::Unspecified` when `nonce_sequence` cannot be advanced.
+    ///
+    #[inline]
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn seal_in_place_scatter<A>(
```
skmcgrail, line 417: Had an offline conversation with Justin, so will summarize it here so that we can get this closed out. Let's remove `SealingKey::seal_in_place_scatter`, and only have support for this feature on `LessSafeKey`. With that small modification we are open to accepting this feature.

camshaft: Works for me!

justsmth (dismissed):

justsmth (approved):

PR #207, opened 2023-07-28
FIPS AEAD API Types

skmcgrail:
### Description of changes: 
Initial pass at proposed AEAD APIs that can be used in a FIPS strict compliance mode. Leaving this as draft as more test cases will needed to be added, and I wanted to seek feedback on the current API direction.

In particular I think `TLSSealingKey` and `TLSOpeningKey` require some conversation. Right now they both take the `Nonce` argument and rely on the user to construct the argument to meeting the protocol requirements for TLS1.2 or TLS1.3. AWS-LC just enforces that the nonce is incremented sequentially. We could offer wrapping types that handle the nonce calculation by giving the initialization vector on key construction, and having the user pass the current sequence number, and doing the once computation for the user. This would mean that we would be exposing TLS protocol implementation pieces which feels outside of our libraries intended scope. Additionally an extended argument to that would be to handle AAD construction along the same line.

### DIFF
--- a/aws-lc-rs/src/aead.rs
+++ b/aws-lc-rs/src/aead.rs
@@ -1041,7 +1041,7 @@ impl Algorithm {
     /// The length of the nonces.
     #[inline]
     #[must_use]
-    pub const fn nonce_len(&self) -> usize {
+    pub fn nonce_len(&self) -> usize {
         NONCE_LEN
     }
 }
--- a/aws-lc-rs/src/aead.rs
+++ b/aws-lc-rs/src/aead.rs
@@ -1570,8 +1570,8 @@ mod tests {
                         );
 
                         match (result, case.expect_err) {
-                            (Ok(_), true) => panic!("expected error for seal_in_place_append_tag"),
-                            (Ok(_), false) => {}
+                            (Ok(()), true) => panic!("expected error for seal_in_place_append_tag"),
+                            (Ok(()), false) => {}
                             (Err(_), true) => return,
                             (Err(e), false) => panic!("{e}"),
                         }
--- a/aws-lc-rs/src/aead.rs
+++ b/aws-lc-rs/src/aead.rs
@@ -14,9 +14,10 @@
 //!
 //! # Examples
 //! ```
+//! use aws_lc_rs::aead::{
+//!     nonce_sequence, Aad, BoundKey, OpeningKey, SealingKey, UnboundKey, AES_128_GCM,
+//! };
 //! use aws_lc_rs::rand;
-//! use aws_lc_rs::aead::{UnboundKey, AES_128_GCM, nonce_sequence, SealingKey,
-//!     OpeningKey, Aad, BoundKey};
 //! use aws_lc_rs::test::from_hex;
 //!
 //! let plaintext = "plaintext value";
@@ -41,7 +42,9 @@
 //!
 //! // Encrypt a value using the SealingKey
 //! let aad = Aad::from(aad_content);
-//! sealing_key.seal_in_place_append_tag(aad, &mut in_out_buffer).expect("Encryption failed");
+//! sealing_key
+//!     .seal_in_place_append_tag(aad, &mut in_out_buffer)
+//!     .expect("Encryption failed");
 //!
 //! // The buffer now contains the ciphertext followed by a "tag" value.
 //! let plaintext_len = in_out_buffer.len() - AES_128_GCM.tag_len();
@@ -55,20 +58,23 @@
 //!
 //! // Decrypt the value using the OpeningKey
 //! let aad = Aad::from(aad_content);
-//! opening_key.open_in_place(aad, &mut in_out_buffer).expect("Decryption failed");
+//! opening_key
+//!     .open_in_place(aad, &mut in_out_buffer)
+//!     .expect("Decryption failed");
 //!
 //! let decrypted_plaintext = std::str::from_utf8(&in_out_buffer[0..plaintext_len]).unwrap();
 //!
 //! assert_eq!(plaintext, decrypted_plaintext);
 //! ```
 
-use crate::{derive_debug_via_id, hkdf};
-use aes_gcm::aead_seal_separate;
-use std::fmt::Debug;
+use crate::{derive_debug_via_id, hkdf, iv::FixedLength};
+use std::{fmt::Debug, ptr::null, sync::Mutex};
 
 use crate::error::Unspecified;
 use aead_ctx::AeadCtx;
-use aws_lc::{EVP_AEAD_CTX_open, EVP_AEAD_CTX_seal};
+use aws_lc::{
+    EVP_AEAD_CTX_open, EVP_AEAD_CTX_open_gather, EVP_AEAD_CTX_seal, EVP_AEAD_CTX_seal_scatter,
+};
 use std::mem::MaybeUninit;
 use std::ops::RangeFrom;
 
@@ -162,7 +168,6 @@ impl<N: NonceSequence> OpeningKey<N> {
     /// # Errors
     /// `error::Unspecified` when ciphertext is invalid. In this case, `in_out` may have been
     /// overwritten in an unspecified way.
-    ///
     #[inline]
     pub fn open_in_place<'in_out, A>(
         &mut self,
@@ -220,7 +225,6 @@ impl<N: NonceSequence> OpeningKey<N> {
     /// # Errors
     /// `error::Unspecified` when ciphertext is invalid. In this case, `in_out` may have been
     /// overwritten in an unspecified way.
-    ///
     #[inline]
     pub fn open_within<'in_out, A>(
         &mut self,
@@ -231,8 +235,9 @@ impl<N: NonceSequence> OpeningKey<N> {
     where
         A: AsRef<[u8]>,
     {
-        open_within_(
-            &self.key,
+        open_within(
+            self.key.algorithm(),
+            self.key.get_inner_key(),
             self.nonce_sequence.advance()?,
             aad,
             in_out,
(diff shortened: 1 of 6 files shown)


aws-lc-rs/src/aead/aead_ctx.rs
```diff
@@ -71,24 +153,48 @@ impl AeadCtx {
...
+                tag_len,
                 null_mut(),
             ) {
                 return Err(Unspecified);
             }
             Ok(aead_ctx.assume_init())
         }
     }
+
+    pub(crate) fn as_mut_ptr(&mut self) -> &mut EVP_AEAD_CTX {
```
justsmth, line 165: It would be more idiomatic to implement `AsMut<EVP_AEAD_CTX>` and `AsRef<EVP_AEAD_CTX>`.

aws-lc-rs/src/aead/aes_gcm.rs
```diff
@@ -10,45 +11,89 @@ use crate::error::Unspecified;
 use aws_lc::EVP_AEAD_CTX_seal_scatter;
 use std::ptr::null;
 
+use super::{NONCE_LEN, TAG_LEN};
```
justsmth, line 14: Should we also move the items from the `use crate::aead::` line to here?

aws-lc-rs/src/aead/aes_gcm.rs
```diff
@@ -10,45 +11,89 @@ use crate::error::Unspecified;
...
-            | AeadCtx::AES_128_GCM(aead_ctx)
-            | AeadCtx::AES_256_GCM(aead_ctx) => aead_ctx,
-        };
-
-        let aad_slice = aad.as_ref();
+) -> Result<(Nonce, Tag), Unspecified> {
+    let aead_ctx = key.as_ptr();
+
+    let aad_slice = aad.as_ref();
+    let mut tag = [0u8; TAG_LEN];
```
justsmth, line 27: * Should we be consistent in using `MAX_TAG_LEN` for the size of such arrays? 
* `MAX_TAG_LEN` and `TAG_LEN` are currently the same value (and might always be). `MAX_TAG_LEN` is the value we expose publicly -- should we get rid of `TAG_LEN` until we have the need for some tag length constant that's a different value.

aws-lc-rs/src/aead.rs
```diff
@@ -647,6 +678,317 @@ impl Debug for LessSafeKey {
...
+
+    /// The key's AEAD algorithm.
+    #[inline]
+    #[must_use]
+    pub fn algorithm(&self) -> &'static Algorithm {
+        self.algorithm
+    }
+}
+
+/// AEAD Encryption key used for TLS porotocol record encryption.
```
justsmth, line 899: "porotocol"

aws-lc-rs/src/aead/aead_ctx.rs
camshaft: Those would be public impls, though

aws-lc-rs/src/aead.rs
camshaft: https://github.com/aws/s2n-quic/blob/8ef0a6b66a856dc9f34ce18159c617ac29154cc7/.github/workflows/ci.yml#L671-L684

aws-lc-rs/src/aead.rs
```diff
@@ -647,6 +678,317 @@ impl Debug for LessSafeKey {
...
+
+    /// TLS 1.3 (RFC 8446)
+    TLS13,
+}
+
+/// AEAD Encryption key used for TLS porotocol record encryption.
+pub struct TLSRecordSealingKey {
+    ctx: Mutex<AeadCtx>,
+    algorithm: &'static Algorithm,
+}
```
justsmth, line 829: We need some documentation as to why the Mutex is needed. (Is this even documented in AWS-LC?)

aws-lc-rs/src/aead.rs
```diff
@@ -647,6 +678,317 @@ impl Debug for LessSafeKey {
...
+    /// # Errors
+    /// `error::Unspecified` if encryption operation fails.
+    #[inline]
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn seal_in_place_append_tag<A, InOut>(
+        &self,
+        nonce: Nonce,
+        aad: Aad<A>,
+        in_out: &mut InOut,
+    ) -> Result<(), Unspecified>
```
justsmth, line 875: For TLS, isn't the AAD length expected to be 11 bytes?  Do we enforce that anywhere in our logic? Do we need a separate more-restrictive `Aad` type?

aws-lc-rs/src/aead/aead_ctx.rs
justsmth: It would be implemented on `AeadCtx` which is only `pub(crate)` so it should be ok.(?)

skmcgrail: Ah this is my mad I had originally meant this to be similar to DigestCtx where these just return the raw pointers. Will take a look again.

aws-lc-rs/src/aead/aes_gcm.rs
skmcgrail: `NONCE_LEN` is actually in `crate::nonce` and will need to remain exported via the top-level `crate::aead` module so that is a bit of a wash. We could move `TAG_LEN` here but its not really `aes_gcm` specific since the same tag length is used with chacha.

aws-lc-rs/src/aead.rs
skmcgrail: I don't see AWS-LC enforcing this at the cipher usage? Nor do I see where you got 11 bytes from? In AWS-LC's AAD construction in libssl specifically it [appears](https://github.com/aws/aws-lc/blob/b84674b2d2512cfd200b7c392215b6eebda3c59a/ssl/ssl_aead_ctx.cc#L263-L264) to allocate 13 bytes for AAD as the upper bound?

But again this doesn't appear to be enforced at the cipher usage level.  It also seems to depend whether the AAD is expected to be just the header or if it includes other information (seems to depend on TLS 1.2 versus TLS 1.3).

Similar to nonce construction should we lave the proper aad construction left to the implementer? If we know what the proper max limit should be we could maybe sanity check it, but if it is variable I don't think we can enforce an exact size.

aws-lc-rs/src/aead/aes_gcm.rs
skmcgrail: I took a stab at doing a broader refactor to try to abstract out the hard-coding of TAG_LEN / NONCE_LEN where possible, and utilize the `tag_len()` and `nonce_len()` from the configured algorithm. That will be in the next revision I push.

aws-lc-rs/src/aead.rs
justsmth: Ok. This comment and the one above were based on my reading of documentation provided by BoringSSL here: https://commondatastorage.googleapis.com/chromium-boringssl-docs/aead.h.html
> These AEAD primitives do not meet the definition of generic AEADs. They are all specific to TLS and should not be used outside of that context. They must be initialized with [EVP_AEAD_CTX_init_with_direction](https://commondatastorage.googleapis.com/chromium-boringssl-docs/aead.h.html#EVP_AEAD_CTX_init_with_direction), are stateful, and may not be used concurrently. Any nonces are used as IVs, so they must be unpredictable. They only accept an ad parameter of length 11 (the standard TLS one with length omitted).

I'll look into it some more.

justsmth (approved):

justsmth (approved):

aws-lc-rs/src/aead.rs
```diff
@@ -739,86 +1113,279 @@ fn check_per_nonce_max_bytes(alg: &Algorithm, in_out_len: usize) -> Result<(), U
...
+
+    debug_assert!(alg_tag_len <= MAX_TAG_LEN);
+
+    let tag_buffer = [0u8; MAX_TAG_LEN];
+
+    in_out.extend(tag_buffer[..alg_tag_len].iter());
+
+    let mut out_len = MaybeUninit::<usize>::uninit();
+    let mut_in_out = in_out.as_mut();
+    let add_str = aad.0;
```
samuel40791765, line 1137: Nit, also in `aead_seal_combined_randnonce`
```suggestion
    let aad_str = aad.0;
```

aws-lc-rs/src/aead.rs
```diff
@@ -739,86 +1113,279 @@ fn check_per_nonce_max_bytes(alg: &Algorithm, in_out_len: usize) -> Result<(), U
...
+    let alg_tag_len = alg.tag_len();
+
+    debug_assert!(alg_tag_len <= MAX_TAG_LEN);
+
+    let tag_buffer = [0u8; MAX_TAG_LEN];
+
+    in_out.extend(tag_buffer[..alg_tag_len].iter());
+
+    let mut out_len = MaybeUninit::<usize>::uninit();
+    let mut_in_out = in_out.as_mut();
```
samuel40791765, line 1136: Is this additional mutation needed? Or is this needed due to the `extend` we do on `in_out` above?

aws-lc-rs/src/aead.rs
```diff
@@ -876,4 +1443,365 @@ mod tests {
...
+                    &mut in_out,
+                )
+                .unwrap();
+
+            assert_eq!(plaintext, in_out[..plaintext.len()]);
+        }
+    }
+
+    #[test]
+    fn test_aes_256_tls13() {
```
samuel40791765, line 1739: There seems to be some shared code among the test functions, would it be worth abstracting them out into another function for reuse?

aws-lc-rs/src/aead.rs
skmcgrail: This is needed as `InOut` is a generic parameter that implements `AsMut<[u8]>` to return a mutable reference to a slice. So we have to invoke the `as_mut()` method to get a reference to the slice for passing across the FFI boundary.

aws-lc-rs/src/aead.rs
```diff
@@ -678,7 +1041,7 @@ impl Algorithm {
     /// The length of the nonces.
     #[inline]
     #[must_use]
-    pub fn nonce_len(&self) -> usize {
+    pub const fn nonce_len(&self) -> usize {
```
justsmth, line 1044: By making this a `const` function, we're prevented from ever having it return something other than a `const` value (unless we break backwards compatibility). This could prevent us from supporting certain algorithms through this API in the future.

aws-lc-rs/src/aead/aead_ctx.rs
```diff
@@ -20,58 +39,223 @@ use aws_lc::{
...
+            AeadCtx::AES_128_GCM(ctx)
+            | AeadCtx::AES_256_GCM(ctx)
+            | AeadCtx::AES_128_GCM_RANDNONCE(ctx)
+            | AeadCtx::AES_256_GCM_RANDNONCE(ctx)
+            | AeadCtx::AES_128_GCM_TLS12(ctx)
+            | AeadCtx::AES_256_GCM_TLS12(ctx)
+            | AeadCtx::AES_128_GCM_TLS13(ctx)
+            | AeadCtx::AES_256_GCM_TLS13(ctx)
+            | AeadCtx::CHACHA20_POLY1305(ctx) => **ctx,
+        }
```
justsmth, line 259: Weren't these changed to `AsRef` and `AsMutRef` impls?

skmcgrail: I wasn't sure if it would just make it more awkward doing something like `AsRef<LcPtr<*mut EVP_AEAD_CTX>>`, instead of this. Originally it was `AsRef<EVP_AEAD_CTX>` but that was before we migrated this type to the heap. Now that it's being allocated by AWS-LC I'm not sure if we want to / can make assumptions about the alignment of the raw pointer, thus I don't think we should be trying to cast the `*mut EVP_AEAD_CTX` stored in `LcPtr<_>` to a `&EVP_AEAD_CTX`. Anyways I went with the awkward approach for now?

justsmth (approved):

justsmth (approved):

samuel40791765 (approved):

PR #211, opened 2023-08-03
Refactor LcPtr, DetachablePtr for readability

skmcgrail:
### Description of changes: 
Just cleans up the usage so they you don't need to specify `LcPtr<*mut T>` or `DetachableLcPtr<*mut T>` but can just say `LcPtr<T>` or `DetacableLcPtr<T>` similar to how `ConstPtr<T>` is already how that type functions.

### DIFF
--- a/aws-lc-rs/src/ptr.rs
+++ b/aws-lc-rs/src/ptr.rs
@@ -41,9 +41,9 @@ impl<P: Pointer> Drop for ManagedPointer<P> {
     }
 }
 
-impl<P: Pointer + Copy> ManagedPointer<P> {
+impl<P: Pointer> ManagedPointer<P> {
     #[inline]
-    pub fn as_const<T>(&self) -> ConstPointer<T> {
+    pub fn as_const(&self) -> ConstPointer<P::T> {
         ConstPointer {
             ptr: self.pointer.as_const_ptr(),
         }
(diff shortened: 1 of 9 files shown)


aws-lc-rs/src/ptr.rs
```diff
@@ -31,14 +34,14 @@ impl<P: Pointer> LcPtr<P> {
...
-impl<P: Pointer> Drop for LcPtr<P> {
+impl<P: Pointer> Drop for ManagedPointer<P> {
     #[inline]
     fn drop(&mut self) {
         self.pointer.free();
     }
 }
 
-impl<P: Pointer + Copy> LcPtr<P> {
+impl<P: Pointer + Copy> ManagedPointer<P> {
```
justsmth, line 44: Why do we need `P` to be `Copy`? This might be an artifact of some method we wrote that has since been removed.(?)

aws-lc-rs/src/ptr.rs
```diff
@@ -85,7 +88,7 @@ impl<P: Pointer> DetachableLcPtr<P> {
     }
 }
 
-impl<P: Pointer + Copy> DetachableLcPtr<P> {
+impl<P: Pointer + Copy> DetachablePointer<P> {
     #[inline]
     pub fn as_const<T>(&self) -> ConstPointer<T> {
```
justsmth, line 93: This has been bothering me -- `T` should be an associated type for the Pointer instead of a generic for this function.

justsmth (changes requested):

samuel40791765 (approved):

PR #216, opened 2023-08-15
FIPS Status Indicator

skmcgrail:
### Description of changes: 

Will provide a better description once I move it out of draft.

### Call-outs:

- [x] Need to investigate HKDF API in more detail, we more then likely still need to make some API changes here for FIPS compliance. https://github.com/aws/aws-lc-rs/pull/217
- [x] ~I'd like to add documentation to methods that are known to result in changes to the FIPS indicator. This is documented by test cases I've started plumbing into `aws-lc-rs-testing` crate, but I want this to be more discoverable for customers.~ (Will do a follow-up PR)


aws-lc-rs/src/fips.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+    #[cfg(not(feature = "fips"))]
+    FipsServiceStatus::NotEnabled(())
+}
+
+/// Clears the FIPS module service status. Useful if a thread has executed an unapproved
+/// algorithm, but you desire to reset the module status without destroying the thread.
+#[cfg(feature = "fips")]
+pub fn reset_fips_service_status() {
+    indicator::reset();
```
torben-hansen, line 23: There could be a "race" on this.

Not a "race" in the typical definition. But an application that is build with two different libraries that both use aws-lc-rs. One resets, while the other relies on an error being thrown later.

aws-lc-rs/src/fips.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+#[cfg(feature = "fips")]
+pub(crate) mod indicator {
+    use std::cell::Cell;
+
+    thread_local! {
+        static APPROVED_OPERATING_MODE: Cell<bool> = const { Cell::new(true)};
```
torben-hansen, line 31: does this mean that at initialisation time, the service indicator is set to "approved"? Can't figure out the Rust :)

aws-lc-rs/src/ed25519.rs
```diff
@@ -58,17 +58,19 @@ impl VerificationAlgorithm for EdDSAParameters {
...
-            ) {
-                return Err(Unspecified);
-            }
-            Ok(())
+            )
+        } {
+            return Err(Unspecified);
         }
+        #[cfg(feature = "fips")]
+        crate::fips::indicator::set_unapproved();
```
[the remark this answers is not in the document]
camshaft: you could do it in aws-lc itself, that way the signal would at least be available to C applications

aws-lc-rs/src/cipher/aes.rs
```diff
@@ -24,6 +24,9 @@ pub(crate) fn encrypt_block_aes(aes_key: &AES_KEY, block: Block) -> Block {
             aes_key,
         );
 
+        #[cfg(feature = "fips")]
```
camshaft, line 27: It might be cleaner to make the `set_unapproved` function a noop with a `#[inline(always)]` when `not(feature = "fips")`

aws-lc-rs/src/lib.rs
```diff
@@ -124,10 +124,49 @@ mod ec;
...
+
+pub use crate::fips::{get_fips_service_status, FipsServiceStatus};
+
+/// Use `check_fips_service_status` to wrap an **aws-lc-rs** cryptographic function, and verify
+/// that the operation was approved by the FIPS module. By default this macro will reset the FIPS module
+/// status if the operation was not approved. You must use `FipsServiceStatus` to determine if
+/// the operation was not approved. If the crate was not built with the `fips` feature, this function will
+/// always wrap the function result with `FipsServiceStatus::NotEnabled`.
+#[macro_export]
+macro_rules! check_fips_service_status {
```
camshaft, line 143: Rather than a macro it might be a bit more ergonomic to do a FIPS "guard" which triggers a check when it goes out of scope:

```
{
    let _fips_guard = aws_lc_rs::fips::guard();

    // do crypto here

    // fips::Guard Drop impl is called, which will check the FIPS status
}
```

aws-lc-rs/src/fips.rs
camshaft: Another thing to consider is nested scopes. If you wrap a `check_fips_service_status` inside of another `check_fips_service_status`, do you end up with a weird state?

skmcgrail:
I've pushed a significant update that basically retracts making the FIPS indicator public, and only make it internally accessible when built with debug builds (`debug_assertions` compiler option enabled). This allows us to have unit-test assertions demonstrating that the API we expect to be FIPS compliant (aka use the aws-lc FIPS crypto approved functions) are continually validated by our CI, but when compiled as a release build the indicator and any runtime costs will not be present.

We can revisit in the future if we want to expose the FIPS indicator in a publicly consumable fashion, but for now we can use by upcoming API documentation PR that will capture what is considered "compliant".

aws-lc-rs/src/aead/poly1305.rs
```diff
@@ -65,6 +65,7 @@ impl Context {
         unsafe {
             let mut tag = MaybeUninit::<[u8; TAG_LEN]>::uninit();
             CRYPTO_poly1305_finish(self.state.0.as_mut_ptr().cast(), tag.as_mut_ptr().cast());
+            crate::fips::set_fips_service_status_unapproved();
```
justsmth, line 68: The pattern seems to be to only set the "unapproved" indicator after the operation has successfully completed.  Would it be more "friendly" to (also) set it at at the start (in `from_key`) instead of waiting until they've completed the operation?

aws-lc-rs/src/cipher/aes.rs
```diff
@@ -24,6 +24,8 @@ pub(crate) fn encrypt_block_aes(aes_key: &AES_KEY, block: Block) -> Block {
             aes_key,
         );
 
+        crate::fips::set_fips_service_status_unapproved();
```
justsmth, line 27: Would using [AES_ecb_encrypt](https://github.com/aws/aws-lc/blob/aa90e509f2e940916fbe9fdd469a4c90c51824f6/crypto/fipsmodule/aes/mode_wrappers.c#L84) instead of `AES_encrypt` cause this to become FIPS approved?

aws-lc-rs/src/cipher/chacha.rs
```diff
@@ -52,6 +52,8 @@ pub(crate) fn encrypt_block_chacha20(
         counter,
     )?;
 
+    crate::fips::set_fips_service_status_unapproved();
```
justsmth, line 55: As I mentioned elsewhere, should we set "unapproved" even when the operation fails?

justsmth:
In rsa.rs, there's a call to [`RSA_check_fips`](https://github.com/aws/aws-lc/blob/aa90e509f2e940916fbe9fdd469a4c90c51824f6/crypto/fipsmodule/rsa/rsa.c#L966C12-L966C12) that I think should have an indicator check.

aws-lc-rs/src/lib.rs
```diff
@@ -148,18 +149,21 @@ pub fn init() {
 }
 
 #[cfg(feature = "fips")]
+#[deprecated = "See `crate::get_fips_service_status`"]
```
justsmth, line 152: `get_fips_service_status` is only exposed as `pub(crate)` within the `fips` module. It's also only built when the "fips" feature is enabled.  What's the plan for exposing this function?

aws-lc-rs/src/aead/poly1305.rs
skmcgrail: I was opting to be similar in consistency to AWS-LC. So for example the indicator is set at the end of functions only when they are successful. Similarly for the indicators are I manually plumbed through to set to false, I opted to do it in similar spots where the indicator was set in AWS-LC for similar functionality.

aws-lc-rs/src/cipher/aes.rs
skmcgrail: Ah good catch, I checked and AES_ecb_encrypt does set the the indicator, will update the code to use that function.

aws-lc-rs/src/cipher/chacha.rs
skmcgrail: See earlier response.

skmcgrail:
> In rsa.rs, there's a call to [`RSA_check_fips`](https://github.com/aws/aws-lc/blob/aa90e509f2e940916fbe9fdd469a4c90c51824f6/crypto/fipsmodule/rsa/rsa.c#L966C12-L966C12) that I think should have an indicator check.

This doesn't appear to have an explicit indicator check associated with it from the documentation I've reviewed.

justsmth (approved):

torben-hansen (approved):

PR #219, opened 2023-08-18
Make defined aead algorithms const instead of static

theli-ua:
There seem to be no reason for them to be static

### Description of changes: 
Makes defined aead algorithms const instead of static. There seem to be no reason for them to be static
### Call-outs:
This cannot  break anyone - const are assignable to statics (if someone refers them in statics) but not the other way
### Testing:
existing tests


justsmth:
`CHACHA20_POLY1305` would also need to be changed to a `const`.

I'm fine with this change as long as it can't break clients, but I still need to convince myself of this.

theli-ua:
> but I still need to convince myself of this.

Well, const can be used in static context, static can't be used in const context.
The only way this can break anything is when that something would try to unsafely modify those statics.

justsmth:
> > but I still need to convince myself of this.
> 
> Well, const can be used in static context, static can't be used in const context. The only way this can break anything is when that something would try to unsafely modify those statics.

[This RFC](https://github.com/rust-lang/rfcs/blob/master/text/1414-rvalue_static_promotion.md#detailed-design) eliminates much of my concern. The only way I can see it breaking client code would be if a client was doing "pointer comparisons" between values initialized separately. We [implement `Eq`](https://docs.rs/aws-lc-rs/1.2.1/aws_lc_rs/aead/struct.Algorithm.html#impl-Eq-for-Algorithm) on the struct so normal comparisons would be fine.

This other concern is the likely increase the size of binaries. Are the performance benefits enough to offset this?

theli-ua:
>Are the performance benefits enough to offset this?

Actually usecase that triggered this wasn't performance - its being able to use it in constant contexts - in my source I would like to assign these to constants (eg "this is my const for envelope encryption, this is my const defining mode for data encryption, etc)

However, could you elaborate on binary size increase please?

justsmth (approved):

skmcgrail (approved):

PR #220, opened 2023-08-21
Make docker configs consistent

justsmth:
### Issues:
N/A

### Description of changes: 
* The "sparse-registry" config in arm64 image is [no longer needed](https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html#sparse-by-default-for-cratesio).
* Docker images were inconsistent at to whether:
    * to install nightly
    * to set CC/CXX

### Call-outs:
N/A

### Testing:
Using for sys crate updates.

### DIFF
--- a/docker/linux-386/Dockerfile
+++ b/docker/linux-386/Dockerfile
@@ -10,6 +10,7 @@ ARG GO_VERSION=go1.20.1
 RUN apt-get update && \
     apt-get -y install clang cmake golang && \
     rustup update && \
+    rustup toolchain install nightly && \
     rustup component add rustfmt clippy && \
     cargo install rust-script && \
     git config --global --add safe.directory '*' && \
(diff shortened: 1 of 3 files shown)


skmcgrail (approved):

docker/linux-386/Dockerfile
```diff
@@ -22,3 +23,7 @@ RUN apt-get update && \
     rm -rf /tmp/*
 
 ENV PATH="/root/sdk/${GO_VERSION}/bin:${PATH}"
+
+# Static FIPS build only supports clang.
```
torben-hansen, line 27: Is that accurate? Or some special requirement for aws-lc-rs?

For example: https://github.com/aws/aws-lc/blob/main/tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml#L58-L66
Same for Arm dimensions.

samuel40791765: We can remove this comment now :). We hadn't fixed the FIPS build for all compilers at the time.

skmcgrail (approved):

PR #224, opened 2023-08-23
Migrate Ed25519 key generation to EVP_PKEY_keygen

skmcgrail:
### Description of changes: 
To be consistent with other key generation strategies introduced on the fips-api-changes branch, migrate Ed25519 key-generation, and remove support for using the SecureRandom trait.

### DIFF
--- a/aws-lc-rs/src/ed25519.rs
+++ b/aws-lc-rs/src/ed25519.rs
@@ -4,14 +4,16 @@
 // SPDX-License-Identifier: Apache-2.0 OR ISC
 
 use crate::error::{KeyRejected, Unspecified};
+use crate::fips::indicator_check;
 use crate::pkcs8::{Document, Version};
 use crate::ptr::LcPtr;
 use crate::rand::SecureRandom;
 use crate::signature::{KeyPair, Signature, VerificationAlgorithm};
 use crate::{constant_time, sealed, test};
 use aws_lc::{
-    ED25519_keypair_from_seed, ED25519_sign, ED25519_verify, EVP_PKEY_get_raw_private_key,
-    EVP_PKEY_get_raw_public_key, EVP_PKEY_new_raw_private_key, EVP_PKEY, EVP_PKEY_ED25519,
+    ED25519_keypair_from_seed, ED25519_sign, ED25519_verify, EVP_PKEY_CTX_new_id,
+    EVP_PKEY_get_raw_private_key, EVP_PKEY_get_raw_public_key, EVP_PKEY_keygen,
+    EVP_PKEY_keygen_init, EVP_PKEY, EVP_PKEY_ED25519,
 };
 use std::fmt;
 use std::fmt::{Debug, Formatter};
@@ -20,7 +22,6 @@ use std::ptr::null_mut;
 
 #[cfg(feature = "ring-sig-verify")]
 use untrusted::Input;
-use zeroize::Zeroize;
 
 /// The length of an Ed25519 public key.
 pub const ED25519_PUBLIC_KEY_LEN: usize = aws_lc::ED25519_PUBLIC_KEY_LEN as usize;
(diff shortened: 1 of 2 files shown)


justsmth (approved):

skmcgrail:
Addressed merge conflict

aws-lc-rs/src/ed25519.rs
```diff
@@ -116,31 +117,22 @@ impl KeyPair for Ed25519KeyPair {
...
+        return Err(());
+    }
+
+    let mut pkey = null_mut::<EVP_PKEY>();
+
+    if 1 != indicator_check!(unsafe { EVP_PKEY_keygen(*pkey_ctx, &mut pkey) }) {
+        return Err(());
+    }
+
+    let pkey = LcPtr::new(pkey)?;
```
torben-hansen, line 133: How does one produce a key that can be used as an input to the sign function (`pub fn sign(&self, msg: &[u8])`) in this file?

My understanding is that you first have to write the EVP_PKEY object to file in a PKCS format(?).

aws-lc-rs/src/ed25519.rs
```diff
@@ -116,31 +117,22 @@ impl KeyPair for Ed25519KeyPair {
...
+pub(crate) unsafe fn generate_key() -> Result<LcPtr<EVP_PKEY>, ()> {
+    let pkey_ctx = LcPtr::new(unsafe { EVP_PKEY_CTX_new_id(EVP_PKEY_ED25519, null_mut()) })?;
+
+    if 1 != unsafe { EVP_PKEY_keygen_init(*pkey_ctx) } {
+        return Err(());
+    }
+
+    let mut pkey = null_mut::<EVP_PKEY>();
+
+    if 1 != indicator_check!(unsafe { EVP_PKEY_keygen(*pkey_ctx, &mut pkey) }) {
```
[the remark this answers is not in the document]
skmcgrail: Not expected to set the indicator, but since `EVP_PKEY_keygen` would set the indicator if it were approved, it seemed like it made sense to just put it here. That way I can have a test that validates that this is not an approved API explicitly.

Where before, I was just manually toggling our internal indicator state in the code for removal.

aws-lc-rs/src/ed25519.rs
skmcgrail: Yeah that is the way the API works currently (an artifact of how the Ring API was designed). The user invokes [`generate_pkcs8`](https://docs.rs/aws-lc-rs/1.3.0/aws_lc_rs/signature/struct.Ed25519KeyPair.html#method.generate_pkcs8) to get a PKCS#8 document back, which they then use [`from_pkcs8`](https://docs.rs/aws-lc-rs/1.3.0/aws_lc_rs/signature/struct.Ed25519KeyPair.html#method.from_pkcs8) to then construct the key. It's a bit silly, and definitely a feature I would like to approve on in the future.

aws-lc-rs/src/ed25519.rs
```diff
@@ -116,31 +117,22 @@ impl KeyPair for Ed25519KeyPair {
...
-    crate::fips::set_fips_service_status_unapproved();
-
-    LcPtr::new(EVP_PKEY_new_raw_private_key(
-        EVP_PKEY_ED25519,
-        null_mut(),
-        private_key.assume_init().as_ptr(),
-        ED25519_PRIVATE_KEY_SEED_LEN,
-    ))
+pub(crate) unsafe fn generate_key() -> Result<LcPtr<EVP_PKEY>, ()> {
+    let pkey_ctx = LcPtr::new(unsafe { EVP_PKEY_CTX_new_id(EVP_PKEY_ED25519, null_mut()) })?;
```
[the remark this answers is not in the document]
skmcgrail: I don't have context on the original decision, but majority of the library's code has opted not to define an internal veneer around these API types, to then just be further wrapped by Rust code. I presume this was done to not bake in shared assumptions, or to avoid potential for performance compromises. I'm not entirely sure I have a good answer otherwise.

aws-lc-rs/src/ed25519.rs
torben-hansen: would be great if that is possible one day.

torben-hansen (approved):

PR #232, opened 2023-09-06
TLS 1.2 PRF Support

skmcgrail:

### DIFF
--- a/aws-lc-rs/src/lib.rs
+++ b/aws-lc-rs/src/lib.rs
@@ -127,6 +127,7 @@ mod fips;
 pub mod iv;
 mod ptr;
 mod rsa;
+pub mod tls_prf;
 
 pub(crate) use debug::derive_debug_via_id;
 
--- a/aws-lc-rs/src/tls_prf.rs
+++ b/aws-lc-rs/src/tls_prf.rs
@@ -1,0 +1,333 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+//! TLS 1.2 PRF API's for usage in [RFC 5246](https://www.rfc-editor.org/rfc/rfc5246) and [RFC 7627](https://www.rfc-editor.org/rfc/rfc7627).
+
+use crate::{
+    digest::match_digest_type, digest::AlgorithmID, error::Unspecified, fips::indicator_check,
+};
+use core::ptr::null;
+
+use aws_lc::CRYPTO_tls1_prf;
+
+/// The TLS PRF `P_hash` Algorithm
+pub struct Algorithm(AlgorithmID);
+
+/// SHA-256 `P_hash` algorithm
+pub const P_SHA256: Algorithm = Algorithm(AlgorithmID::SHA256);
+
+/// SHA-384 `P_hash` algorithm
+pub const P_SHA384: Algorithm = Algorithm(AlgorithmID::SHA384);
+
+/// SHA-512 `P_hash` algorithm
+pub const P_SHA512: Algorithm = Algorithm(AlgorithmID::SHA512);
+
+/// Encapsulates a PRF algorithm and secret bytes to be used to derive output.
+pub struct Secret {
+    algorithm: &'static Algorithm,
+    secret: Box<[u8]>,
+}
+
+impl Secret {
+    /// Constructs a new `Secret` for use with the associated `P_hash` `Algorithm`.
+    ///
+    /// # Errors
+    /// * `Unspecified`: If `secret.is_empty() == true`.
+    pub fn new(algorithm: &'static Algorithm, secret: &[u8]) -> Result<Self, Unspecified> {
+        if secret.is_empty() {
+            return Err(Unspecified);
+        }
+
+        let secret = Vec::from(secret).into_boxed_slice();
+
+        Ok(Self { algorithm, secret })
+    }
+
+    /// Calculates `len` bytes of TLS PRF using the configured [`Algorithm`], and returns [`Secret`] of length `len`.
+    /// See [RFC5246](https://datatracker.ietf.org/doc/html/rfc5246#section-5)
+    ///
+    /// # Errors
+    /// * `Unspecified`: Returned if the PRF derivation fails.
+    pub fn derive(self, label: &[u8], seed: &[u8], output: usize) -> Result<Secret, Unspecified> {
+        prf(self.algorithm, &self.secret, label, seed, None, output)
+    }
+
+    /// Calculates `len` bytes of TLS PRF using the configured [`Algorithm`], and returns [`Secret`] of length `len`.
+    ///
+    /// In this function calling convention `seed1` and `seed2` will be concatenated together: `seed1 || seed2`.
+    ///
+    /// See [RFC5246](https://datatracker.ietf.org/doc/html/rfc5246#section-5)
+    ///
+    /// # Errors
+    /// * `Unspecified`: Returned if the PRF derivation fails.
+    pub fn derive_with_seed_concatination(
+        self,
+        label: &[u8],
+        seed1: &[u8],
+        seed2: &[u8],
+        len: usize,
+    ) -> Result<Secret, Unspecified> {
+        prf(self.algorithm, &self.secret, label, seed1, Some(seed2), len)
+    }
+}
+
+impl Drop for Secret {
+    fn drop(&mut self) {
+        use zeroize::Zeroize;
+        self.secret.zeroize();
+    }
+}
+
+impl AsRef<[u8]> for Secret {
+    fn as_ref(&self) -> &[u8] {
+        &self.secret
+    }
+}
+
+impl<const L: usize> TryFrom<Secret> for [u8; L] {
+    type Error = Unspecified;
+
+    fn try_from(value: Secret) -> Result<Self, Self::Error> {
+        if value.secret.len() != L {
+            return Err(Unspecified);
+        }
+
+        let mut ret = [0u8; L];
+        ret.copy_from_slice(&value.secret);
+
+        Ok(ret)
+    }
+}
+
+fn prf(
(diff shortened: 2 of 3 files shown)


aws-lc-fips-sys/include/rust_wrapper.h
```diff
@@ -106,4 +106,12 @@ int ERR_GET_FUNC_RUST(uint32_t packed_error);
 #include "openssl/ssl3.h"
 #endif
 
+// The TLS PRF API is exported in AWS-LC main branch, but in the current FIPS branch, this API is defined in a private header.
+int CRYPTO_tls1_prf(const EVP_MD *digest,
```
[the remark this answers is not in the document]
skmcgrail: Bindings need to be regenerated, just hadn't done that yet as I had been waiting for the most recent release to avoid merge conflicts. Otherwise this will just work as normal. Please provide feedback on the TLS PRF API at least and don't block on this in particular.

aws-lc-rs/src/lib.rs
```diff
@@ -128,6 +128,7 @@ mod fips;
 pub mod iv;
 mod ptr;
 mod rsa;
+pub mod tlsprf;
```
samuel40791765, line 131: Minor nit: Should we do `tls_prf` or `tlsprf`?
```suggestion
pub mod tls_prf; 
```

aws-lc-rs/src/tlsprf/tests/fips.rs
```diff
@@ -0,0 +1,64 @@
...
+#![cfg(debug_assertions)]
+
+use crate::fips::{assert_fips_status_indicator, FipsServiceStatus};
+use crate::tlsprf::{Secret, P_SHA256, P_SHA384, P_SHA512};
+
+macro_rules! prf {
+    ($name:ident, $alg:expr, $size:expr, $label:expr, $expect:path) => {
+        #[test]
+        fn $name() {
+            let secret = Secret::new($alg, &[42u8; $size]).expect("secret created");
```
samuel40791765, line 13: Is there any particular reason why `42u8` is used here?

aws-lc-rs/src/tlsprf.rs
```diff
@@ -0,0 +1,295 @@
...
+            0x5e, 0x66, 0x12, 0x51, 0x88, 0xbf,
+        ];
+
+        let mut output = vec![0u8; EXPECT.len()];
+
+        let secret = Secret::new(P_SHA256, SECRET).expect("secret created");
+
+        secret
+            .derive_with_seed_concatination(LABEL, SEED1, SEED2, &mut output)
+            .expect("dervice successful");
```
samuel40791765, line 213:
```suggestion
            .expect("derive successful");
```
Some others below as well.

aws-lc-rs/src/tlsprf.rs
```diff
@@ -0,0 +1,295 @@
...
+            0x32, 0x81,
+        ];
+
+        const TLS_OUTPUT_P_SHA256: &[u8; 32] = &[
+            0xe2, 0x1d, 0xd6, 0xc2, 0x68, 0xc7, 0x57, 0x03, 0x2c, 0x2c, 0xeb, 0xbb, 0xb8, 0xa9,
+            0x7d, 0xe9, 0xee, 0xe6, 0xc9, 0x47, 0x83, 0x0a, 0xbd, 0x11, 0x60, 0x5d, 0xd5, 0x2c,
+            0x47, 0xb6, 0x05, 0x88,
+        ];
+
+        let secret = super::Secret::new(super::P_SHA256, TLS_SECRET).expect("secret created");
```
samuel40791765, line 147: Nit: `super::` here and below might be redundant

samuel40791765 (commented):
Just some minor nits and questions, but the API design looks good to me.
I'm curious about the different use cases for `derive` and `derive_with_seed_concatination` though, and why we can't just choose a single one to expose for the user.

aws-lc-rs/src/tlsprf/tests/fips.rs
skmcgrail: Arbitrary with no meaning, just needed a value for testing

aws-lc-rs/src/tls_prf.rs
```diff
@@ -0,0 +1,297 @@
...
+pub const P_SHA256: &Algorithm = &Algorithm(AlgorithmID::SHA256);
+
+/// SHA-384 P_hash algorithm
+pub const P_SHA384: &Algorithm = &Algorithm(AlgorithmID::SHA384);
+
+/// SHA-512 P_hash algorithm
+pub const P_SHA512: &Algorithm = &Algorithm(AlgorithmID::SHA512);
+
+/// Encapsulates a PRF algorithm and secret bytes to be used to derive output.
+pub struct Secret<'a> {
```
justsmth, line 26: * I'm not crazy about having a struct with a lifetime; reuse requires this struct to tag along with whatever struct actually owns the "secret" value it's referencing. The alternatives are to either make the struct generic or put the secret in a Box, which also have downsides. Could this be a trait with a default implementation? Are there other alternatives?
* The name "Secret" is super-generic for a type with only one real use, but when prefixed with its owning module  (`tls_prf::Secret`) the purpose is clear, so maybe this is expected in Rust. I found guidance in the [Rust API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#iterator-type-names-match-the-methods-that-produce-them-c-iter-ty) that seems to support the use of generic type name: "*These type names make the most sense when prefixed with their owning module, for example vec::IntoIter.*".
 * Should the label be persisted in this struct as well?  How is this commonly used?

justsmth (commented):
Random notes from my review:
* OpenSSL hacks this functionality into their EVP_PKEY_CTX in (IMO) a really obnoxious way: https://github.com/openssl/openssl/blob/21f0b80cd4b32ba80843b812b01a6056daf14093/test/pkey_meth_kdf_test.c#L26
* We should add tests for this. I found a couple of places linking to the test vectors provided here: https://mailarchive.ietf.org/arch/msg/tls/fzVCzk-z3FShgGJ6DOXqM1ydxms/

aws-lc-rs/src/tls_prf.rs
skmcgrail: > I'm not crazy about having a struct with a lifetime ... The alternatives are to either make the struct generic or put the secret in a Box, which also have downsides.

I opted for the lifetime as the secret (called the pre_master_secret in [RFC 7627]) size is dependent on the key exchange protocol, thus didn't want to allocate the worst case buffer size. A box was the alternative choice that I had in mind.

> Should the label be persisted in this struct as well? How is this commonly used?

From all indications these are really meant to be used in a one-shot manner in the protocols either using a label of "master secret", the Record Protocol related keys and IVs used where the labels are one or more of ("key expansion", "client write key", and "server write key", "IV block") depending on the whether TLS 1.0, 1.1, 1.2, and the "extended master secret" extension. So the flow is that a set of inputs (secret, label, seed) will be used to derive some variable length of output key material (secret') that can then subsequently be used to derive more material via (secret', label', seed'). It's vaguely similar to the HKDF module in some sense, so I might look at that to see if we could map similar concepts to roughly match the design, but keep it intuitive.

TLS 1.0: [RFC 2246](https://www.rfc-editor.org/rfc/rfc2246)
TLS 1.1: [RFC 4346](https://www.rfc-editor.org/rfc/rfc4346)
TLS 1.2: [RFC 5246](https://www.rfc-editor.org/rfc/rfc5246)
Extended Master Secret Extension: [RFC 7627](https://www.rfc-editor.org/rfc/rfc7627)

aws-lc-rs/src/tls_prf.rs
```diff
@@ -0,0 +1,340 @@
...
+    /// Calculates `output.len()` bytes of TLS PRF using the configured [`Algorithm`], and writes it to `output`.
+    /// See [RFC5246](https://datatracker.ietf.org/doc/html/rfc5246#section-5)
+    ///
+    /// # Errors
+    /// * `Unspecified`: Returned if the PRF derivation fails.
+    pub fn derive(self, label: &[u8], seed: &[u8], output: usize) -> Result<Secret, Unspecified> {
+        prf(self.algorithm, &self.secret, label, seed, None, output)
+    }
+
+    /// Calculates `output.len()` bytes of TLS PRF using the configured [`Algorithm`], and writes it to `output`.
```
skmcgrail, line 62:
```suggestion
    /// Calculates `len` bytes of TLS PRF using the configured [`Algorithm`], and writes it to `output`.
```

aws-lc-rs/src/tls_prf.rs
```diff
@@ -0,0 +1,340 @@
...
+        if secret.is_empty() {
+            return Err(Unspecified);
+        }
+
+        let secret = Vec::from(secret).into_boxed_slice();
+
+        Ok(Self { algorithm, secret })
+    }
+
+    /// Calculates `output.len()` bytes of TLS PRF using the configured [`Algorithm`], and writes it to `output`.
```
skmcgrail, line 53:
```suggestion
    /// Calculates `len` bytes of TLS PRF using the configured [`Algorithm`], and writes it to `output`.
```

aws-lc-rs/src/tls_prf/tests/fips.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#![cfg(debug_assertions)]
+
+use crate::fips::{assert_fips_status_indicator, FipsServiceStatus};
+use crate::tls_prf::{Secret, P_SHA256, P_SHA384, P_SHA512};
+
+macro_rules! prf {
```
justsmth, line 9: NP: For clarity, could this macros be named `prf_test`
```
macro_rules! prf_test { 
  ...
}
```

justsmth (approved):

samuel40791765 (approved):

PR #242, opened 2023-10-12
feat: add AEAD open_separate_gather method

camshaft:
### Description of changes: 

In #206, we added a method that avoids copying data into a target buffer when performing an AEAD sealing operation.

This PR adds a corresponding method for AEAD opening. It works by decrypting the provided ciphertext into a destination output cleartext slice.

Note that I moved all of the EVP calls to associated methods on the `UnboundKey`, rather than being spread across the `aead` module as well as `aead::aes_gcm`.

### Testing:

I've refactored the AEAD tests a bit to test all of the supported permutations of sealer and opener functions, since the current form was getting more complicated. Note that SIV algorithms don't currently support scatter/gather operations.

### DIFF
--- a/aws-lc-rs/src/aead.rs
+++ b/aws-lc-rs/src/aead.rs
@@ -114,16 +114,9 @@
 //! assert_eq!(plaintext, decrypted_plaintext);
 //! ```
 
-use crate::{derive_debug_via_id, fips::indicator_check, hkdf, iv::FixedLength};
-use std::{fmt::Debug, ptr::null};
-
-use crate::error::Unspecified;
+use crate::{derive_debug_via_id, error::Unspecified, hkdf};
 use aead_ctx::AeadCtx;
-use aws_lc::{
-    EVP_AEAD_CTX_open, EVP_AEAD_CTX_open_gather, EVP_AEAD_CTX_seal, EVP_AEAD_CTX_seal_scatter,
-};
-use std::mem::MaybeUninit;
-use std::ops::RangeFrom;
+use core::{fmt::Debug, ops::RangeFrom};
 
 mod aead_ctx;
 mod aes_gcm;
@@ -135,6 +128,7 @@ mod poly1305;
 pub mod quic;
 mod rand_nonce;
 mod tls;
+mod unbound_key;
 
 pub use self::{
     aes_gcm::{AES_128_GCM, AES_128_GCM_SIV, AES_256_GCM, AES_256_GCM_SIV},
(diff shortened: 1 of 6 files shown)


aws-lc-rs/src/aead.rs
```diff
@@ -590,6 +611,42 @@ impl LessSafeKey {
...
+    ///
+    /// The `out_plaintext` length must match the provided `in_ciphertext`.
+    ///
+    /// # Errors
+    /// `error::Unspecified` when ciphertext is invalid. In this case, `out_plaintext` may
+    /// have been overwritten in an unspecified way.
+    ///
+    #[inline]
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn open_separate_gather<A>(
```
justsmth, line 629: Is this preferable over allowing the buffer to be in/out, which is supported by [the underlying API](https://github.com/aws/aws-lc/blob/main/include/openssl/aead.h#L381). This also matches "scatter". 

For example:
```rust
    pub fn open_in_place_gather<A>(
        &self,
        nonce: Nonce,
        aad: Aad<A>,
        in_tag: &[u8],
        in_out: &mut [u8],
    ) -> Result<(), Unspecified>
```

aws-lc-rs/src/aead/aes_gcm.rs
```diff
@@ -93,6 +93,49 @@ pub(crate) fn aead_seal_separate_scatter(
     }
 }
 
+#[inline]
+#[allow(clippy::needless_pass_by_value)]
+pub(crate) fn aead_open_separate_gather(
+    key: &AeadCtx,
+    nonce: Nonce,
+    aad: Aad<&[u8]>,
```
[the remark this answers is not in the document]
camshaft: yeah that's fine with me

aws-lc-rs/src/aead.rs
camshaft: not sure I follow... this API is specifically for decrypting from one buffer into another. Having an `in_out` means we would need to copy it into the destination buffer, either before or after calling decrypt.

justsmth: Sorry, I wasn't paying attention -- the whole point of this PR is to allow a copy to be avoided.

aws-lc-rs/src/aead/aes_gcm.rs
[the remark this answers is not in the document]
camshaft: I've cleaned up the code to be associated methods on UnboundKey, which gets rid of all of the wrapper functions. I've also removed the clippy allows where possible.

justsmth (dismissed):

skmcgrail:
Working on rebasing this on latest main

camshaft:
The changes look good to me! Thanks for the rebase

skmcgrail (approved):

justsmth (approved):

PR #243, opened 2023-10-14
Documentation cleanup

skmcgrail:

### DIFF
--- a/aws-lc-rs/src/aead.rs
+++ b/aws-lc-rs/src/aead.rs
@@ -49,7 +49,7 @@
 //!
 //! # TLS AEAD APIs
 //!
-//! Application developers developing TLS protocol implementations should use
+//! Systems developers creating TLS protocol implementations should use
 //! [`TlsRecordSealingKey`] and [`TlsRecordOpeningKey`] respectively for AEAD.
 //!
 //! # Nonce Sequence APIs
@@ -215,10 +215,6 @@ impl<N: NonceSequence> OpeningKey<N> {
     /// has been overwritten by the plaintext; `plaintext` will refer to the
     /// plaintext without the tag.
     ///
-    /// # Errors
-    /// `error::Unspecified` when ciphertext is invalid. In this case, `in_out` may have been
-    /// overwritten in an unspecified way.
-    ///
     /// # FIPS
     /// Use this method with one of the following algorithms:
     /// * `AES_128_GCM`
@@ -226,6 +222,10 @@ impl<N: NonceSequence> OpeningKey<N> {
     ///
     /// Prefer [`RandomizedNonceKey::open_in_place`], as it provides API symmetry
     /// with sealing operations.
+    ///
+    /// # Errors
+    /// `error::Unspecified` when ciphertext is invalid. In this case, `in_out` may have been
+    /// overwritten in an unspecified way.
     #[inline]
     pub fn open_in_place<'in_out, A>(
         &mut self,
@@ -280,10 +280,6 @@ impl<N: NonceSequence> OpeningKey<N> {
     ///
     /// This reassembly be accomplished with three calls to `open_within()`.
     ///
-    /// # Errors
-    /// `error::Unspecified` when ciphertext is invalid. In this case, `in_out` may have been
-    /// overwritten in an unspecified way.
-    ///
     /// # FIPS
     /// Use this method with one of the following algorithms:
     /// * `AES_128_GCM`
@@ -291,6 +287,10 @@ impl<N: NonceSequence> OpeningKey<N> {
     ///
     /// Prefer [`RandomizedNonceKey::open_within`], as it provides API symmetry
     /// with sealing operations.
+    ///
+    /// # Errors
+    /// `error::Unspecified` when ciphertext is invalid. In this case, `in_out` may have been
+    /// overwritten in an unspecified way.
     #[inline]
     pub fn open_within<'in_out, A>(
         &mut self,
@@ -359,6 +359,7 @@ fn open_within<'in_out, A: AsRef<[u8]>>(
 ///
 /// Intentionally not `Clone` or `Copy` since cloning would allow duplication
 /// of the nonce sequence.
+///
 /// # FIPS
 /// Prefer [`RandomizedNonceKey`] for sealing operations.
 pub struct SealingKey<N: NonceSequence> {
@@ -391,13 +392,13 @@ impl<N: NonceSequence> Debug for SealingKey<N> {
 impl<N: NonceSequence> SealingKey<N> {
     /// Deprecated. Renamed to `seal_in_place_append_tag`.
     ///
-    /// # Errors
-    /// See `seal_in_place_append_tag`
-    ///
     /// # FIPS
-    /// This method should not be used.
+    /// This method must not be used.
     ///
     /// See [`RandomizedNonceKey::seal_in_place_append_tag`].
+    ///
+    /// # Errors
+    /// See `seal_in_place_append_tag`
     #[deprecated(note = "Renamed to `seal_in_place_append_tag`.")]
     #[inline]
     pub fn seal_in_place<A, InOut>(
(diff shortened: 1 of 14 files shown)


aws-lc-rs/src/hmac.rs
```diff
@@ -420,7 +424,7 @@ impl Context {
 /// return value of `sign` to a tag. Use `verify` for verification instead.
 ///
 /// # FIPS
-/// FIPS users should only utilize this method with `HMAC_SHA1_FOR_LEGACY_USE_ONLY`, `HMAC_SHA224`, `HMAC_SHA256`,
+/// Use this method with `HMAC_SHA1_FOR_LEGACY_USE_ONLY`, `HMAC_SHA224`, `HMAC_SHA256`,
```
justsmth, line 427: In hkdf and digest we present the algorithms in a bulleted list, which is more clear.

aws-lc-rs/src/pbkdf2.rs
```diff
@@ -168,7 +168,7 @@ const MAX_USIZE32: u64 = u32::MAX as u64;
 /// algorithm's output length, per the PBKDF2 specification.
 ///
 /// # FIPS
-/// FIPS users should utilize this method when meeting the following conditions:
+/// Use this method when meeting the following conditions:
 /// * Algorithms: `HMAC_SHA1_FOR_LEGACY_USE_ONLY`, `HMAC_SHA224`, `HMAC_SHA256`, `HMAC_SHA384`, or `HMAC_SHA512`.
 /// * `salt.len()` >= 16
 /// * `sercet.len()` >= 14
```
justsmth, line 174: `secret` is misspelled.

aws-lc-rs/src/signature.rs
```diff
@@ -370,6 +376,11 @@ impl<B: AsRef<[u8]>> UnparsedPublicKey<B> {
     ///
     /// # Errors
     /// `error::Unspecified` if inputs not verified.
+    ///
+    /// # FIPS
+    /// * RSA Key Sizes: 1024, 2048, 3072, 4096
+    /// * NIST Elliptic Curves: P256, P384, P521
+    /// * Digest Algorithms: SHA1, SHA256, SHA384, SHA512
```
justsmth, line 383: Should we document about FIPS on the individual static values (e.g. `RSA_PKCS1_2048_8192_SHA512`) that are approved?

skmcgrail: I didn't do it on the top-level for two reasons. It's context specific depending on usage of signing versus verifying, and I wasn't sure how many people would click through since the first sentence of a constant gets embedded into the top-level documentation page.

aws-lc-rs/src/ed25519.rs
```diff
@@ -157,7 +157,7 @@ impl Ed25519KeyPair {
     /// `error::Unspecified` if `rng` cannot provide enough bits or if there's an internal error.
     ///
     /// # FIPS
-    /// FIPS users should not use this method.
+    /// This function should not be used.
```
justsmth, line 160: NP: we're not inconsistent with our use of the terms “method” and “function”. This should say “method” — I think any function defined within an `impl` block of a type/struct would be a “method”.

aws-lc-rs/src/pbkdf2.rs
```diff
@@ -168,8 +168,13 @@ const MAX_USIZE32: u64 = u32::MAX as u64;
...
-/// * Algorithms: `HMAC_SHA1_FOR_LEGACY_USE_ONLY`, `HMAC_SHA224`, `HMAC_SHA256`, `HMAC_SHA384`, or `HMAC_SHA512`.
+/// Use this method when meeting the following conditions:
+/// * Algorithms:
+///   * `HMAC_SHA1_FOR_LEGACY_USE_ONLY`
+///   * `HMAC_SHA224`
+///   * `HMAC_SHA256`
+///   * `HMAC_SHA384`
+///   * `HMAC_SHA512`
 /// * `salt.len()` >= 16
 /// * `sercet.len()` >= 14
```
justsmth, line 179: Don't keep `sercets` 🤣

aws-lc-rs/src/aead.rs
```diff
@@ -49,8 +49,8 @@
 //!
 //! # TLS AEAD APIs
 //!
-//! Application developers developing TLS protocol implementations, with FIPS compliance requirements
-//! should use the [`TlsRecordSealingKey`] and [`TlsRecordOpeningKey`] respectively.
+//! Application developers developing TLS protocol implementations should use
```
justsmth, line 52: NP: “Application developer” seems like the wrong term for someone who's implementing the TLS protocol. Maybe “Systems developer”?

aws-lc-rs/src/aead/chacha20_poly1305_openssh.rs
```diff
@@ -21,7 +21,7 @@
 //! [RFC 4253]: https://tools.ietf.org/html/rfc4253
 //!
 //! # FIPS
-//! FIPS users should not use the APIs offered in this module.
+//! The APIs offered in this module should not be used.
```
justsmth, line 24: NP: Shorten to “This module should not be used.”?

aws-lc-rs/src/hkdf.rs
```diff
@@ -306,6 +314,11 @@ impl Prk {
     ///
     /// # Panics
     /// Panics if the given Prk length exceeds the limit
+    ///
+    /// # FIPS
+    /// This function should not be used.
+    ///
+    /// See [`Salt::extract`].
```
justsmth, line 321: Should this line be before the “Panics” section above?

aws-lc-rs/src/hkdf.rs
```diff
@@ -411,13 +429,13 @@ impl<L: KeyType> Okm<'_, L> {
     /// `L: KeyType`.
     ///
     ///# FIPS
-    /// FIPS users should utilize this method when meeting the following conditions:
+    /// The following conditions should be met:
```
justsmth, line 432: * There are other FIPS sections containing a list of conditions, but are missing “the following conditions should be met”. I like the clarity this statement provides for the list.
* Should we say the conditions “must be met”, or just “should be”?

aws-lc-rs/src/tls_prf.rs
```diff
@@ -54,7 +76,7 @@ impl Secret {
 
     /// Calculates `len` bytes of TLS PRF using the configured [`Algorithm`], and returns [`Secret`] of length `len`.
     ///
-    /// In this function calling convention `seed1` and `seed2` will be concatenated together: `seed1 || seed2`.
+    /// In this method call signature `seed1` and `seed2` will be concatenated together: `seed1 || seed2`.
```
justsmth, line 79: NP: Shorten to “In this method, `seed1` and `seed2`...”

aws-lc-rs/src/cipher.rs
```diff
@@ -584,6 +593,12 @@ impl EncryptingKey {
     ///
     /// # Errors
     /// * [`Unspecified`]: Returned if there is an error constructing the `EncryptingKey`.
+    ///
+    /// # FIPS
+    /// This constructor functiion should be used with an `UnboundCipherKey` constructed with one
```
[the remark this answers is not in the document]
skmcgrail: I was being consistent with the difference between a method (that takes self), and a function or associated type function.

aws-lc-rs/src/pbkdf2.rs
skmcgrail: Old copy and paste strikes again!

aws-lc-rs/src/ed25519.rs
justsmth: I was wrong -- the term "method" also requires the function to take [the `self` param](https://doc.rust-lang.org/book/ch05-03-method-syntax.html#method-syntax):
> Unlike functions, methods are defined within the context of a struct (or an enum or a trait object, ... ), and their first parameter is always `self`, ...

justsmth (approved):

aws-lc-rs/src/pbkdf2.rs
```diff
@@ -168,8 +168,13 @@ const MAX_USIZE32: u64 = u32::MAX as u64;
 /// algorithm's output length, per the PBKDF2 specification.
 ///
 /// # FIPS
-/// FIPS users should utilize this method when meeting the following conditions:
-/// * Algorithms: `HMAC_SHA1_FOR_LEGACY_USE_ONLY`, `HMAC_SHA224`, `HMAC_SHA256`, `HMAC_SHA384`, or `HMAC_SHA512`.
+/// The following conditions must be met:
+/// * Algorithm is one of the following:
+///   * `HMAC_SHA1_FOR_LEGACY_USE_ONLY`
+///   * `HMAC_SHA224`
```
samuel40791765, line 174: 1. It seems like we're missing a `PBKDF2_HMAC_SHA224`, so this might not be applicable :(
2. Should we change these to point to the PBKDF2 constructs instead?
```
/// * `PBKDF2_HMAC_SHA1`
/// * `PBKDF2_HMAC_SHA256`
/// * `PBKDF2_HMAC_SHA384`
/// * `PBKDF2_HMAC_SHA512`
```

skmcgrail: Good catch!

samuel40791765 (approved):

justsmth (approved):

PR #246, opened 2023-10-18
Supports compresssed octets format in agreement::agree_ephemeral() #1255

hansonchar:
### Issues:
Addresses #ISSUE-1255

### Description of changes: 
* Current behavior: `agreement::agree_ephemeral()` supports the peer public key in uncompressed octets format, but rejects if the peer public key is in compressed octets format.
* Behavior after code change: supports the peer public key in either uncompressed or compressed octets format.
* Added method `compute_public_key_compressed` to `EphemeralPrivateKey` so that users can retrieve the EC public key in compressed format when applicable.

### Call-outs:

I've changed the test cases that used to disallow peer public key in compressed format to now allow them.

### Testing:
```cargo test```

### CODE REVIEW

File: aws-lc-rs/src/data/agreement_tests.txt [hunk 1]
```diff
@@ -250,31 +250,6 @@ Curve = P-521
 PeerQ = 0400D0B3975AC4B799F5BEA16D5E13E9AF971D5E9B984C9F39728B5E5739735A219B97C356436ADC6E95BB0352F6BE64A6C2912D4EF2D0433CED2B6171640012D9460F
 Error = Peer public key is missing the Y coordinate completely.
 
-Curve = P-256
```
justsmth, line 253: Instead of deleting these "negative" tests for public keys using the compressed key format, we should transform them into "positive" KATs.

PR #247, opened 2023-10-18
Validate EC_KEY created from public point

justsmth:
### Issues:
N/A

### Description of changes: 
* Validate a public EC key at time of creation.
* Construct `LcPtr` instead of `DetachableLcPtr`.

### Call-outs:
N/A

### Testing:
Tests pass locally

### DIFF
--- a/aws-lc-rs/src/ec.rs
+++ b/aws-lc-rs/src/ec.rs
@@ -6,7 +6,7 @@
 use crate::error::{KeyRejected, Unspecified};
 use core::fmt;
 
-use crate::ptr::{ConstPointer, DetachableLcPtr, LcPtr};
+use crate::ptr::{ConstPointer, DetachableLcPtr, LcPtr, Pointer};
 
 use crate::signature::{Signature, VerificationAlgorithm};
 use crate::{digest, sealed, test};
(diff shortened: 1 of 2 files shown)


aws-lc-rs/src/ec.rs
```diff
@@ -6,7 +6,7 @@
 use crate::error::{KeyRejected, Unspecified};
 use core::fmt;
 
-use crate::ptr::{ConstPointer, DetachableLcPtr, LcPtr};
+use crate::ptr::{ConstPointer, DetachableLcPtr, LcPtr, Pointer};
```
samuel40791765, line 9: Nit: Is the `Pointer` import needed?

samuel40791765 (approved):

aws-lc-rs/src/ec.rs
justsmth: Yeah, it's needed for the use of `as_const_ptr` which is [defined on that trait](https://github.com/aws/aws-lc-rs/blob/e8632a4f80ad059e644b23e6e80c7988a4b7646a/aws-lc-rs/src/ptr.rs#L185).

aws-lc-rs/src/ec.rs
```diff
@@ -228,14 +228,16 @@ pub(crate) fn marshal_public_key(ec_key: &ConstPointer<EC_KEY>) -> Result<Public
...
+) -> Result<LcPtr<EC_KEY>, Unspecified> {
+    let nid = EC_GROUP_get_curve_name(ec_group.as_const_ptr());
+    let ec_key = LcPtr::new(EC_KEY_new())?;
     if 1 != EC_KEY_set_group(*ec_key, **ec_group) {
         return Err(Unspecified);
     }
     if 1 != EC_KEY_set_public_key(*ec_key, **public_ec_point) {
         return Err(Unspecified);
     }
+    validate_ec_key(&ec_key.as_const(), nid)?;
```
dkostic, line 240: should you then add validation to `ec_key_from_public_private`?

justsmth: I agree, the validation is not invoked inside `ec_key_from_public_private`. (The function was only used [one place](https://github.com/aws/aws-lc-rs/blob/e8632a4f80ad059e644b23e6e80c7988a4b7646a/aws-lc-rs/src/ec/key_pair.rs#L159) and was making its own validation call.)

samuel40791765 (approved):

dkostic (approved):

PR #248, opened 2023-10-19
Cleanup for fips-api-changes branch

skmcgrail:
### Description of changes: 
Implemented recent changes discussed in offline review.


darylmartin100 (approved):

aws-lc-rs/src/aead.rs
```diff
@@ -392,11 +391,11 @@ impl<N: NonceSequence> Debug for SealingKey<N> {
...
     ///
-    /// # FIPS
-    /// This method must not be used.
-    ///
-    /// See [`RandomizedNonceKey::seal_in_place_append_tag`].
-    ///
+    // # FIPS
+    // This method must not be used.
+    //
+    // See [`RandomizedNonceKey::seal_in_place_append_tag`].
```
justsmth, line 397: Here and in other places, should our documentation continue to reference a "preferred"  `RandomizedNonceKey` method? Possibly changing the wording to "Prefer..." instead of "See..." or "Use..."?

aws-lc-rs/src/aead.rs
```diff
@@ -1169,7 +1137,7 @@ impl TlsRecordSealingKey {
     #[inline]
     #[allow(clippy::needless_pass_by_value)]
     pub fn seal_in_place_append_tag<A, InOut>(
-        &self,
+        &mut self,
```
justsmth, line 1140: With this, you should now be able to remove the `Mutex` from the struct as well.

aws-lc-rs/src/aead.rs
```diff
@@ -1237,20 +1205,18 @@ impl TlsRecordSealingKey {
...
 ///
-/// # FIPS
 /// Use this type in place of `LessSafeKey`, `OpeningKey`, `SealingKey`.
 pub struct TlsRecordOpeningKey {
     // The TLS specific construction for TLS ciphers in AWS-LC are not thread-safe!
     // The choice here was either wrap the underlying EVP_AEAD_CTX in a Mutex as done here,
     // or force this type to !Sync. Since this is an implementation detail of AWS-LC
     // we have optex to manage this behavior internally.
-    ctx: Mutex<AeadCtx>,
+    ctx: AeadCtx,
```
justsmth, line 1219: Was the `Mutex` needed here? If so, we should also change the "open" methods to take an exclusive/mut reference.

aws-lc-rs/src/aead.rs
```diff
@@ -360,7 +360,6 @@ fn open_within<'in_out, A: AsRef<[u8]>>(
 /// Intentionally not `Clone` or `Copy` since cloning would allow duplication
 /// of the nonce sequence.
 ///
-/// # FIPS
 /// Prefer [`RandomizedNonceKey`] for sealing operations.
```
justsmth, line 363: Should we just convert this line to a comment instead of deleting it? Should  the documentation indicate a preference for `RandomizedNonceKey` regardless?

justsmth (commented):
On a side note, the `aead.rs` file has grown huge. Can the new structs be moved out to their own source files?

aws-lc-rs/src/aead.rs
skmcgrail: As discussed in the meeting the underlying AeadCtx on the open path doesn't have any mutable state. So the call out was that it neither needs the mutex nor does it need the &mut self.

aws-lc-rs/src/aead.rs
skmcgrail: Blah missed this, thanks!

aws-lc-rs/src/aead.rs
skmcgrail: I was leaning towards making that the preferred type regardless of FIPS. Just seems like the easier API to use without throwing the complexity of NonceSequence and counters, and general problems of having to worry about avoiding nonce re-usage.

justsmth (approved):

PR #252, opened 2023-10-21
Support compressed octets format in agreement::agree_ephemeral()

hansonchar:
### Issues:
Addresses #1255

### Description of changes: 
* Current behavior: `agreement::agree_ephemeral()` supports the peer public key in uncompressed octets format, but rejects if the peer public key is in compressed octets format.
* Behavior after code change: supports the peer public key in either uncompressed or compressed octets format.
* Added method `compute_public_key_compressed` to `EphemeralPrivateKey` so that users can retrieve the EC public key in compressed format when applicable.

### Call-outs:
I've changed the test cases that used to disallow peer public key in compressed format to now allow them.

### Testing:
```
cargo test
```


skmcgrail:
I've stashed this PR contents at https://github.com/aws/aws-lc-rs/tree/hansonchar/main until we have time to review this work as part of the broader service teams' requirements.

Issue #255, opened 2023-10-22
Supports peer public key in both Octet String and X509 encoding for key agreement


hansonchar:
### Problem:

Currently only octet string encoding is supported for the input peer public key for key agreement.  This makes it difficult to interoperate with other platforms such as Java where the public key is output in X509 encoding.

### Solution:

Support the input of peer public key in both Octet String and X509 encoding for key agreement purposes.

* **Does this change any public APIs?** If yes, explain.
The proposed change should be 100% backward compatible while providing new API to specify the encoding of the peer public key for key agreement purposes.

* **Which algorithm(s) will this impact?**
The encoding specification of the input public key Key agreement for P-256, P-384, P-521 and X22519.

### Requirements / Acceptance Criteria:

Providing a peer public key in X509 DER encoding for key agreement works.

* **Will the Usage Guide or other documentation need to be updated?**
Potentially can provide examples on how to specify different encoding scheme for the peer public key.

* **Testing:** How will this change be tested? Call out new integration tests, functional tests, or particularly
  interesting/important unit tests.
Add unit test for key agreement with the peer public key in X509 DER encoding  for P-256, P-384, P-521 and X22519.

### Out of scope:

Is there anything the solution will intentionally NOT address?

Currently the output bytes of a EC public key from `aws-lc-rs` is always in octet string encoding.  It should also provide the API to output the public key in X509 DER encoding as an option.  This is however out of scope for now with respect to this issue.

justsmth:
This is an API change that we're not currently planning to make. We will reopen the issue if needed.

PR #256, opened 2023-10-22
Add support of Octet String and X509 bytes encoding (#255)

hansonchar:
### Issues:
Addresses #255

### Description of changes: 
Currently only octet string encoding is supported for the input peer public key for key agreement. This makes it difficult to interoperate with other platforms such as Java where the public key is output in X509 encoding. This change supports the input of peer public key in both Octet String and X509 encoding for key agreement purposes.

### Call-outs:
Currently the output bytes of a EC public key from `aws-lc-rs` is always in octet string encoding. It should also provide the API to output the public key in X509 DER encoding as an option.  Will however address this in a separate issue and PR.

### Testing:
```
cargo test
```

### DIFF
--- a/aws-lc-rs/src/agreement.rs
+++ b/aws-lc-rs/src/agreement.rs
@@ -51,21 +51,26 @@
 //! ```
 use crate::ec::{
     ec_group_from_nid, ec_point_from_bytes, evp_key_generate, evp_pkey_from_public_point,
+    marshal_x509_public_key_to_buffer,
 };
 use crate::error::Unspecified;
 use crate::fips::indicator_check;
-use crate::ptr::LcPtr;
+use crate::hex;
+use crate::ptr::{ConstPointer, LcPtr};
+use crate::public_key::evp_pkey_from_x509_pubkey;
 use crate::rand::SecureRandom;
-use crate::{ec, hex};
+use crate::{ec, public_key};
 use aws_lc::{
-    EVP_PKEY_CTX_new, EVP_PKEY_CTX_new_id, EVP_PKEY_derive, EVP_PKEY_derive_init,
-    EVP_PKEY_derive_set_peer, EVP_PKEY_get_raw_public_key, EVP_PKEY_keygen, EVP_PKEY_keygen_init,
+    d2i_X509_PUBKEY, i2d_PUBKEY, EVP_PKEY_CTX_new, EVP_PKEY_CTX_new_id, EVP_PKEY_derive,
+    EVP_PKEY_derive_init, EVP_PKEY_derive_set_peer, EVP_PKEY_get0_EC_KEY,
+    EVP_PKEY_get_raw_public_key, EVP_PKEY_keygen, EVP_PKEY_keygen_init,
     EVP_PKEY_new_raw_public_key, NID_X9_62_prime256v1, NID_secp384r1, NID_secp521r1, EVP_PKEY,
-    EVP_PKEY_X25519, NID_X25519,
+    EVP_PKEY_X25519, NID_X25519, X25519_PUBLIC_VALUE_LEN,
 };
 
 use core::fmt;
 use std::fmt::{Debug, Formatter};
+use std::os::raw::{c_long, c_uchar};
 use std::ptr::null_mut;
 
 #[allow(non_camel_case_types)]
@@ -360,6 +365,37 @@ impl EphemeralPrivateKey {
         }
     }
 
+    /// Computes the public key in X509 DER encoding format from the private key.
+    ///
+    /// # Errors
+    /// `error::Unspecified` when operation fails due to internal error.
+    pub fn compute_x509_pubkey(&self) -> Result<Vec<u8>, Unspecified> {
+        let mut buffer = std::ptr::null_mut::<u8>();
+        match &self.inner_key {
+            KeyInner::ECDH_P256(evp_pkey)
+            | KeyInner::ECDH_P384(evp_pkey)
+            | KeyInner::ECDH_P521(evp_pkey) => {
+                let ec_key = ConstPointer::new(unsafe { EVP_PKEY_get0_EC_KEY(**evp_pkey) })?;
+                let len = unsafe { aws_lc::i2d_EC_PUBKEY(*ec_key, &mut buffer) };
+                if len < 0 {
+                    return Err(Unspecified);
+                }
+                let buffer = LcPtr::new(buffer)?;
+                let der = unsafe { std::slice::from_raw_parts(*buffer, len.try_into()?).to_vec() };
+                Ok(der)
+            }
+            KeyInner::X25519(priv_key) => {
+                let len = unsafe { i2d_PUBKEY(**priv_key, &mut buffer) };
+                if len < 0 {
+                    return Err(Unspecified);
+                }
+                let buffer = LcPtr::new(buffer)?;
+                let der = unsafe { std::slice::from_raw_parts(*buffer, len.try_into()?).to_vec() };
+                Ok(der)
+            }
+        }
+    }
+
     /// The algorithm for the private key.
     #[inline]
     #[must_use]
@@ -446,6 +482,7 @@ impl Clone for PublicKey {
 pub struct UnparsedPublicKey<B: AsRef<[u8]>> {
     alg: &'static Algorithm,
     bytes: B,
+    encoding: &'static public_key::Encoding,
 }
 
 impl<B: Copy + AsRef<[u8]>> Copy for UnparsedPublicKey<B> {}
@@ -461,20 +498,30 @@ impl<B: Debug + AsRef<[u8]>> Debug for UnparsedPublicKey<B> {
 }
 
 impl<B: AsRef<[u8]>> UnparsedPublicKey<B> {
-    /// Constructs a new `UnparsedPublicKey`.
+    /// Constructs a new `UnparsedPublicKey` with [`public_key::OCTET_STRING`] encoding.
     pub fn new(algorithm: &'static Algorithm, bytes: B) -> Self {
         UnparsedPublicKey {
             alg: algorithm,
             bytes,
+            encoding: &public_key::OCTET_STRING,
         }
     }
 
-    /// The agreement algorithm associated with this public key
+    /// Constructs a new `UnparsedPublicKey` with [`public_key::X509`] encoding.
+    pub fn new_with_x509(algorithm: &'static Algorithm, bytes: B) -> Self {
+        UnparsedPublicKey {
+            alg: algorithm,
+            bytes,
+            encoding: &public_key::X509,
+        }
+    }
+
+    /// The agreement algorithm associated with this public key.
     pub fn algorithm(&self) -> &'static Algorithm {
         self.alg
     }
 
-    /// The bytes provided for this public key
+    /// The bytes provided for this public key.
     pub fn bytes(&self) -> &B {
         &self.bytes
     }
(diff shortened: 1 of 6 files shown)


aws-lc-rs/src/agreement.rs
```diff
@@ -530,25 +575,90 @@ where
         return Err(error_value);
     }
     let peer_pub_bytes = peer_public_key.bytes.as_ref();
-    if peer_pub_bytes.len() != expected_pub_key_len {
-        return Err(error_value);
+    if peer_public_key.encoding == &OCTET_STRING {
+        let peer_pub_bytes_len = peer_pub_bytes.len();
+        if peer_pub_bytes_len != expected_pub_key_len {
```
Mark-Simulacrum, line 580: `expected_pub_key_len` here feels confusing if it's OCTET_STRING only. Maybe we can change the method name (pub_key_len) on the Algorithm or move the expected variable declaration into this if?

aws-lc-rs/src/agreement.rs
```diff
@@ -530,25 +575,90 @@ where
...
 
+#[inline]
+fn x25519_dh_with_octstr_peer_pubkey<E>(
+    peer_pub_bytes: &[u8],
+    shared_secret: &mut [u8; MAX_AGREEMENT_SECRET_LEN],
+    priv_key: &LcPtr<EVP_PKEY>,
+    error_value: E,
+) -> Result<(), E> {
+    let mut pub_key = [0u8; X25519_PUBLIC_VALUE_LEN as usize];
+    pub_key.copy_from_slice(peer_pub_bytes);
```
Mark-Simulacrum, line 631: nit: this will panic if the lengths of the two slices don't match. It might be worth requiring the caller to pass the array form (&[u8; X25519_PUBLIC_VALUE_LEN]) and/or using `.try_into().map_err(|_| error_value)?` here?

I think in practice this should be fine since have a check for the octet string length above, but it seems a bit separated out. Moving that check into this function could also work.

aws-lc-rs/src/agreement.rs
```diff
@@ -530,25 +575,90 @@ where
...
+    error_value: E,
+) -> Result<(), E> {
+    let len = match c_long::try_from(peer_x509_pubkey.len()) {
+        Ok(len) => len,
+        Err(_) => return Err(error_value),
+    };
+    let x509_pubkey = LcPtr::new(unsafe {
+        d2i_X509_PUBKEY(
+            null_mut(),
+            &peer_x509_pubkey.as_ptr() as *const *const u8 as *mut *const c_uchar,
```
Mark-Simulacrum, line 650: I think this is UB, if I understand the OpenSSL [docs](https://www.openssl.org/docs/man3.0/man3/d2i_X509_PUBKEY.html) correctly (and presuming that AWS-LC does the same thing): "ppin is incremented to the byte following the parsed data".

IIUC, that will attempt to write through this reference to update the peer_x509_pubkey.as_ptr(). That's fine, it lives on the stack and so changing it is ok, but not while it's behind `&`; the below should work:

```suggestion
            &mut peer_x509_pubkey.as_ptr() as *const *const u8 as *mut *const c_uchar,
```

aws-lc-rs/src/agreement.rs
```diff
@@ -530,25 +575,90 @@ where
...
+            len,
+        )
+    })
+    .map_err(|()| error_value)?;
+    let mut peer_octstr_pubkey = [0u8; PUBLIC_KEY_MAX_LEN];
+    marshal_x509_public_key_to_buffer(&mut peer_octstr_pubkey, &x509_pubkey)
+        .map_err(|_| error_value)?;
+    x25519_diffie_hellman_x509(shared_secret, priv_key, &peer_octstr_pubkey)
```
Mark-Simulacrum, line 658: Nit, maybe this should call x25519_dh_with_octstr_peer_pubkey rather than copying the end of that function here, now that we have the peer_octstr_pubkey?

It's interesting that we have X25519_PUBLIC_VALUE_LEN as the local buffer length vs. PUBLIC_KEY_MAX_LEN here -- I would have expected those to match unless I'm missing something?

aws-lc-rs/src/agreement.rs
```diff
@@ -558,11 +668,18 @@ fn ec_key_ecdh<'a>(
     buffer: &'a mut [u8; MAX_AGREEMENT_SECRET_LEN],
     priv_key: &LcPtr<EVP_PKEY>,
     peer_pub_key_bytes: &[u8],
+    peer_pub_key_bytes_encoding: &Encoding,
     nid: i32,
 ) -> Result<&'a [u8], ()> {
-    let ec_group = unsafe { ec_group_from_nid(nid)? };
-    let pub_key_point = unsafe { ec_point_from_bytes(&ec_group, peer_pub_key_bytes) }?;
-    let pub_key = unsafe { evp_pkey_from_public_point(&ec_group, &pub_key_point) }?;
+    // let peer_ec_key = match peer_pub_key_bytes_encoding.id {
```
Mark-Simulacrum, line 674: nit: leftover comment

aws-lc-rs/src/agreement.rs
```diff
@@ -627,10 +744,42 @@ fn x25519_diffie_hellman<'a>(
...
+    buffer: &mut [u8],
+    evp_priv_key: &LcPtr<EVP_PKEY>,
+    peer_pub_key: &[u8],
+) -> Result<(), ()> {
+    let mut priv_key = [0u8; X25519_PRIVATE_KEY_LEN];
+    if 1 != unsafe {
+        EVP_PKEY_get_raw_private_key(
+            **evp_priv_key,
+            priv_key.as_mut_ptr(),
+            &priv_key.len() as *const _ as *mut usize,
```
Mark-Simulacrum, line 758: Similar to above: "len is updated with the number of bytes actually written" (https://www.openssl.org/docs/man3.1/man3/EVP_PKEY_get_raw_private_key.html)

```suggestion
            &mut priv_key.len() as *const _ as *mut usize,
```

aws-lc-rs/src/ec.rs
```diff
@@ -265,6 +266,23 @@ unsafe fn validate_evp_key(
...
+pub(crate) fn marshal_x509_public_key_to_buffer(
+    buffer: &mut [u8; PUBLIC_KEY_MAX_LEN],
+    x509_pubkey: &LcPtr<X509_PUBKEY>,
+) -> Result<usize, Unspecified> {
+    let evp_pkey = LcPtr::new(unsafe { X509_PUBKEY_get(**x509_pubkey) })?;
+    let bio = LcPtr::new(unsafe { BIO_new(BIO_s_mem()) })?;
+    if unsafe { i2d_PUBKEY_bio(*bio, *evp_pkey) } <= 0 {
+        return Err(Unspecified);
+    }
+    let size = BIO_get_mem_data(*bio, buffer.as_mut_ptr().cast::<*mut c_char>());
```
Mark-Simulacrum, line 279: This cast() looks suspicious to me -- again caveats of not looking at the sources, but generally it feels weird to cast a `*mut u8` to `*mut *mut c_char`, it seems like this is likely going to end up writing to some arbitrary memory address:

> BIO_get_mem_data() sets *pp to a pointer to the start of the memory BIOs data

(https://www.openssl.org/docs/man3.1/man3/BIO_get_mem_data.html)

BIO_get_mem_data should really be an `unsafe` function too, not sure why it isn't...

Mark-Simulacrum (commented):
This is exposing taking different encodings as a new parameter and function (new_with_encoding) rather than adding a new new_with_x509 creation function, like that PR did (once):

>         fn from_private_key_der(alg: &'static EcdsaSigningAlgorithm, key: &[u8]) -> Result<Self, KeyRejected>;

I think generally speaking either is fine but it is a slight deviation. I think in most cases the caller is going to know for sure what they have so adding extra functions in the public API for different encodings rather than having an Encoding struct + constants might be cleaner.

But, no strong opinions.

aws-lc-rs/src/agreement.rs
hansonchar: Will move.

hansonchar:
> This is exposing taking different encodings as a new parameter and function (new_with_encoding) rather than adding a new new_with_x509 creation function, like that PR did (once):
> 
> > ```
> >     fn from_private_key_der(alg: &'static EcdsaSigningAlgorithm, key: &[u8]) -> Result<Self, KeyRejected>;
> > ```
> 
> I think generally speaking either is fine but it is a slight deviation. I think in most cases the caller is going to know for sure what they have so adding extra functions in the public API for different encodings rather than having an Encoding struct + constants might be cleaner.
> 
> But, no strong opinions.

Will change.

aws-lc-rs/src/agreement.rs
hansonchar: >Nit, maybe this should call x25519_dh_with_octstr_peer_pubkey rather than copying the end of that function here, now that we have the peer_octstr_pubkey?

This is not a nit.  A great catch!

aws-lc-rs/src/agreement.rs
hansonchar: Thanks.  Turns out I can get rid of `x25519_dh_with_octstr_peer_pubkey` entirely.

aws-lc-rs/src/agreement.rs
hansonchar: Will fix.

aws-lc-rs/src/agreement.rs
hansonchar: Thanks.

aws-lc-rs/src/agreement.rs
hansonchar: `x25519_diffie_hellman_x509` is now gone.  (Or will be once I update the PR)

aws-lc-rs/src/ec.rs
hansonchar: Will fix this and rebase to the latest main.

aws-lc-rs/src/ec.rs
```diff
@@ -266,21 +266,29 @@ unsafe fn validate_evp_key(
...
-        return Err(Unspecified);
+        return Err(());
     }
-    let size = BIO_get_mem_data(*bio, buffer.as_mut_ptr().cast::<*mut c_char>());
-    if size <= 0 {
-        return Err(Unspecified);
+
+    let mut ptr = std::ptr::null_mut::<c_char>();
+    let size = BIO_get_mem_data(*bio, &mut ptr);
+    let size = usize::try_from(size).map_err(|_| Unspecified)?;
```
hansonchar, line 283: This should be:

```
let size = usize::try_from(size).map_err(|_| ())?;
```

Fixing ...

aws-lc-rs/src/agreement.rs
```diff
@@ -513,7 +555,7 @@ impl<B: AsRef<[u8]>> UnparsedPublicKey<B> {
 #[inline]
 #[allow(clippy::needless_pass_by_value)]
 #[allow(clippy::missing_panics_doc)]
-pub fn agree_ephemeral<B: AsRef<[u8]>, F, R, E>(
+pub fn agree_ephemeral<B: AsRef<[u8]>, F, R, E: Copy>(
```
Mark-Simulacrum, line 558: `E: Copy` here is a breaking change to the API... I think it shouldn't be needed though if you refactor the map_err calls to `if let Err(()) = ... { return Err(error_value); }` or so -- the compiler can't track the ownership through the map_err but will be able to with if let.

aws-lc-rs/src/agreement.rs
```diff
@@ -523,32 +565,81 @@ where
...
+    error_value: E,
+) -> Result<(), E> {
+    let len = match c_long::try_from(peer_x509_pubkey.len()) {
+        Ok(len) => len,
+        Err(_) => return Err(error_value),
+    };
+    let x509_pubkey = LcPtr::new(unsafe {
+        d2i_X509_PUBKEY(
+            null_mut(),
+            &mut peer_x509_pubkey.as_ptr() as *const *const u8 as *mut *const c_uchar,
```
Mark-Simulacrum, line 630:
```suggestion
            &mut peer_x509_pubkey.as_ptr() as *mut *const u8 as *mut *const c_uchar,
```

nit: this is likely not important at runtime but helps avoid future lints etc.

aws-lc-rs/src/agreement.rs
hansonchar: Good catch.  Will fix.

aws-lc-rs/src/agreement.rs
hansonchar: Will change.

aws-lc-rs/src/agreement.rs
```diff
@@ -386,6 +387,37 @@ impl EphemeralPrivateKey {
         }
     }
 
+    /// Computes the public key in X509 format from the private key.
+    ///
+    /// # Errors
+    /// `error::Unspecified` when operation fails due to internal error.
+    pub fn compute_x509_pubkey(&self) -> Result<Vec<u8>, Unspecified> {
```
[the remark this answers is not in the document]
hansonchar: Will make it clear in the doc.  Thanks!

skmcgrail:
I've stashed this PR contents at https://github.com/aws/aws-lc-rs/tree/hansonchar/main-x509 until we have time to review this work as part of the broader service teams' requirements.

PR #259, opened 2023-10-24
Expose accessors for private/public key on elliptic curve keys

justsmth:
A rebase of #234 onto the latest on main. 

### Description of changes:
* Rebased changes from #234 
* Exposes a Ed25519Seed type.
```
mod signature {
    // Technically not exposed before (private type but visible via KeyPair impl on EcdsaKeyPair)
    pub struct EcdsaPublicKey { ... }
    pub struct EcdsaPrivateKey<'a> { ... }
    pub struct Ed25519Seed<'a> { ... }


    // These are buffer discriminants, not actual buffer types.
    pub struct EcPublicDer { ... }
    pub struct EcPrivateKeyBuffer { ... }
    pub struct EcPrivateKeyDer { ... }
    pub struct Ed25519SeedBuffer { ... }

    impl EcdsaKeyPair {
        fn generate(alg: &'static EcdsaSigningAlgorithm) -> Result<Self, Unspecified>;
        fn to_pkcs8v1(&self) -> Result<Document, Unspecified>;
        fn private_key(&self) -> EcdsaPrivateKey<'_>;
        fn from_private_key_der(alg: &'static EcdsaSigningAlgorithm, key: &[u8]) -> Result<Self, KeyRejected>;
    }
    impl EcdsaPrivateKey {
        fn to_buffer(&self) -> Result<Buffer<'static, EcPrivateKeyBuffer>, Unspecified>;
        fn to_der(&self) -> Result<Buffer<'static, EcPrivateKeyDer>, Unspecified>;
    }
    impl EcdsaPublicKey {
        /// Provides the public key as a DER-encoded SubjectPublicKeyInfo structure.
        fn as_der(&self) -> Buffer<'_, EcPublicDer>;
    }
    impl Ed25519Keypair {
        pub fn to_pkcs8v1(&self) -> Result<Document, Unspecified>;
        pub fn to_pkcs8v2(&self) -> Result<Document, Unspecified>;
        pub fn seed(&self) -> Result<Ed25519Seed, Unspecified>;
        impl Drop for Ed25519KeyPair { ... }
    }
    impl Ed25519Seed {
    	fn to_buffer(&self) -> Result<Buffer<'static, Ed25519SeedBuffer>, Unspecified>;
    }
}

mod buffer {
    pub struct Buffer<'a, T> { ... }
    impl<T> AsRef<[u8]> for Buffer<'_, T> { ... }
    impl<T> fmt::Debug for Buffer<'_, T> { ... }
}
````

### Call-outs:
* There’s inconsistency currently around whether to have a function names like `to_pkcs8v1` versus `to_pkcs8`. Perhaps we need both?

(From #234)

> We should discuss & agree to the general shape of the EcdsaKeyPair / EcdsaPrivateKey / EcdsaPublicKey types, as well as the new PrivateBuffer, to decide whether that shape makes sense. It's likely that replicating it for the other keys (e.g., RSA) will make sense. It's worth noting that the openssl crate takes a different approach with a general PKey type with functions to do a variety of operations -- that approach definitely leads to a simpler API, but it's less ring-like.

> The current API surface is already a bit awkward -- for example, the only way to generate a key is via EcdsaKeyPair::generate_pkcs8 which gives you a serialized form rather than the Rust type. This PR adds a dedicated generate function, for latter use via to_pkcs8v1() if desired, but doesn't deprecate or remove the old API. We're also asking for a full signature algorithm rather than just the key type.

### Testing:
(From #234)
> A few unit tests are added. I'd like to agree on the general API shape before extending these further and it would help to gain some understanding of how much we want to test "trivial" wrappers around underlying AWS-LC functions for serialization/deserialization.

### CODE REVIEW

File: aws-lc-rs/src/ed25519.rs [hunk 1]
```diff
@@ -90,22 +92,56 @@ impl Debug for Ed25519KeyPair {
...
+
+impl Seed<'_> {
+    /// Exposes the seed encoded as a big-endian fixed-length integer.
+    ///
+    /// For most use-cases, `EcdsaKeyPair::to_pkcs8()` should be preferred.
+    ///
+    /// # Errors
+    /// `error::Unspecified` if serialization failed.
+    pub fn to_buffer(&self) -> Result<Buffer<'static, Ed25519SeedBuffer>, Unspecified> {
+        let buffer = Vec::from(&self.0.private_key[..ED25519_PRIVATE_KEY_SEED_LEN]);
```
Mark-Simulacrum, line 120: Pre-existing, but the private_key should probably be wrapped in Zeroed and/or be heap allocated within the Ed25519KeyPair type. Probably worth filing an issue for cleanup in a future PR. But the buffer added here seems like it's doing the right thing.

aws-lc-rs/src/ed25519.rs
```diff
@@ -90,22 +92,56 @@ impl Debug for Ed25519KeyPair {
...
+
+impl Debug for Seed<'_> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.write_str("Ed25519Seed()")
+    }
 }
 
+#[derive(Clone)]
+#[allow(clippy::module_name_repetitions)]
+pub struct PublicKey([u8; ED25519_PUBLIC_KEY_LEN]);
```
Mark-Simulacrum, line 133: Should we expose this as a Buffer instead? I think before this PR the PublicKey type wasn't directly public, only visible in the public API via the KeyPair trait, so we have that flexibility if we want it.

But no strong feeling either way.

Mark-Simulacrum (commented):
Just a couple tiny nits.

aws-lc-rs/src/ed25519.rs
justsmth: In the latest commit, the `Ed25519KeyPair` private key is in a Box to avoid the potential security concern.

aws-lc-rs/src/buffer.rs
```diff
@@ -0,0 +1,83 @@
...
+        Buffer(Cow::Owned(owned), PhantomData)
+    }
+
+    pub(crate) fn take_from_slice(slice: &mut [u8]) -> Buffer<'a, T> {
+        let owned = slice.to_vec();
+        slice.zeroize();
+        Buffer(Cow::Owned(owned), PhantomData)
+    }
+
+    pub(crate) fn public_from_slice(slice: &[u8]) -> Buffer<'_, T> {
```
skmcgrail, line 41: nit: Maybe `borrow_from_slice`. The take function is clear cause its going to copy and clear. This is more borrowing where the lifetime is bound to that of the passed in slice. But not sure on the choice of `public`.

aws-lc-rs/src/ec.rs
```diff
@@ -114,25 +115,40 @@ impl AlgorithmID {
...
+/// An elliptic curve public key as a DER-encoded `SubjectPublicKeyInfo` structure
+#[allow(clippy::module_name_repetitions)]
+pub struct EcPublicKeyDer {
+    _priv: (),
 }
 
 impl PublicKey {
-    fn new(pubkey_box: Box<[u8]>) -> Self {
-        PublicKey(pubkey_box)
+    /// Provides the public key as a DER-encoded `SubjectPublicKeyInfo` structure.
```
skmcgrail, line 132:
```suggestion
    /// Provides the public key as an ASN.1 DER-encoded X.509 `SubjectPublicKeyInfo` structure. [RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280).
```

aws-lc-rs/src/ec.rs
```diff
@@ -114,25 +115,40 @@ impl AlgorithmID {
...
+    #[must_use]
+    pub fn as_der(&self) -> Buffer<'_, EcPublicKeyDer> {
```
skmcgrail, line 134: I'm really not to fond of how we are using the names `as_der` and `EcPublicKeyDer` write the output as DER following the X.509 ASN.1 `SubjectPublicKeyInfo` type defiition. DER is an ASN.1 encoding scheme, `SubjectPublicKeyInfo` is the structure to describe a public key that could be encoded in various formats, PEM, DER, BER, etc. I feel like we should be more explicit for own benefit in case we need to serialize to different structures.

Case in point, you could have a function that just writes the raw key bytes as a DER encode OCTET STRING. Just won't want overload where DER is so to speak.

Thoughts?

aws-lc-rs/src/ec/key_pair.rs
```diff
@@ -224,3 +290,57 @@ fn compute_ecdsa_signature<'a>(
...
+            )?;
+            Ok(Buffer::<EcPrivateKeyBuffer>::new(buffer))
+        }
+    }
+
+    /// Encode the private key via DER into bytes.
+    ///
+    /// # Errors
+    /// `error::Unspecified`  if serialization failed.
+    pub fn to_der(&self) -> Result<Buffer<'static, EcPrivateKeyDer>, Unspecified> {
```
skmcgrail, line 334: Just wanted to highlight this as another example of ambiguity around meaning of to `to_der`. EcsdaPrivateKey encode to DER but follow PKCS#8 (v1 or v2) and is explicit with the functions. In this case though `to_der` is serializing only to the specific `PrivateKey` format, specifically from [RFC 5915](https://www.rfc-editor.org/rfc/rfc5915). This is the same structure that is wrapped in the same PKCS#8 structure. Now you could argue that KeyPair type makes sense to use PKCS#8 cause that supports both being present (V2). But it also supports serializing just the public key, so why does `to_der` here use that? Basically I think the intent needs to be more clear for the serialization functions.

aws-lc-rs/src/buffer.rs
Mark-Simulacrum: Public is intended to say that callers shouldn't use this for private material, mostly to push towards making sure take_from_slice is used where needed (otherwise easy to get this wrong).

aws-lc-rs/src/ec/key_pair.rs
justsmth: I agree with your concerns. I'm still thinking about what changes need to be made regarding them. 

Here are a few thoughts I have about what's important regarding the serialization/deserialization methods we provide for our structs:
* Documentation should be unambiguous about the encoding(s) provided or required.
* Method names should be clear about their purpose, but don't need to indicate details of the encoding.
* Method names should be consistent/symmetric within a given struct. For example, the encoding serialized by a `to_der` function should be appropriate for a `from_der` to deserialize the same struct (if such a method exists).
* Structs/Types (e.g., `EcPrivateKeyDer`) should be unambiguous and guide users toward proper usage of serialized data.

justsmth: My latest commit adds/improves the documentation so that the encoding provided by serialization methods is more clear. I also moved the structs that differentiate buffer types to a `signature::encoding` module.  We might also rename the (e.g.) `to_der` methods to more directly indicate the type of encoding provided, but I'm not yet sure that's needed.

This is a one-way-door API decision, so it should be carefully considered.

aws-lc-rs/src/ed25519.rs
```diff
@@ -90,22 +92,56 @@ impl Debug for Ed25519KeyPair {
...
     fn as_ref(&self) -> &[u8] {
-        &self.public_key
+        &self.0
     }
 }
 
 impl Debug for PublicKey {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        f.write_str(&format!("PublicKey(\"{}\")", test::to_hex(self.public_key)))
+        f.write_str(&format!("PublicKey(\"{}\")", test::to_hex(self.0)))
```
skmcgrail, line 144: Can we pull out `to_hex` to a private `fmt` module? Just a code smell to see us using code from our `test` module for non-test code paths. To not have to change all the test cases, just leave an alias behind in the `test` module for now.

aws-lc-rs/src/ec/key_pair.rs
```diff
@@ -224,3 +293,57 @@ fn compute_ecdsa_signature<'a>(
...
+            )?;
+            Ok(Buffer::<EcPrivateKeyBin>::new(buffer))
+        }
+    }
+
+    /// Serializes the key as a DER-encoded `ECPrivateKey` (RFC 5915) structure.
+    ///
+    /// # Errors
+    /// `error::Unspecified`  if serialization failed.
+    pub fn to_der(&self) -> Result<Buffer<'static, EcPrivateKeyRfc5915Der>, Unspecified> {
```
skmcgrail, line 337: Which library and/or language is serializing to `ECPrivateKey` directly that we need to inter-opt with?

skmcgrail (commented):
Just some minor feedback and one question, otherwise this looks good.

aws-lc-rs/src/ec.rs
hansonchar: Would `as_x509` and `X509PublicKey` any better?

As a reference, in the `BC-FIPS` Java world, the import of the der bytes (after base64 decoding if any) is something like:
```
new X509EncodedKeySpec(bytes)
```

aws-lc-rs/src/ec/key_pair.rs
hansonchar: One popular library is BC-FIPS (in Java).

aws-lc-rs/src/buffer.rs
hansonchar: This isn't obvious to the reader.  Adding method doc should help.

hansonchar (commented):
Instead of always marshaling a public key eagerly into both octet string and x509 formats, would it make more sense to have an API that allows either format to be lazily retrieved instead?

I can imagine cases where only one format is needed but not the other.

aws-lc-rs/src/ed25519.rs
justsmth: I agree.  I have this change in a separate commit that I plan to rebase after this merges.

aws-lc-rs/src/buffer.rs
```diff
@@ -0,0 +1,85 @@
...
+use std::marker::PhantomData;
+
+use zeroize::Zeroize;
+
+/// This is a buffer type for some data exposed by various APIs in this crate.
+///
+/// `T` acts as a discriminant between different kinds of data.
+///
+/// The buffer will be zeroed on drop if it is owned.
+pub struct Buffer<'a, T>(Cow<'a, [u8]>, PhantomData<T>);
```
skmcgrail, line 20: Should we maybe call this `SecretBuffer` or `PrivateBuffer` I could see this type being useful for other data like public-key data etc. But the description implies that the content is meant to be private. Should this be communicated through the name, or should the documentation be updated to reflect that the data MAY be private.

hansonchar: >But the description implies that the content is meant to be private.

Note FIPS140-3 Security Level 4 requires zeroisation of all unprotected SSPs (Security Sensitive Parameters), where SSP includes PSP (Public Security Parameters).  In other words, the need to zeroize a buffer do NOT necessarily mean the buffer contains any secret/private data.  It may also contain public data, such as public key, and still may need to be zeroised depending on the specific context.

In short, "the description implies that the content is meant to be private" is not technically correct.

aws-lc-rs/src/ec.rs
```diff
@@ -114,25 +114,63 @@ impl AlgorithmID {
...
+        }
+        let mut buffer = null_mut::<u8>();
+        let len = unsafe { aws_lc::i2d_EC_PUBKEY(*ec_key, &mut buffer) };
+        if len < 0 || buffer.is_null() {
+            return Err(Unspecified);
+        }
+        let buffer = LcPtr::new(buffer)?;
+        let mut der = unsafe { std::slice::from_raw_parts(*buffer, len.try_into()?) }.to_owned();
+
+        Ok(Buffer::take_from_slice(&mut der))
```
skmcgrail, line 154: `to_owned` returns a `Vec`, but then we are passing it to `take_from_slice` that then again converts it into a `Vec` using `to_vec`. Just pass the slice created from `from_raw_parts_mut` (without calling `to_owned`) to to `take_from_slice` directly. Otherwise we are making 2 clones rather then 1.

skmcgrail (approved):
Hopefully my final comment :)

skmcgrail (approved):

aws-lc-rs/src/ec.rs
```diff
@@ -114,25 +114,63 @@ impl AlgorithmID {
     }
 }
 
+/// Elliptic curve public key.
 #[derive(Clone)]
-pub struct PublicKey(Box<[u8]>);
+pub struct PublicKey {
+    algorithm: &'static EcdsaSigningAlgorithm,
+    octets: Box<[u8]>,
```
samuel40791765, line 121: Just a general question: We were doing this previously as well, so its not a blocker. We allocate the public key with a maximum length on the stack in `marshal_public_key`, does `octets` have to be `Box` here?

aws-lc-rs/src/lib.rs
```diff
@@ -103,6 +103,7 @@ extern crate core;
 
 pub mod aead;
 pub mod agreement;
+mod buffer;
```
samuel40791765, line 106: minor nit: Noticed this when checking if `Buffer` was a new public interface, should we move `buffer` down to list of other non-public modules below

samuel40791765 (approved):

hansonchar:
:clap:

PR #260, opened 2023-10-24
KEM: Key-Encapsulation Mechanisms API Support

skmcgrail:
### Description of changes: 
Introduces a new `aws_lc_rs::kem` module with support for the Kyber Round 3 submission algorithms with 512, 768, and 1024 bit strengths.

### Call-outs:
PR(s) that need to still be merged:
* [x] https://github.com/aws/aws-lc-rs/pull/265

### Testing:
The AWS-LC KATs are used to validate the the implemented APIs.


aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,575 @@
...
+                shared_secret.as_mut_ptr(),
+                &mut shared_secret_len,
+                ciphertext.as_mut_ptr(),
+                ciphertext.len(),
+            )
+        } {
+            return Err(error_value);
+        }
+
+        kdf(&shared_secret)
```
dkostic, line 292: where does this come from??

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,575 @@
...
+    #[must_use]
+    pub fn algorithm(&self) -> &'static Algorithm {
+        self.algorithm
+    }
+
+    /// Performs the encapsulate operation using the current KEM public key, generating a ciphertext
+    /// and associated shared secret.
+    ///
+    /// After the encapsulation is finished, `encapsulate` calls `kdf` with the ciphertext and raw shared secret
+    /// from the operation and then returns what `kdf` returns.
```
dkostic, line 354: do you have a reference for this definition?

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,575 @@
...
+                ciphertext.as_mut_ptr(),
+                &mut ciphertext_len,
+                shared_secret.as_mut_ptr(),
+                &mut shared_secret_len,
+            )
+        } {
+            return Err(error_value);
+        }
+
+        kdf(&ciphertext, &shared_secret)
```
dkostic, line 387: so in `decaps` you call `kdf(shared_secret)` and here you call `kdf(ciphertext, shared_secret)`. But the output of these two function calls should be the same for the protocol to work. How does this even pass the tests?

aws-lc-rs-testing/benches/kem_benchmark.rs
```diff
@@ -0,0 +1,143 @@
...
+        b.iter(|| {
+            let private_key = aws_lc_rs_benchmarks::new_private_key(config);
+            aws_lc_rs_benchmarks::decapsulate(config, &private_key);
+        });
+    });
+}
+
+fn bench_kem(c: &mut Criterion) {
+    test::run(
+        test_file!("data/kem_benchmarks.txt"),
```
dkostic, line 126: why do you need KATs to perform benchmarks?

skmcgrail: Good question, not sure why this path was chosen in the prior PRs. I agree, will follow-up a change in a separate PR for review.

aws-lc-rs/src/kem.rs
skmcgrail: `kdf` here is a user provided function of type `F` where is a `FnOnce(&[u8]) -> Result<R, E>`. The decapsulate flow the function is provided the plaintext of the secret, and the function can either return the secret as-is or do subsequent key-derivation using some other KDF of their choosing. The return signature of the provide function `F` is `Result<R, E>`, where `R` and `E` are generic parameters determined by the function provided by the user. This is the same result type that `decapsulate` returns. So to summarize: `decapsulate` takes a function provided by the user that is given the plaintext secret, and that function then influence the final result outcome of `decapsulate`, by either returning the secret as-is, performing further derivation logic, or returning an error.

This is similar for `encapsulate`, except that encapsulate takes a user function that excepts both the plaintext and ciphertext output from encaps algorithm, and that function can then determine what the final outputs are from the overall function.

aws-lc-rs/src/kem.rs
skmcgrail: Addressed in the earlier comment about the `kdf` argument.

aws-lc-rs/src/kem.rs
skmcgrail: Rust is weird, I assume you are looking for the function definition of `kdf` argument? It's a generic parameter define in the function definition's where clause.

```
where
    F: FnOnce(&[u8], &[u8]) -> Result<R, E>
```

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,564 @@
...
+    Id: AlgorithmIdentifier,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Debug::fmt(&self.id, f)
+    }
+}
+
+/// A serializable private key usable with KEMs. This can be randomly generated with `KemPrivateKey::generate`
+/// or constructed from raw bytes.
+pub struct DecapsulationKey<Id = AlgorithmId>
```
ctz, line 143: Would be good if this could be `Sync` and `Send` if that is possible.

```
error[E0277]: `*mut aws_lc_sys::linux_x86_64_crypto::evp_pkey_st` cannot be shared between threads safely
   --> rustls-postquantum/src/lib.rs:83:28
    |
83  | impl ActiveKeyExchange for Active {
    |                            ^^^^^^ `*mut aws_lc_sys::linux_x86_64_crypto::evp_pkey_st` cannot be shared between threads safely
    |
    = help: within `Active`, the trait `Sync` is not implemented for `*mut aws_lc_sys::linux_x86_64_crypto::evp_pkey_st`
note: required because it appears within the type `ManagedPointer<*mut evp_pkey_st>`
   --> /home/jbp/.cargo/git/checkouts/aws-lc-rs-815c4a1fbe3a9dc5/de40172/aws-lc-rs/src/ptr.rs:18:19
    |
18  | pub(crate) struct ManagedPointer<P: Pointer> {
    |                   ^^^^^^^^^^^^^^
note: required because it appears within the type `KemPrivateKey`
   --> /home/jbp/.cargo/git/checkouts/aws-lc-rs-815c4a1fbe3a9dc5/de40172/aws-lc-rs/src/kem.rs:163:12
    |
163 | pub struct KemPrivateKey {
    |            ^^^^^^^^^^^^^
note: required because it appears within the type `Active`
   --> rustls-postquantum/src/lib.rs:77:8
    |
77  | struct Active {
    |        ^^^^^^
note: required by a bound in `ActiveKeyExchange`
   --> /home/jbp/rustls/rustls/src/crypto/mod.rs:268:37
    |
268 | pub trait ActiveKeyExchange: Send + Sync {
    |                                     ^^^^ required by this bound in `ActiveKeyExchange`

error[E0277]: `*mut aws_lc_sys::linux_x86_64_crypto::evp_pkey_st` cannot be sent between threads safely
   --> rustls-postquantum/src/lib.rs:83:28
    |
83  | impl ActiveKeyExchange for Active {
    |                            ^^^^^^ `*mut aws_lc_sys::linux_x86_64_crypto::evp_pkey_st` cannot be sent between threads safely
```

(I have picked up a different version of this PR that I rebased on top of main, so the type names are the old ones.)

skmcgrail: Good catch this was a miss, these API types should be Send and Sync. Thanks!

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,530 @@
...
+
+/// A KEM algorithm
+#[derive(PartialEq)]
+pub struct Algorithm<Id = AlgorithmId>
+where
+    Id: AlgorithmIdentifier,
+{
+    pub(crate) id: Id,
+    pub(crate) secret_key_size: usize,
+    pub(crate) public_key_size: usize,
```
dkostic, line 79: should these be `encapsulation/decapsulation_key_size` now that we renamed the types?

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,530 @@
...
+    }
+}
+
+impl crate::sealed::Sealed for AlgorithmId {}
+
+impl<Id> DecapsulationKey<Id>
+where
+    Id: AlgorithmIdentifier,
+{
+    /// Generate a new KEM private key for the given algorithm.
```
dkostic, line 155:
```suggestion
    /// Generate a new KEM decapsulation key for the given algorithm.
```

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,530 @@
...
+        })
+    }
+
+    /// Return the algorithm associated with the given KEM private key.
+    #[must_use]
+    pub fn algorithm(&self) -> &'static Algorithm<Id> {
+        self.algorithm
+    }
+
+    /// Computes the KEM public key from the KEM private key
```
dkostic, line 185:
```suggestion
    /// Computes the KEM encapsulation key from the KEM decapsulation key
```

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,530 @@
...
+    /// Return the algorithm associated with the given KEM public key.
+    #[must_use]
+    pub fn algorithm(&self) -> &'static Algorithm<Id> {
+        self.algorithm
+    }
+
+    /// Performs the encapsulate operation using the current KEM public key, generating a ciphertext
+    /// and associated shared secret.
+    ///
+    /// After the encapsulation is finished, `encapsulate` calls `kdf` with the ciphertext and raw shared secret
```
dkostic, line 344: this sentence should be dropped i guess.

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,530 @@
...
+where
+    Id: AlgorithmIdentifier,
+{
+    /// Return the algorithm associated with the given KEM public key.
+    #[must_use]
+    pub fn algorithm(&self) -> &'static Algorithm<Id> {
+        self.algorithm
+    }
+
+    /// Performs the encapsulate operation using the current KEM public key, generating a ciphertext
```
dkostic, line 341:
```suggestion
    /// Performs the encapsulate operation using the current KEM encapsulation key, generating a ciphertext
```

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,530 @@
...
+                &mut ciphertext_len,
+                shared_secret.as_mut_ptr(),
+                &mut shared_secret_len,
+            )
+        } {
+            return Err(Unspecified);
+        }
+
+        ciphertext.truncate(ciphertext_len);
+        shared_secret.truncate(shared_secret_len);
```
dkostic, line 370: why is truncation needed here?

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,530 @@
...
+
+        Ok((
+            Ciphertext::new(ciphertext),
+            SharedSecret::new(shared_secret.into_boxed_slice()),
+        ))
+    }
+
+    /// Creates a new KEM public key from raw bytes. This method is MUST NOT be used to generate
```
dkostic, line 378:
```suggestion
    /// Creates a new KEM encapsulation key from raw bytes. This method MUST NOT be used to generate
```

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,530 @@
...
+impl<Id> Drop for DecapsulationKey<Id>
+where
+    Id: AlgorithmIdentifier,
+{
+    fn drop(&mut self) {
+        self.priv_key.zeroize();
+    }
+}
+
+impl<Id> AsRef<[u8]> for DecapsulationKey<Id>
```
justsmth, line 300: Our current use-case doesn't have a need for `DecapsulationKey::as_ref`. This (potentially dangerous) function is also not useful w/o an (e.g.) `DecapsulationKey::new` function that can construct from the byte slice.

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,530 @@
...
+    Id: AlgorithmIdentifier,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("PublicKey")
+            .field("algorithm", &self.algorithm)
+            .finish_non_exhaustive()
+    }
+}
+
+/// A set of encrypted bytes produced by [`PublicKey::encapsulate`], and used as an input to [`PrivateKey::decapsulate`].
```
justsmth, line 438: Throughout the documentation there are references to "PublicKey" and "PrivateKey", but these are actually referring to `EncapsulationKey` and `DecapsulationKey`, respectively.

aws-lc-rs/build.rs
```diff
@@ -26,4 +28,17 @@ fn main() {
...
+    // configuration to allow the tests to be built and enabled.
+    // Additionally these APIs are not thread safe. So force tests to run single threaded.
+    if env::var("AWS_LC_RUST_PRIVATE_INTERNALS")
+        .unwrap_or("0".into())
+        .eq("1")
+        && cfg!(feature = "bindgen")
+    {
+        println!("cargo:rustc-cfg=private_api");
+        println!("cargo:rustc-env=RUST_TEST_THREADS=1");
+    }
```
justsmth, line 43: Here and in the sys-crates, we should also rebuild on change to the environment variable:
```
println!("cargo:rerun-if-env-changed=AWS_LC_RUST_PRIVATE_INTERNALS");
```

aws-lc-rs/src/unstable/kem.rs
```diff
@@ -0,0 +1,339 @@
...
+
+                let alice_shared_secret = priv_key.decapsulate(ciphertext).unwrap();
+                assert_eq!(alice_shared_secret.as_ref(), shared_secret_bytes);
+
+                Ok(())
+            });
+        };
+    }
+
+    #[cfg(private_api)]
```
justsmth, line 190: Getting these tests to run takes some digging. Should we have a comment here indicating how to setup the build so that these tests execute?
```
AWS_LC_RUST_PRIVATE_INTERNALS=1 cargo test --features unstable,bindgen
```

aws-lc-rs/src/unstable/kem.rs
```diff
@@ -0,0 +1,339 @@
...
+                let ciphertext_bytes = test_case.consume_bytes("ct");
+                let shared_secret_bytes = test_case.consume_bytes("ss");
+
+                // Set randomness generation in deterministic mode.
+                unsafe {
+                    pq_custom_randombytes_use_deterministic_for_testing();
+                    pq_custom_randombytes_init_for_testing(seed.as_ptr());
+                }
+
+                let priv_key = PrivateKey::generate($alg).unwrap();
```
justsmth, line 171: At some point these tests became broken:
```
   Compiling aws-lc-sys v0.12.1 (/Users/justsmth/repos/aws-lc-rs/aws-lc-sys)
error[E0433]: failed to resolve: use of undeclared type `PrivateKey`
   --> aws-lc-rs/src/unstable/kem.rs:171:32
    |
171 |                 let priv_key = PrivateKey::generate($alg).unwrap();
    |                                ^^^^^^^^^^ use of undeclared type `PrivateKey`
...
193 |         kem_kat_test!("../../tests/data/kyber512r3.txt", &KYBER512_R3);
    |         -------------------------------------------------------------- in this macro invocation
````

File: aws-lc-rs/src/kem.rs [hunk 18]
```diff
@@ -0,0 +1,530 @@
...
+                self.algorithm.id.nid(),
+                pubkey_bytes.as_ptr(),
+                public_key_size,
+            )
+        })?;
+
+        Ok(EncapsulationKey {
+            algorithm: self.algorithm,
+            evp_pkey: pubkey,
+            pub_key: pubkey_bytes.into(),
```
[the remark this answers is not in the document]
skmcgrail: No, I think we should bias to just serializing this on-demand, will include in this next commit.

aws-lc-rs/src/kem.rs
skmcgrail: This is just safety step to ensure that AWS-LC didn't return us a smaller length then value then the buffer we passed in. Right now the buffer is sized to the expected size so it should be a no-op.

aws-lc-rs/build.rs
skmcgrail: We discussed offline removing these KATs since they are covered by AWS-LC, and we agreed that the round-trip tests in aws-lc-rs should thus be sufficient. I have removed the sys crate changes and the corresponding private_api crate feature.

aws-lc-rs/src/unstable/kem.rs
skmcgrail: See earlier comment.

aws-lc-rs/src/unstable/kem.rs
skmcgrail: See previous comment about this being removed.

aws-lc-rs-testing/benches/kem_benchmark.rs
```diff
@@ -0,0 +1,74 @@
...
+use criterion::{criterion_group, criterion_main, Criterion};
+
+const UNSTABLE_ALGORITHMS: &[Option<&aws_lc_rs::kem::Algorithm<AlgorithmId>>] = &[
+    get_algorithm(AlgorithmId::Kyber512_R3),
+    get_algorithm(AlgorithmId::Kyber768_R3),
+    get_algorithm(AlgorithmId::Kyber1024_R3),
+];
+
+fn bench_kem_keygen(c: &mut Criterion) {
+    for ele in UNSTABLE_ALGORITHMS {
```
dkostic, line 17: just curious, what does `ele` stand for? :)

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,504 @@
...
+}
+
+mod encoding {
+    pub struct PublicKeyBinDerType {
+        _priv: (),
+    }
+}
+
+/// A KEM Encapsulation Key Big-Endian number representation encoded using DER.
+pub type EncapsulationKeyBinDer = Buffer<'static, encoding::PublicKeyBinDerType>;
```
dkostic, line 289: what is this? The encapsulation key is not necessarily a number, for example in Kyber.

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,504 @@
...
+    }
+}
+
+impl<'a> From<&'a [u8]> for Ciphertext<'a> {
+    fn from(value: &'a [u8]) -> Self {
+        Self(Cow::Borrowed(value))
+    }
+}
+
+/// The cryptograpic shared secret output from the KEM encapsulate / decapsulate process.
```
dkostic, line 444:
```suggestion
/// The cryptographic shared secret output from the KEM encapsulate / decapsulate process.
```

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,504 @@
...
+
+    let mut key_raw: *mut EVP_PKEY = null_mut();
+    if 1 != unsafe { EVP_PKEY_keygen(*ctx, &mut key_raw) } {
+        return Err(Unspecified);
+    }
+    Ok(LcPtr::new(key_raw)?)
+}
+
+#[cfg(test)]
+mod tests {
```
dkostic, line 483: can you please add a comment about what this test does?

aws-lc-rs-testing/benches/kem_benchmark.rs
skmcgrail: Just a shortened variable name for the word "element".

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,504 @@
...
+    /// associated with this KEM decapsulation key.
+    ///
+    /// # Errors
+    /// `Unspecified` when operation fails due to internal error.
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn decapsulate(&self, ciphertext: Ciphertext<'_>) -> Result<SharedSecret, Unspecified> {
+        let mut shared_secret_len = self.algorithm.shared_secret_size();
+        let mut shared_secret: Vec<u8> = vec![0u8; shared_secret_len];
+
+        let ctx = LcPtr::new(unsafe { EVP_PKEY_CTX_new(*self.evp_pkey, null_mut()) });
```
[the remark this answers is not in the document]
skmcgrail: Ah I think this is a left-over from when we took a KDF function enclosure with a custom error type. Will cleanup.

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,504 @@
...
+            return Err(Unspecified);
+        }
+
+        shared_secret.truncate(shared_secret_len);
+
+        Ok(SharedSecret(shared_secret.into_boxed_slice()))
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn new(alg: &'static Algorithm<Id>, bytes: &[u8]) -> Result<Self, KeyRejected> {
```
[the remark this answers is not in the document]
justsmth: Yes, sorry.  This comment was intended for `EncapsulationKey::new`

aws-lc-rs/src/kem.rs
skmcgrail: Alright so what is it? We are trying to have a type that signifies that the given EncapsulationKey is being serialized to some sort of DER encoded value. Do we just call it `EncapsulationKeyRawDer`?

aws-lc-rs/src/kem.rs
skmcgrail: It's not a test, it is a Rust module, so there is nothing really to document since its a standard idiom.

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,504 @@
...
+    }
+}
+
+impl Drop for SharedSecret {
+    fn drop(&mut self) {
+        self.0.zeroize();
+    }
+}
+
+impl AsRef<[u8]> for SharedSecret {
```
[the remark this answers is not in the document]
skmcgrail: I believe the guidance would be the user makes use of the `hkdf` module if they can't use the full output length of the shared secret (no truncation), or if they need more bytes then what is provided. Probably a question for @dkostic to confirm. We can provide a code example on SharedSecret to show the intended usage as such.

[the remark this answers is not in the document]
justsmth: A code example that shows a proper way to derive an (AES?) key from a SharedSecret  would be good. We could potentially add a function that makes it explicit.  Something like:
```
impl SharedSecret {
   fn to_hkdf_salt(&self, hkdf::Algorithm) -> hkdf::Salt { ... }}
}
```

aws-lc-rs/src/kem.rs
dkostic: > Alright so what is it?

it can be different things, depends on the KEM.

> Do we just call it EncapsulationKeyRawDer?

sounds good to me.

aws-lc-rs/src/kem.rs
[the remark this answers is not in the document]
dkostic: > Are there standard ways of deriving (e.g.) an AES key from a SharedSecret?

KDFs are designed for this purpose.

> Exposing the bytes doesn't provide much guidance on best practice for how the bytes should be used.

How the bytes should be used should be defined by the protocol where the KEM is being used. Establishing the shared secret and using it for AES key is only one example of a protocol. Maybe the guidance should be something like "do not use KEMs on your own, KEM is only one piece of the puzzle, it should be used within a well defined protocol..."

aws-lc-rs/src/kem.rs
skmcgrail: Hmm I double checked and believe there was a misunderstanding on my part that this was encoded as ASN.1 as well, when it was not. Further investigation appears that these bytes, in the case of the draft ML-KEM specification, is actually little-endian encoded? I assume this is another aspect for KEM that could become algorithm dependent. 

Since this type will be generic over multiple algorithms that match the KEM interface, I wonder if we should just support raw public keys using the `AsRef<[u8]>` trait instead? This would seem more appropriate then having trait implementations of `AsBigEndian` or `AsLittleEndian` as those will be algorithm dependent, and we should only support the representations that are inline with the specifications. We'd then wait to have a standardized serialization format like PKCS#8 / X.509 SubjectPublicKeyInfo before implementing an encoding trait.

Otherwise if we don't want this to behind a flexible interface, then maybe we should take a step-back and adopt a design that is closer to the [AES Key Wrap](https://github.com/aws/aws-lc-rs/pull/250) module design. Where we have a general trait on the behavior, and implementations of the behavior as child modules.

Thoughts @justsmth & @dkostic ?

aws-lc-rs/src/kem.rs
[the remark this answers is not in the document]
skmcgrail: Closing this in favor of an earlier thread with Dusan where I want to riff on the raw bytes story some more.

aws-lc-rs/src/kem.rs
dkostic: > Since this type will be generic over multiple algorithms that match the KEM interface, I wonder if we should just support raw public keys using the AsRef<[u8]> trait instead?

I think that's a good idea. Once a serialization format is standardized we can define a proper encoding.

aws-lc-rs/src/kem.rs
```diff
@@ -0,0 +1,473 @@
...
+                ciphertext.as_mut_ptr(),
+                &mut ciphertext_len,
+                shared_secret.as_mut_ptr(),
+                &mut shared_secret_len,
+            )
+        } {
+            return Err(Unspecified);
+        }
+
+        ciphertext.truncate(ciphertext_len);
```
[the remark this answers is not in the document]
skmcgrail: Added comments in the latest commit.

dkostic (dismissed):

justsmth (approved):

dkostic (approved):

justsmth (approved):

darylmartin100 (approved):

PR #273, opened 2023-11-13
RSA Module Refactor

skmcgrail:
Just draft at the moment, contains a mostly complete step towards some of the API changes introduced in 0.17.X (the backwards compatible ones), but doesn't introduce any of the new types like `KeyPairComponenets`. I've opted for a different module hierarchy internally, and use it as an opportunity to cleanup our very big rsa.rs file.


aws-lc-rs/src/signature.rs
```diff
@@ -390,133 +397,133 @@ impl<B: AsRef<[u8]>> UnparsedPublicKey<B> {
 }
 
 /// Verification of signatures using RSA keys of 1024-8192 bits, PKCS#1.5 padding, and SHA-1.
-pub static RSA_PKCS1_1024_8192_SHA1_FOR_LEGACY_USE_ONLY: RsaParameters = RsaParameters(
+pub static RSA_PKCS1_1024_8192_SHA1_FOR_LEGACY_USE_ONLY: RsaParameters = RsaParameters::new(
```
justsmth, line 400: NP: Should we migrate these `pub static` variables to instead be `pub const`?

aws-lc-rs/src/rsa/key.rs
```diff
@@ -0,0 +1,490 @@
...
+            }
+            Err(_) => verify_unreachable!(),
+        }
+    }
+}
+
+impl Debug for KeyPair {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), fmt::Error> {
+        f.write_str(&format!(
+            "KeyPair {{ public_key: {:?} }}",
```
justsmth, line 270: I think it would be better for the debug message to indicate that it's an "RsaKeyPair".

justsmth (approved):

aws-lc-rs/src/signature.rs
skmcgrail: I'm going to do this one is a separate PR, mainly as moving these to `const` will require moving the `digest` module statics to const as well.

Issue #274, opened 2023-11-15
Add support for signature verification using public key in X509 DER encoding format.


hansonchar:
### Problem:

In addition to [Exposing accessors for private/public key on elliptic curve keys](https://github.com/aws/aws-lc-rs/pull/259), we need signature verification to support public key in X509 DER encoding format.  This is similar to the issue addressed by https://github.com/aws/aws-lc-rs/pull/256 for key agreement, but for signature verification. 

### Solution:

The proposed solution is to build on top of https://github.com/aws/aws-lc-rs/pull/259, but add support for signature verification using public key not only in Octet String format, but also X509 DER encoding format.

* **Does this change any public APIs?** If yes, explain.

Adding functions:
* signature::UnparsedPublicKey.new_with_x509()

* **Which algorithm(s) will this impact?**
RSA, ed25519 and EC

### Requirements / Acceptance Criteria:

Signature verification works with public key provided in either Octet String format or X509 DER encoding format.

* **Related Issues:** Link any relevant issues
* https://github.com/aws/aws-lc-rs/pull/259
* https://github.com/aws/aws-lc-rs/pull/256

* **Will the Usage Guide or other documentation need to be updated?**

Yes

* **Testing:** How will this change be tested? Call out new integration tests, functional tests, or particularly
  interesting/important unit tests.

Unit tests added to cover RSA, EC and ed25519 for signature verification using public key in X509 DER encoding format.  More unit tests will be added if this proposed pull request looks reasonable enough to be potentially accepted.

justsmth:
This is an API change that we're not currently planning to make. We will reopen the issue if needed.

PR #275, opened 2023-11-15
Add support for signature verification using public key in X509 DER encoding format.

hansonchar:
### Issues:
Addresses #274

### Description of changes: 

Currently signature can only be verified with public key provided in Octet String format.  This PR would additionally allow signature to be verified with public key provided in X509 DER encoding format.

### Call-outs:

This PR is built on top of https://github.com/aws/aws-lc-rs/pull/259

### Testing:
```
cargo test
```


skmcgrail:
I've stashed this PR contents at https://github.com/aws/aws-lc-rs/tree/hansonchar/ec-raw-api_verify_sig_with_x509_pubkey until we have time to review this work as part of the broader service teams' requirements.

PR #277, opened 2023-11-15
fix(aws-lc-sys): add `links` attribute to Cargo.toml

camshaft:
### Description of changes: 

The `aws-lc-sys` crate does not currently set the `links` attribute in the Cargo.toml. This prevents other crates from reading the `include` path and using it in other C libraries.

This change adds the `links` attribute and adds an integration test to ensure this functionality is preserved.

### Call-outs:

I did not refactor the `aws-lc-fips-sys` crate. There's quite a bit of copy/paste between the build scripts and could probably use a refactor.

### Testing:

I've added an integration test crate (`aws-lc-sys-testing`) showing that the change is now working and will be preserved in the future.

### DIFF
--- a/aws-lc-sys-testing/build.rs
+++ b/aws-lc-sys-testing/build.rs
@@ -1,3 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
 fn main() {
     // ensure that the include path is exported and set up correctly
     cc::Build::new()
--- a/aws-lc-sys-testing/build.rs
+++ b/aws-lc-sys-testing/build.rs
@@ -1,0 +1,16 @@
+fn main() {
+    // ensure that the include path is exported and set up correctly
+    cc::Build::new()
+        .include(env("DEP_AWS_LC_0_12_0_INCLUDE"))
+        .file("src/testing.c")
+        .compile("aws_ls_sys_testing");
+
+    // ensure the libcrypto artifact is linked
+    println!("cargo:rustc-link-lib=aws_lc_0_12_0_crypto");
+}
+
+fn env<S: AsRef<str>>(s: S) -> String {
+    let s = s.as_ref();
+    println!("cargo:rerun-if-env-changed={s}");
+    std::env::var(s).unwrap_or_else(|_| panic!("missing env var {s}"))
+}
--- a/aws-lc-sys-testing/build.rs
+++ b/aws-lc-sys-testing/build.rs
@@ -1,15 +1,22 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR ISC
 
+use toml_edit::Document;
+
 fn main() {
+    let cargo_toml = std::fs::read_to_string("../aws-lc-sys/Cargo.toml").unwrap();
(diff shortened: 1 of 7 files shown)


skmcgrail:
> I did not refactor the aws-lc-fips-sys crate. There's quite a bit of copy/paste between the build scripts and could probably use a refactor.

Not incorrect, but the solution would require us to at least publish one additional crate dependency which I think we were trying to avoid until we have better automation in place for the crate publishing process.

camshaft:
> but the solution would require us to at least publish one additional crate (build) dependency 

I don't think that's the only solution. It could be solved with symlinks, [path attribute](https://doc.rust-lang.org/beta/reference/items/modules.html#the-path-attribute) or [include!](https://doc.rust-lang.org/std/macro.include.html). All of those could be used to pull the `aws-lc-sys` code in to the fips crate.

skmcgrail:
> > but the solution would require us to at least publish one additional crate (build) dependency
> 
> I don't think that's the only solution. It could be solved with symlinks, [path attribute](https://doc.rust-lang.org/beta/reference/items/modules.html#the-path-attribute) or [include!](https://doc.rust-lang.org/std/macro.include.html). All of those could be used to pull the `aws-lc-sys` code in to the fips crate. But, yes, you would need some automation in place to ensure that source is copied before publishing.

Can path-attributes reference files outside the crate root, same question for `include!`? I think a file system symlink would be a non-started. One primary concern is requiring some build process to even make the sys crates use-able. When they were located in aws-lc we had the complexity of having template files and having to generate the crate each time to make it usable. The sys crates were ephemeral and we didn't have the ability to track changes to them. The benefit we have today is that you only need to initialize the repository git submodules, and then the sys crates are use-able as-is.  I think you could still have such a world where you commit the generated solution to the repository, but I think it makes developing changes to the build scripts painful (in my opinion at least).

Anyways we don't need to solve that here, I can make a corresponding PR for the aws-lc-fips-sys crate to go along with this.

aws-lc-sys-testing/build.rs
```diff
@@ -0,0 +1,19 @@
...
+
+fn main() {
+    // ensure that the include path is exported and set up correctly
+    cc::Build::new()
+        .include(env("DEP_AWS_LC_0_12_0_INCLUDE"))
+        .file("src/testing.c")
+        .compile("aws_ls_sys_testing");
+
+    // ensure the libcrypto artifact is linked
+    println!("cargo:rustc-link-lib=aws_lc_0_12_0_crypto");
```
skmcgrail, line 12: I'd say just to reduce the number of times we need to track this version number add a build dependency on toml 0.8.

```suggestion
let content = read_to_string("../aws-lc-sys/Cargo.toml").unwrap();

    let table = content.parse::<Table>().unwrap();

    let name = table["package"]
        .as_table()
        .unwrap()
        .get("links")
        .unwrap()
        .as_str()
        .unwrap();

    // ensure that the include path is exported and set up correctly
    cc::Build::new()
        .include(env(format!("DEP_{}_INCLUDE", name.to_uppercase())))
        .file("src/testing.c")
        .compile("aws_ls_sys_testing");

    // ensure the libcrypto artifact is linked
    println!("cargo:rustc-link-lib={}_crypto", name);
```

camshaft: Fixed

skmcgrail (approved):

justsmth (approved):

Issue #300, opened 2023-12-18
HPKE base mode support or ephemeral-static ECDH API


ctz:
### Problem:

Over in rustls, we'd like to implement Encrypted Client Hello. This uses [HPKE "Base" mode](https://datatracker.ietf.org/doc/rfc9180/). We've defined traits to generalise over this use, and would like to implement these traits on top of aws-lc-rs. HPKE is a pretty straightforward construction and I think the existing AEAD and HKDF APIs can be reused for this. Unfortunately the ECDH API cannot, because it only supports ephemeral-ephemeral key exchange -- the server key in ECH is static.

### Solution:

I think there are two possible solutions:

1. aws-lc-rs exposes HPKE base mode as a first-class API, and we use this directly.
2. aws-lc-rs exposes an ephemeral-static ECDH API (and we do the glueing-together and known-answer testing in rustls.)

For reference these have issues/PRs in *ring*:

1. https://github.com/briansmith/ring/pull/1462
2. https://github.com/briansmith/ring/issues/331

justsmth:
Seeing as AWS-LC has [API to support HPKE](https://github.com/aws/aws-lc/tree/main/crypto/hpke), we'll pursue "option 1" and provide a first-class API for HPKE.

justsmth:
A PR for "option 2" is here: https://github.com/aws/aws-lc-rs/pull/302

justsmth:
Hi @ctz -- as of [aws-lc-rs v1.6.0](https://crates.io/crates/aws-lc-rs/1.6.0) we support agreement with non-ephemeral keys.  Does this resolve this issue for you?

cpu:
:wave: Hi folks,

Just wanted to report back that we implemented [HPKE in Rustls](https://github.com/rustls/rustls/blob/835c17c6b840c10d1c4b656f4a1b220a285d2c79/rustls/src/crypto/aws_lc_rs/hpke.rs) using the non-ephemeral key agreement support added in 1.6.0+ - thanks again! 

In case it's interesting I also bumped into a challenge w.r.t the HKDF API: in short, the `*ring*` inspired API is fairly tightly coupled to the TLS 1.3 use-case in which the PRK extracted using the HKDF `extract` step can only be used directly for an `expand` (via the returned `Prk`'s `Prk::expand` fn) - the extracted PRK is opaque/inaccessible to the caller. 

[...]

PR #302, opened 2023-12-20
Persistent private keys for agreement

justsmth:
### Issues:
https://github.com/aws/aws-lc-rs/issues/300

### Description of changes: 
* Allow PrivateKeys to persist across multiple agreements.
* Support seralizing/deserializing `agreement::PrivateKey`

### Call-outs:
* Mostly a rename of `EphemeralPrivateKey` to `PrivateKey`.  I made `EphemeralPrivateKey` a wrapper around `PrivateKey` and moved it to another file.

### Testing:
* Augmented existing tests for coverage.


hansonchar:
Nice!  Thanks for this PR.

aws-lc-rs/src/agreement/ephemeral.rs
```diff
@@ -0,0 +1,462 @@
...
+        } else {
+            panic!("Unsupported curve: {curve_name}");
+        }
+    }
+
+    fn x25519(private_key: &[u8], public_key: &[u8]) -> Vec<u8> {
+        x25519_(private_key, public_key).unwrap()
+    }
+
+    fn x25519_(private_key: &[u8], public_key: &[u8]) -> Result<Vec<u8>, Unspecified> {
```
skmcgrail, line 453: Personally not a fan of the `_` suffix, can we call it `try_x25519`?

```suggestion
    fn try_x25519(private_key: &[u8], public_key: &[u8]) -> Result<Vec<u8>, Unspecified> {
```

Or just combine the functions as it appears `x25519_` is not used by the tests directly?

aws-lc-rs/src/agreement/ephemeral.rs
```diff
@@ -0,0 +1,462 @@
...
+                        .is_err());
+                    }
+                }
+
+                Ok(())
+            },
+        );
+    }
+
+    fn h(s: &str) -> Vec<u8> {
```
skmcgrail, line 426:
```suggestion
    fn from_hex(s: &str) -> Vec<u8> {
```

aws-lc-rs/src/agreement.rs
```diff
@@ -218,40 +259,75 @@ impl EphemeralPrivateKey {
...
+        alg: &'static Algorithm,
+        key_bytes: &[u8],
+    ) -> Result<Self, KeyRejected> {
+        let evp_pkey = if AlgorithmID::X25519 == alg.id {
+            LcPtr::new(unsafe {
+                EVP_PKEY_new_raw_private_key(
+                    EVP_PKEY_X25519,
+                    null_mut(),
+                    key_bytes.as_ptr(),
+                    X25519_PRIVATE_KEY_LEN,
```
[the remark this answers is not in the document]
justsmth: In the latest commit, the constant key length values should all be accessed via the enum functions.

skmcgrail (dismissed):

aws-lc-rs/src/encoding.rs
```diff
@@ -0,0 +1,58 @@
...
+pub trait AsBigEndian<T> {
+    /// Serializes into a big-endian format.
+    ///
+    /// # Errors
+    /// Returns Unspecified if serialization fails.
+    fn as_be_bytes(&self) -> Result<T, crate::error::Unspecified>;
+}
+
+/// Elliptic curve private key data encoded as a big-endian fixed-length integer.
+pub type EcPrivateKeyBin = Buffer<'static, EcPrivateKeyBinType>;
```
skmcgrail, line 49: Should the lifetime parameter still be defined on `EcPrivateKeyBin` to give us flexibility?

justsmth: So far the types defined here are all `'static`. I don't foresee a situation where we would need to provide serialization in a form that required a lifetime, and if we did, we could define a separate type for that.

skmcgrail: The only possibility I could see us needing it is if we wanted to have pre-computed encodings (like we did in other instance of the library today). I'd like to move away from that anyways, but I also want to avoid having plethora of types because we were overly aggressive on not exposing lifetime parameters in the API.

aws-lc-rs/src/ec.rs
```diff
@@ -323,18 +319,40 @@ pub(crate) unsafe fn marshal_private_key_to_buffer(
...
+    let evp_pkey = LcPtr::new(aws_lc::d2i_PrivateKey(
+        EVP_PKEY_EC,
+        &mut out,
+        &mut key_bytes.as_ptr(),
+        key_bytes
+            .len()
+            .try_into()
+            .map_err(|_| KeyRejected::too_large())?,
+    ))?;
+    validate_evp_key(&evp_pkey.as_const(), nid)?;
```
andrewhop, line 336: I think this is redundant, [d2i_PrivateKey](https://github.com/aws/aws-lc/blob/main/crypto/evp_extra/p_ec_asn1.c#L264) ends up calling [EC_KEY_parse_private_key](https://github.com/aws/aws-lc/blob/main/crypto/evp_extra/p_ec_asn1.c#L158) which calls [EC_KEY_check_key](https://github.com/aws/aws-lc/blob/main/crypto/ec_extra/ec_asn1.c#L188).

aws-lc-rs/src/ec/key_pair.rs
```diff
@@ -305,42 +285,30 @@ impl Debug for PrivateKey<'_> {
...
+                self.0.algorithm.id.private_key_size(),
                 &self.0.evp_pkey.as_const(),
             )?;
             Ok(EcPrivateKeyBin::new(buffer))
         }
     }
 }
 
-impl AsDer<EcPrivateKeyRfc5915Der> for PrivateKey<'_> {
+impl AsDer<EcPrivateKeyRfc5915Der<'static>> for PrivateKey<'_> {
```
andrewhop, line 306: What does adding `'static` do?

aws-lc-rs/src/agreement.rs
```diff
@@ -368,10 +434,80 @@ impl EphemeralPrivateKey {
...
+    /// `error::Unspecified`  if serialization failed.
+    fn as_der(&self) -> Result<EcPrivateKeyRfc5915Der<'static>, Unspecified> {
+        if AlgorithmID::X25519 == self.inner_key.algorithm().id {
+            return Err(Unspecified);
+        }
+
+        let mut outp = null_mut::<u8>();
+        let ec_key = {
+            ConstPointer::new(unsafe {
+                EVP_PKEY_get0_EC_KEY(self.inner_key.get_evp_pkey().as_const_ptr())
```
andrewhop, line 452: Should this use EVP_PKEY_get1_EC_KEY/EC_KEY_free so the ref count is incremented? In theory AWS-LC could free this key since get0 doesn't increment the ref count. Disclaimer: I'm also not sure how Rust + aws-lc-rs play with threads and if anything can free `PrivateKey` concurrently to this call.

andrewhop: Actually I think get0 makes sense because the overall EVP key will live as long as this function.

aws-lc-rs/src/agreement.rs
```diff
@@ -218,59 +251,94 @@ impl EphemeralPrivateKey {
...
+        alg: &'static Algorithm,
+        key_bytes: &[u8],
+    ) -> Result<Self, KeyRejected> {
+        let evp_pkey = if AlgorithmID::X25519 == alg.id {
+            LcPtr::new(unsafe {
+                EVP_PKEY_new_raw_private_key(
+                    EVP_PKEY_X25519,
+                    null_mut(),
+                    key_bytes.as_ptr(),
+                    AlgorithmID::X25519.private_key_len(),
```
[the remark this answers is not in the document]
justsmth: Good catch. We need to verify that the slice provided is the appropriate length.

aws-lc-rs/src/ec/key_pair.rs
justsmth: In this case, `'static` essentially means that the value "owns" all of the data it references. Specifically, the [`Cow`](https://doc.rust-lang.org/std/borrow/enum.Cow.html) value that the `Buffer` (thinly) wraps must be a [`Cow::Owned`](https://doc.rust-lang.org/std/borrow/enum.Cow.html#variant.Owned)

aws-lc-rs/src/ec.rs
justsmth: Removed in the next commit.

skmcgrail (dismissed):

aws-lc-rs/src/agreement.rs
```diff
@@ -218,59 +251,97 @@ impl EphemeralPrivateKey {
...
         }
+        let evp_pkey = unsafe { ec::unmarshal_der_to_private_key(key_bytes, alg.id.nid())? };
+        Ok(Self::new(alg, evp_pkey))
+    }
+
+    /// Constructs an ECDH key from private key bytes
+    ///
+    /// The private key must encoded as a big-endian fixed-length integer. For
+    /// example, a P-256 private key must be 32 bytes prefixed with leading
+    /// zeros as needed.
```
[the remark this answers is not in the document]
justsmth: Yeah, this expects the "raw" encoding of an EC private key. For X25519, [RFC 7748 section 5](https://datatracker.ietf.org/doc/html/rfc7748#section-5) describes its encoding. For P-256, P384, P512, [RFC 5915 section 3](https://datatracker.ietf.org/doc/html/rfc5915#section-3) describes the format.

aws-lc-rs/src/agreement.rs
```diff
@@ -368,10 +437,80 @@ impl EphemeralPrivateKey {
...
+    /// `error::Unspecified`  if serialization failed.
+    fn as_der(&self) -> Result<EcPrivateKeyRfc5915Der<'static>, Unspecified> {
+        if AlgorithmID::X25519 == self.inner_key.algorithm().id {
+            return Err(Unspecified);
+        }
+
+        let mut outp = null_mut::<u8>();
+        let ec_key = {
+            ConstPointer::new(unsafe {
+                EVP_PKEY_get0_EC_KEY(self.inner_key.get_evp_pkey().as_const_ptr())
```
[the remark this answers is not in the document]
justsmth: If the EVP_PKEY doesn't contain an EC_KEY, then AWS-LC would (after setting the error `EVP_R_EXPECTING_AN_EC_KEY_KEY`) [return NULL](https://github.com/aws/aws-lc/blob/7226be17c2268f5c44cd110dc7e0580903ab9763/crypto/fipsmodule/evp/evp.c#L389), causing `ConstPointer::new` to also return an `Err`.

aws-lc-rs/src/agreement.rs
```diff
@@ -218,59 +251,97 @@ impl EphemeralPrivateKey {
...
+        }
+        let evp_pkey = if AlgorithmID::X25519 == alg.id {
+            LcPtr::new(unsafe {
+                EVP_PKEY_new_raw_private_key(
+                    EVP_PKEY_X25519,
+                    null_mut(),
+                    key_bytes.as_ptr(),
+                    AlgorithmID::X25519.private_key_len(),
+                )
+            })?
```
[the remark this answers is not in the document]
justsmth: This function only returns KeyRejected variants as errors. Unless a specific one is named (like `invalid_encoding` below), most underlying errors will get transformed into [KeyRejected::unexpected_error()](https://github.com/aws/aws-lc-rs/blob/516ffefc2da1614206cd7e1ce8a691cb102cdd80/aws-lc-rs/src/error.rs#L216). Unfortunately, we don't have much visibility/granularity around why AWS-LC has failed to parse a key. We previously tried to be aligned with the same errors `ring` provides, but we now deviate in many places.

aws-lc-rs/src/agreement/ephemeral.rs
```diff
@@ -0,0 +1,462 @@
...
+        expect_iterated_x25519(
+            "684cf59ba83309552800ef566f2f4d3c1c3887c49360e3875f2eb94d99532c51",
+            1..1_000,
+            &mut k,
+            &mut u,
+        );
+
+        // The spec gives a test vector for 1,000,000 iterations but it takes
+        // too long to do 1,000,000 iterations by default right now. This
+        // 10,000 iteration vector is self-computed.
```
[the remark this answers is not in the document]
justsmth: Our current tests (include doc-tests) only take ~14 seconds on my Mac, so taking an additional 32s for this would be excessive.  We do have some tests that only run for a `--release` build, so we could potentially have more extensive testing there.  But, for this case, I don't think it's needed.

aws-lc-rs/src/agreement.rs
```diff
@@ -307,13 +307,11 @@ impl PrivateKey {
                 )
             })?
         } else {
-            let ec_group = unsafe { ec_group_from_nid(alg.id.nid())? };
-            let private_bn = DetachableLcPtr::try_from(key_bytes)?;
+            let ec_group: LcPtr<EC_GROUP> = ec_group_from_nid(alg.id.nid())?;
```
skmcgrail, line 310: Minor but `ec_group_from_nid` doesn't technically need this since you aren't trying to convert to the type like the line right below it. Not really picky, just wasn't sure if you were doing it for clarity or had expected different.

skmcgrail (approved):

skmcgrail (approved):

andrewhop (approved):

Issue #296, opened 2023-12-13
feature request: RSA key generation support


cpu:
### Problem:

It looks as though aws-lc-rs shares the same API limitation as *ring* with respect to key generation being supported for [EcdsaKeyPair](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/signature/struct.EcdsaKeyPair.html) and [Ed25519KeyPair](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/signature/struct.Ed25519KeyPair.html), but **not** [RsaKeyPair](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/signature/struct.RsaKeyPair.html).

As we're evaluating using aws-lc-rs as [a backend for rcgen](https://github.com/rustls/rcgen/pull/201) it would be really nice if aws-lc-rs supported RSA key generation. There's a class of use-case (e.g. Rustls' integration tests) where we have to use the `openssl` command line in order to generate RSA keypairs. Having support for this in `rcgen` would make replacing that usage feasible.

### Solution:

aws-lc-rs could augment the `RsaKeyPair` struct to offer methods for generating a keypair comparable to those offered by `Ed25519KeyPair` and `EcdsaKeyPair`.

* **Does this change any public APIs?** Yes: the `RsaKeyPair` struct would gain a `generate_pkcs8` and possibly a `generate_pkcs1`  fns.
* **Which algorithm(s) will this impact?** RSA.

### Requirements / Acceptance Criteria:

It must be possible to generate an RSA keypair with both the public and private components. It should be convenient to load the serialized private key using the existing `RsaKeyPair::from_pkcs8` and `RsaKeyPair::from_der` methods. Acceptance should look something like an end-to-end test generating a RSA keypair, loading it, using it to produce a signature, and then verifying the signature.

* **RFC links:** https://www.rfc-editor.org/rfc/rfc8017, https://www.rfc-editor.org/rfc/rfc5208
* **Related Issues:** https://github.com/rustls/rcgen/pull/201, https://github.com/briansmith/ring/issues/219, https://github.com/briansmith/ring/pull/733
* **Will the Usage Guide or other documentation need to be updated?** Yes - API documentation would need updating.
* **Testing:** I expect testing would look similar to the existing Ed25519KeyPair and EcdsaKeyPair key generation test coverage.

skmcgrail:
I have a branch for RSA OAEP support that also adds RSA key generation. I just need to rebase that branch and open a PR for the change.

PR #303, opened 2023-12-20
RSA OAEP Support

skmcgrail:
### Description of changes: 
Implements support for RSA OAEP encryption and decryption to support usage in a broader key-transport scheme. Additionally adds support for RSA key generation (#296).


aws-lc-rs/src/rsa/oaep.rs
```diff
@@ -0,0 +1,501 @@
...
+        }) {
+            return Err(Unspecified);
+        };
+
+        Ok(&mut output[..out_len])
+    }
+}
+
+impl AsDer<Pkcs8V1Der> for PrivateDecryptingKey {
+    fn as_der(&self) -> Result<Pkcs8V1Der, Unspecified> {
```
skmcgrail, line 208: Would be good to get some feedback on this `Pkcs8V1Der` type, and this pattern of using `AsDer` trait. Formally on other existing API types we would have used `to_pkcs8` / `to_pkcs8v1` but this was prior to the introduction of the AsDer and corresponding buffer type.

I'm conflicted on trying to deprecate the `pkcs8::Document` type, but it would be nice to distinguish V1 and V2 types in a manner that could make usage with AsDer more compatible. Alternatively we introduce a type alias for Buffer and using generic in that manner, but then that seems rather unnecessary, the AsDer contract is generic and not specific to using a buffer type.

Either way we

aws-lc-rs/src/rsa/oaep.rs
```diff
@@ -0,0 +1,501 @@
...
+    OaepSha256Mgf1sha256,
+
+    /// RSA-OAEP with SHA384 Hash and SHA384 MGF1
+    OaepSha384Mgf1sha384,
+
+    /// RSA-OAEP with SHA512 Hash and SHA512 MGF1
+    OaepSha512Mgf1sha512,
+}
+
+type HashFn = unsafe extern "C" fn() -> *const EVP_MD;
```
justsmth, line 73: These are both types of hash/digest functions. Would `OaepDigestFn` and `Mgf1DigestFn` be more appropriate names?

aws-lc-rs/src/rsa/key.rs
```diff
@@ -31,14 +33,65 @@ use super::{
...
+    #[inline]
+    fn bit_len(self) -> i32 {
+        match self {
+            KeySize::Rsa2048 => 2048,
+            KeySize::Rsa4096 => 4096,
+            KeySize::Rsa8192 => 8192,
+        }
+    }
+
+    pub(super) fn from_evp_pkey(evp_pkey: &LcPtr<EVP_PKEY>) -> Result<Self, Unspecified> {
```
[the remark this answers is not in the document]
skmcgrail: It's used in the private `new` constructor functions for `PrivateDecryptingKey` and `PublicEncryptingKey` which take a ` LcPtr<EVP_PKEY>`.

aws-lc-rs/src/rsa/oaep.rs
```diff
@@ -0,0 +1,472 @@
...
+
+/// An RSA Public Key used for decrypting ciphertext encrypted by [`PublicEncryptingKey`].
+pub struct PublicEncryptingKey {
+    key: LcPtr<EVP_PKEY>,
+    size: KeySize,
+}
+
+impl PublicEncryptingKey {
+    fn new(key: LcPtr<EVP_PKEY>) -> Result<Self, Unspecified> {
+        if !is_rsa_evp_pkey(&key) {
```
justsmth, line 219: * This check is called both here and in `KeySize::from_evp_pkey`.  
* To avoid some of the redundancy between `PublicEncryptingKey` and `PrivateDecryptingKey` you could make them wrappers around a common type: 
```rust
pub struct PublicEncryptingKey(RsaEvpPkey);

impl PublicEncryptingKey {
    fn new(key: LcPtr<EVP_PKEY>) -> Result<Self, Unspecified> {
        Ok(Self(RsaEvpPkey::new(key)?))
    }
...
}

pub struct PrivateDecryptingKey(RsaEvpPkey)

impl PrivateDecryptingKey {
    fn new(key: LcPtr<EVP_PKEY>) -> Result<Self, Unspecified> {
        Ok(Self(RsaEvpPkey::new(key)?))
    }
...
}


pub(crate) struct RsaEvpPkey {
    key: LcPtr<EVP_PKEY>,
    size: KeySize,
}
 impl RsaEvpPkey {
     pub(crate) fn new(key: LcPtr<EVP_PKEY>) -> Result<Self, Unspecified> {
        if EVP_PKEY_RSA != unsafe { EVP_PKEY_id(key.as_const_ptr()) } {
            return Err(Unspecified);
        }
        let size = match unsafe { EVP_PKEY_size(key.as_const_ptr()) } {
            256 => Ok(KeySize::Rsa2048),
            512 => Ok(KeySize::Rsa4096),
            1024 => Ok(KeySize::Rsa8192),
            _ => Err(Unspecified),
        }?;
        Ok(Self { key, size })
    }
...
}
```

aws-lc-rs/src/rsa/key.rs
```diff
@@ -3,43 +3,94 @@
...
+                    Self::Rsa8192 => 8192,
+                }
+            }
+        }
+    };
+}
+
+pub(super) use rsa_key_size_enum;
+
+rsa_key_size_enum!(SignatureKeySize);
```
justsmth, line 93: * Why do we need to distinguish `EncryptionKeySize` from `SignatureKeySize`?
* Should this instead be in "signature.rs"?  I think either both `EncryptionKeySize` and `SignatureKeySize` should be here, or neither.

aws-lc-rs/src/rsa/key.rs
```diff
@@ -3,43 +3,94 @@
...
+                    Self::Rsa2048 => 256,
+                    Self::Rsa3072 => 384,
+                    Self::Rsa4096 => 512,
+                    Self::Rsa8192 => 1024,
+                }
+            }
+
+            /// Returns the key size in bits.
+            #[inline]
+            fn bit_len(self) -> i32 {
```
justsmth, line 79: Why not return `usize`?

aws-lc-rs/src/rsa/key.rs
skmcgrail: Originally I had it as a single KeySize type shared between the two. I wasn't sure if there would be a case in the future where we would want to add size asymmetrically to one and not the other. Kind of a moot points given that nothing stops you from using one RSA key bytes with the other. I can recombine them into a single type.

aws-lc-rs/src/rsa/key.rs
skmcgrail: If I recall the aws-lc function takes an i32/int, so I just coded this private function to match for simplicity.

aws-lc-rs/src/rsa/key.rs
```diff
@@ -486,16 +454,287 @@ where
...
+
+    pub fn key_size(&self) -> usize {
+        // Safety: RSA modulous byte sizes supported fit an usize
+        unsafe { EVP_PKEY_size(self.key.as_const_ptr()) }
+            .try_into()
+            .expect("modulous should fit in usize")
+    }
+}
+
+impl AsDer<RsaPublicKeyX509Der<'_>> for RsaEvpPkey {
```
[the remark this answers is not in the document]
skmcgrail: I discussed the issues of doing this before offline as I had the same thought originally. The issue with that is the need for potentially one or more constructor arguments in addition to the raw key bytes in order to configure a given type.

skmcgrail:
Moving this to draft for the time being. Currently splitting out the key generation from the OAEP support to reduce the feature scope for our next release.

aws-lc-rs/src/rsa/encoding.rs
```diff
@@ -0,0 +1,138 @@
...
+        ptr::{DetachableLcPtr, LcPtr},
+    };
+    use aws_lc::{
+        EVP_PKEY_assign_RSA, EVP_PKEY_new, RSA_parse_private_key, RSA_parse_public_key,
+        RSA_public_key_to_bytes, EVP_PKEY,
+    };
+    use std::ptr::null_mut;
+
+    /// DER encode a RSA public key to `RSAPublicKey` structure.
+    pub(in super::super) unsafe fn encode_public_key_der(
```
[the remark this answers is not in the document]
skmcgrail: For private key, we previously have only supported decoding the format due to the fact that `KeyPair` has never supported encoding of the private key to a pkcs8 document or to the RSAPrivateKey ASN.1 structure. My interpretation of the reasoning behind that is due to the fact that ring never supported generation of RSA keys, there was really no reason to serialize a private key that was already read from some serialized bytes.

Now that we have introduced support for `KeyPair` generation, and now `PrivateDecryptingKey` keypair generation it obviously makes sense that we add support for serialization of the private key. Currently that is PKCS#8 v1 document format that we support for both types of keys mentioned previously. I don't think it makes sense to add support for this `RSAPrivateKey` ASN.1 serialization format until we have an interoperability story that requires it. My primary beef with its usage is that the format is not self describing like PKCS#8 that properly annotates that the type of private key document that is embedded (A topic I know I've discussed in length offline). 

So TL;DR I'd rather support a better format like PKCS#8 for private key serialization, until there is a proper usage requirement that necessitates RSAPrivateKey format for interoperability with other libraries.

skmcgrail:
I took a pass at implemtning the initial feedback, just need to circle back on seeing if we can provide a helper utility or two for calculating proper ciphertext and plaintext buffer sizes and/or documentation to explain the OAEP size expectations.

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,359 @@
...
+            UsageContext::Encryption,
+        )?))
+    }
+
+    /// Decrypts the contents in `ciphertext` and writes the corresponding plaintext to `output`.
+    ///
+    /// # Sizing `output`
+    /// For `OAEP_SHA1_MGF1SHA1`, `OAEP_SHA256_MGF1SHA256`, `OAEP_SHA384_MGF1SHA384`, `OAEP_SHA512_MGF1SHA512` a safe
+    /// `output` length is the key size in bytes. The RSA key size in bytes can be retrieved using [`Self::key_size`].
+    /// The actual output length will at most `rsaKeySizeBytes - (2 * hashLengthBytes) - 2`.
```
justsmth, line 196: There are two hashes involved, and currently they're the same. If the hashes were to differ, I believe it would be the output length of OAEP hash that determines the length here. Should we clarify: `rsaKeySizeBytes - (2 * oaepHashLengthBytes) - 2`?

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,359 @@
...
+
+    /// Returns the RSA key size in bytes.
+    #[must_use]
+    pub fn key_size(&self) -> usize {
+        self.0.key_size()
+    }
+
+    /// Encrypts the contents in `plaintext` and writes the corresponding ciphertext to `output`.
+    ///
+    /// # Sizing `output`
```
justsmth, line 278: The max size of `input` is pertinent here as well.

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,359 @@
...
+            UsageContext::Encryption,
+        )?))
+    }
+
+    /// Returns the RSA key size in bytes.
+    #[must_use]
+    pub fn key_size(&self) -> usize {
+        self.0.key_size()
+    }
+
```
justsmth, line 275: We should provide a utility function for consumers:
```
pub fn max_plaintext_size(&self, algorithm: &'static EncryptionAlgorithm) -> usize { ... }
```

I don't think we need one for ciphertext size as long as the documentation is clear that it matches the key size.

aws-lc-rs/src/rsa.rs
```diff
@@ -4,24 +4,89 @@
...
+//! let public_key_der_bytes = public_key_der.as_ref();
+//!
+//! // Load a RSA public key from DER encoded X.509 SubjectPublicKeyInfo.
+//! let public_key = PublicEncryptingKey::from_der(public_key_der_bytes)?;
+//! 
+//! // Construct a RSA-OAEP public encrypting key
+//! let public_key = OaepPublicEncryptingKey::new(public_key)?;
+//!
+//! let message = b"hello world";
+//! let mut ciphertext = vec![0u8; KeySize::Rsa2048.len()]; // Output will be the size of the RSA key length in bytes rounded up.
```
justsmth, line 42: I think it would be good to also provide a `max_ciphertext_size` (as an alias for a `key_size_bytes`?) so this line becomes:
```
let mut ciphertext = vec![0u8; public_key.max_ciphertext_size()];
```

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,474 @@
...
+    /// Returns a boolean indicator if this RSA key is an approved FIPS 140-3 key.
+    #[cfg(feature = "fips")]
+    #[must_use]
+    pub fn is_valid_fips_key(&self) -> bool {
+        super::key::is_valid_fips_key(&self.0)
+    }
+
+    /// Returns the RSA key size in bytes.
+    #[must_use]
+    pub fn key_size(&self) -> usize {
```
justsmth, line 168: Since RSA "key size" is almost always talked about in "bits", calling this `key_size` is ambiguous. I think we should provide both `key_size_bits` and `key_size_bytes`.

aws-lc-rs/src/rsa.rs
```diff
@@ -4,24 +4,89 @@
...
+//!
+//! // Encrypt a message with the public key.
```
justsmth, line 44: NP: For clarity mention the optional label argument?
```
//! // Encrypt the message with the public key w/o the optional label.
```

aws-lc-rs/src/rsa.rs
```diff
@@ -4,24 +4,89 @@
...
+//! let ciphertext = public_key.encrypt(&OAEP_SHA256_MGF1SHA256, message, &mut ciphertext, None)?;
+//!
+//! assert_ne!(message, ciphertext);
+//! 
+//! // Construct a RSA-OAEP private decrypting key
+//! let private_key = OaepPrivateDecryptingKey::new(private_key)?;
+//!
+//! // Decrypt a message with the private key.
+//! let mut plaintext = vec![0u8; KeySize::Rsa2048.len()]; // Plaintext output will be at most RSA Key length bytes - 2 * HashLength − 2 bytes
```
justsmth, line 53: We should provide a `OaepPrivateDecryptingKey::max_plaintext_size` to avoid the need for consumers to remember this formula and perform the arithmetic on their own.

aws-lc-rs/src/rsa/key.rs
```diff
@@ -147,125 +137,53 @@ impl KeyPair {
...
-        // could be pulled out and invoked seperatly from the standard checks.
-        #[cfg(feature = "fips")]
-        if 1 != RSA_check_fips(*rsa as *mut RSA) {
-            return Err(KeyRejected::inconsistent_components());
+    fn validate_key(key: &LcPtr<EVP_PKEY>) -> Result<(), Unspecified> {
+        if !is_rsa_key(key) {
+            return Err(Unspecified);
+        };
+        match key_size_bits(key) {
+            2048..=8192 => Ok(()),
```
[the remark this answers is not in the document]
skmcgrail: Note this is for signature algorithms, which for Ring always limited generation of digital signatures to keys between 2048 and 4096: https://docs.rs/ring/0.16.20/ring/signature/struct.RsaKeyPair.html#method.from_pkcs8

So this is increasing the upper range on the private key size that can be used but is keeping the lower bound fixed to the same size. We can always revisit that later.

aws-lc-rs/src/rsa.rs
skmcgrail: AWS-LC requires this output buffer to be greater then or equal to the key size. So the computed formula is useless on the `OaepPrivateDecryptingKey` end of the API:

See [encrypt output buffer check](https://github.com/aws/aws-lc/blob/0d7b965b5d470bc5b77e6347555ed7bcaf8eaeff/crypto/fipsmodule/evp/p_rsa.c#L415)
and [decrypt output buffer check](https://github.com/aws/aws-lc/blob/0d7b965b5d470bc5b77e6347555ed7bcaf8eaeff/crypto/fipsmodule/evp/p_rsa.c#L447)

justsmth (commented):
There are CI failures for [clippy](https://github.com/aws/aws-lc-rs/actions/runs/8576485565/job/23507524002?pr=303#step:5:107) and [semver](https://github.com/aws/aws-lc-rs/actions/runs/8576485565/job/23507524937?pr=303#step:3:56). I'm not worried about clippy, but for semver we'll need to add back a [`rsa::KeySize::len`](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/rsa/enum.KeySize.html#method.len) before release.

Otherwise, it looks great!

aws-lc-rs/src/rsa/encoding.rs
```diff
@@ -33,7 +34,126 @@ pub(in super::super) mod pkcs8 {
     }
 
     // Supports v1 and v2 encodings through a single API entry-point.
-    pub(in super::super) fn decode_der(pkcs8: &[u8]) -> Result<LcPtr<EVP_PKEY>, KeyRejected> {
-        LcPtr::try_from(pkcs8)
+    pub(in crate::rsa) fn decode_der(pkcs8: &[u8]) -> Result<LcPtr<EVP_PKEY>, KeyRejected> {
+        let mut cbs = unsafe { cbs::build_CBS(pkcs8) };
+        let key = LcPtr::new(unsafe { EVP_parse_private_key(&mut cbs) })
+            .map_err(|()| KeyRejected::unspecified())?;
```
justsmth, line 40: If we change this to returning `KeyRejected::invalid_encoding()` it fixes the Rustls integ test failure. (I was able to verify this locally.)

justinwsmith (approved):

justsmth (approved):

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,510 @@
...
+    fn new(evp_pkey: LcPtr<EVP_PKEY>) -> Result<Self, Unspecified> {
+        Self::validate_key(&evp_pkey)?;
+        Ok(Self(evp_pkey))
+    }
+
+    fn validate_key(key: &LcPtr<EVP_PKEY>) -> Result<(), Unspecified> {
+        if !is_rsa_key(key) {
+            return Err(Unspecified);
+        };
+        match key_size_bits(key) {
```
andrewhop, line 219: Is size the only thing worth checking? What about calling AWS-LC's validate which does a few more checks for the public/private key.

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,510 @@
...
+    #[must_use]
+    pub fn key_size_bytes(&self) -> usize {
+        key_size_bytes(&self.0)
+    }
+
+    /// Returns the RSA key size in bits.
+    #[must_use]
+    pub fn key_size_bits(&self) -> usize {
+        key_size_bits(&self.0)
+    }
```
andrewhop, line 243: Same size comment

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,510 @@
...
+        // Safety: Don't pass zero-length slice pointers to C code :)
+        if 1 != unsafe { EVP_PKEY_CTX_set0_rsa_oaep_label(**evp_pkey_ctx, null_mut(), 0) } {
+            return Err(Unspecified);
+        }
+        return Ok(());
+    }
+
+    // AWS-LC takes ownership of the label memory, and will call OPENSSL_free, so we are forced to copy it for now.
+    let label_ptr =
+        DetachableLcPtr::<u8>::new(unsafe { OPENSSL_malloc(size_of_val(label)) }.cast())?;
```
andrewhop, line 484: NP: is it possible o avoid this duplication and tell rust to release the original label that was passed in?

aws-lc-rs/src/rsa/key.rs
```diff
@@ -146,126 +136,54 @@ impl KeyPair {
...
-    ///
-    /// ```sh
-    /// openssl pkcs8 -nocrypt -outform der \
-    ///     -in rsa-2048-private-key.pem > rsa-2048-private-key.pk8
-    /// ```
+    /// *aws-lc-rs* does not impose the same limitations that *ring* does for
+    /// RSA keys. Thus signatures may be generated by keys that are not accepted
+    /// by *ring*. In particular:
+    /// * RSA private keys ranging between 2048-bit keys and 8192-bit keys are supported.
+    /// * The public exponent does not have a required minimum size.
```
andrewhop, line 153: Can this generate end up calling the FIPS version or is that a different Rust public API? If so the FIPS check does set a minimum/max for the public exponent.

aws-lc-rs/tests/data/rsa_from_pkcs8_tests.txt
```diff
@@ -22,67 +22,50 @@ Input = 30821242020100300d06092a864886f70d01010105000482122c30821228020100028204
 
 # RSA 512-bit key (too small)
 Input = 30820154020100300d06092a864886f70d01010105000482013e3082013a02010002410094f95f85cae61920ffca1deeb2135de7bf20fc05523198433178e74bd3090445f5089488a69bed2984425e497e255abe329129f5b166ad8b204768b8e374fbdd020301000102403e65280e088a513212105c731220ac6f190d3ea117e8efc60cdd7fbab15c106e7a9c8d9be4c0e03470742bc2d0e97233656b771be841a00e0957824c6091fc9d022100e7689ca4d01c8f0a190defcf4bd59038f01879605cba34d98a8e8cd04b00974f022100a4ce28ec8938a3012604174433260b8e0ed783fb2720c4c34abe004f7ea9ef13022100c94768d8c25bd4df07a3e3aeff5b27331f436a6fb2b814381418a1ca9b324c69022027f7efa8dde6a471a4dc9d1fe71fb4f1b0dbc7a0636ee9094bd9248e22f3e78d02201d849cc8c96c732c542a5ba82c7a7a474e05aeb5fb5689ee6a461e3ad545bd85
-Error = TooSmall
+Error = Unspecified
```
andrewhop, line 25: Why are we losing visibility into why the tests are failing?

aws-lc-rs/tests/rsa_test.rs
```diff
@@ -55,6 +59,7 @@ fn test_signature_rsa_pkcs1_sign() {
     let rng = rand::SystemRandom::new();
     test::run(
         test_file!("data/rsa_pkcs1_sign_tests.txt"),
+        // test_file!("data/debug.txt"),
```
andrewhop, line 62: ?

aws-lc-rs/tests/rsa_test.rs
```diff
@@ -332,6 +339,84 @@ generate_fips_encode_decode!(rsa3072_generate_fips_encode_decode, KeySize::Rsa30
 generate_fips_encode_decode!(rsa4096_generate_fips_encode_decode, KeySize::Rsa4096);
 generate_fips_encode_decode!(rsa8192_generate_fips_encode_decode, KeySize::Rsa8192, false);
 
+macro_rules! encryption_generate_encode_decode {
```
andrewhop, line 342: How hard would it be to add interoperability tests here with ring or other rust libraries?

aws-lc-rs/tests/rsa_test.rs
```diff
@@ -340,3 +425,406 @@ fn public_key_components_clone_debug() {
...
+
+                public_key
+                    .encrypt($alg, &message, ciphertext.as_mut(), None)
+                    .expect_err("plaintext too large");
+            }
+        }
+    };
+}
+
+round_trip_algorithm!(
```
andrewhop, line 588: NP: nested loops? Or are there some invalid combinations?

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,510 @@
...
+use core::{fmt::Debug, mem::size_of_val, ptr::null_mut};
+use mirai_annotations::verify_unreachable;
+
+use super::{
+    encoding,
+    key::{generate_rsa_key, is_rsa_key, key_size_bits, key_size_bytes},
+    KeySize,
+};
+
+/// RSA-OAEP with SHA1 Hash and SHA1 MGF1
```
andrewhop, line 25: Question: can customers define there own mixed modes? Or are these the only ones supported by aws-lc-rs? For the record I like starting with only theses since matching hashes is the best practice though other libraries support mixing so we might need them in the future for compatibility.

aws-lc-rs/src/rsa.rs
```diff
@@ -4,24 +4,90 @@
...
+//! ```rust
+//! # use std::error::Error;
+//! # fn main() -> Result<(), Box<dyn Error>> {
+//! use aws_lc_rs::{
+//!     encoding::{AsDer, Pkcs8V1Der, PublicKeyX509Der},
+//!     rsa::{KeySize, OAEP_SHA256_MGF1SHA256, OaepPublicEncryptingKey, OaepPrivateDecryptingKey, PublicEncryptingKey, PrivateDecryptingKey}
+//! };
+//!
+//! // Generate a RSA 2048-bit key.
+//! let private_key = PrivateDecryptingKey::generate(KeySize::Rsa2048)?;
```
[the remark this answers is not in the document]
skmcgrail: Signing key is handled by the poorly named [`KeyPair`](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/rsa/struct.KeyPair.html), but we want to encourage users to use separate signing/verficiation keys from encryption/decryption uses.

aws-lc-rs/src/rsa.rs
```diff
@@ -4,24 +4,90 @@
...
+//!
+//! // Serialize the RSA public key to DER encoded X.509 SubjectPublicKeyInfo for later usage.
+//! let public_key_der = AsDer::<PublicKeyX509Der>::as_der(&public_key)?;
+//! let public_key_der_bytes = public_key_der.as_ref();
+//!
+//! // Load a RSA public key from DER encoded X.509 SubjectPublicKeyInfo.
+//! let public_key = PublicEncryptingKey::from_der(public_key_der_bytes)?;
+//!
+//! // Construct a RSA-OAEP public encrypting key
+//! let public_key = OaepPublicEncryptingKey::new(public_key)?;
```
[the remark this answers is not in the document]
skmcgrail: More consistency with the rest of the APIs we have inherited. Having the encryption algorithm baked into the type also makes it less likely to use incorrectly in an application, as there aren't two different `encrypt` function present on a single type then that output very different ciphertexts.

aws-lc-rs/src/rsa/encoding.rs
```diff
@@ -33,7 +34,126 @@ pub(in super::super) mod pkcs8 {
...
+    }
+
+    /// Decode a DER encoded `RSAPublicKey` structure.
+    #[inline]
+    pub(in crate::rsa) fn decode_public_key_der(
+        public_key: &[u8],
+    ) -> Result<LcPtr<EVP_PKEY>, KeyRejected> {
+        let mut cbs = unsafe { cbs::build_CBS(public_key) };
+
+        let rsa = DetachableLcPtr::new(unsafe { RSA_parse_public_key(&mut cbs) })?;
```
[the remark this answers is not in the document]
skmcgrail: Good catch!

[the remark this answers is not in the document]
skmcgrail: It returns an `Result::Err(..)` type that will be returned from this function instead. The `?` is an error propagation, so its `Result::Ok(...)` then the type inside `Ok` will be assigned to the left, otherwise this will return the error.

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,510 @@
...
+        Ok(Self(evp_pkey))
+    }
+
+    fn validate_key(key: &LcPtr<EVP_PKEY>) -> Result<(), Unspecified> {
+        if !is_rsa_key(key) {
+            return Err(Unspecified);
+        };
+        match key_size_bits(key) {
+            2048..=8192 => Ok(()),
+            _ => Err(Unspecified),
```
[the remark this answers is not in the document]
skmcgrail: Just following the BR guidance on only using 2048-bit or higher RSA keys with OAEP.

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,510 @@
...
+        }
+    }
+
+    /// Generate a new RSA private key for use with asymmetrical encryption.
+    ///
+    /// # Errors
+    /// * `Unspecified` for any error that occurs during the generation of the RSA keypair.
+    pub fn generate(size: KeySize) -> Result<Self, Unspecified> {
+        let key = generate_rsa_key(size.bits(), false)?;
+        Self::new(key)
```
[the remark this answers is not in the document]
skmcgrail: `KeySize` is an enum we define and it only allows for 2048,  3072, 4096, 8192 keys. So it is not an arbitrary input value from the user.

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,510 @@
...
+    }
+
+    /// Generate a RSA `KeyPair` of the specified key-strength.
```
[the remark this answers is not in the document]
skmcgrail: Updated them to match in the description. With the one exception that the FIPS generation logic doesn't support larger then 4096.

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,510 @@
...
+    }
+
+    /// Returns a boolean indicator if this RSA key is an approved FIPS 140-3 key.
+    #[cfg(feature = "fips")]
+    #[must_use]
+    pub fn is_valid_fips_key(&self) -> bool {
+        super::key::is_valid_fips_key(&self.0)
+    }
+
+    /// Returns the RSA key size in bytes.
```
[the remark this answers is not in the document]
skmcgrail: > For an RSA key, this returns the number of bytes needed to represent the modulus.

This is the key size of the `n` component of RSA which is what the RSA key size refers to. Did you have a specific concern?

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,510 @@
...
+    #[must_use]
+    pub fn key_size_bytes(&self) -> usize {
+        key_size_bytes(&self.0)
+    }
+
+    /// Returns the RSA key size in bits.
```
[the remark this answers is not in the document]
skmcgrail: Same as the above comment this is the bit length of the `n` component.

aws-lc-rs/src/rsa/encryption.rs
skmcgrail: AWS-LC already perform the key validation checks when we parse a private or public key, per @dkostic we didn't need to perform any additional checks.

aws-lc-rs/src/rsa/encryption.rs
skmcgrail: See comments above.

aws-lc-rs/src/rsa/encryption.rs
skmcgrail: No there is not, and on the Rust side since we take a reference to the label content (`&[u8]`) that could actually just be a fixed slice that is `'static` lifetime so there is nothing that needs to be freed. Problem is there is no way to tell AWS-LC that we own the allocation and stop it from trying to free the memory pointer too.

aws-lc-rs/src/rsa/key.rs
skmcgrail: There is a separate `generate_fips` which calls `RSA_generate_key_fips` in AWS-LC, and the Rust `generate` calls `RSA_generate_key_ex` in AWS-LC instead, which explicitly sets doesn't has the FIPS flow path disabled (`check_fips=0`) in the AWS-LC implementation.

aws-lc-rs/tests/data/rsa_from_pkcs8_tests.txt
skmcgrail: This required duplicating logic AWS-LC validation logic checks on RSA key sizes, so in order to not repeat / duplicate checks we are opting to remove this level of specificity.  So the error is till `KeyReject` in aws-lc-rs, but the string message contain is now `Unspecified`.

aws-lc-rs/tests/rsa_test.rs
skmcgrail: Leftover cruft will remove.

aws-lc-rs/tests/rsa_test.rs
skmcgrail: The macro is preferred as it results in a test function per combination, versus a loop would bail out the test earlier and you don't get a full picture of the failures.

aws-lc-rs/src/rsa/encryption.rs
skmcgrail: No, they can only use the ones provided by aws-lc-rs. The `OaepAlgorithm` type has private struct fields with no constructor.

aws-lc-rs/tests/rsa_test.rs
skmcgrail: It's doable, but I'm not sure if aws-lc-rs is the right level to be extensively testing that at. We can explore having a smoketest / sanity test to just validate with a quick inter-opt for keys between another library, but we are just a pass-through to AWS-LC in terms of the keys it produces, and the serialization outputs and inputs it supports.

aws-lc-rs/src/rsa/encryption.rs
```diff
@@ -0,0 +1,510 @@
...
+    mgf1_hash_fn: EVP_sha512,
+};
+
+/// RSA Encryption Algorithm Identifier
+#[allow(clippy::module_name_repetitions)]
+#[derive(Debug, Clone, Copy, PartialEq)]
+#[non_exhaustive]
+pub enum EncryptionAlgorithmId {
+    /// RSA-OAEP with SHA1 Hash and SHA1 MGF1
+    OaepSha1Mgf1sha1,
```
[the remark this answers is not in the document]
skmcgrail: Actually not that I look at it again, I wanted to match the naming of the constants in the module where we mash the MGF1SHA1 together to show them as a pair. So this is technically correct if you are camel casing that as one word. Let me get Justin input and I'll push a change for this specifically if need be.

andrewhop (dismissed):

aws-lc-rs/src/rsa/encoding.rs
```diff
@@ -130,14 +130,19 @@ pub(in crate::rsa) mod rfc5280 {
...
     pub(in crate::rsa) fn encode_public_key_der(
         key: &LcPtr<EVP_PKEY>,
     ) -> Result<PublicKeyX509Der<'static>, Unspecified> {
-        let mut der = LcCBB::new(1024);
+        // Data shows that the SubjectPublicKeyInfo is roughly 356% to 375% increase in size comapred to the RSA key
+        // size in bytes for keys ranging from 2048-bit to 4096-bit. So size the initial capacity to be roughly
+        // 400% as a consernative estimate to avoid needing to reallocate for any key in that range.
+        let key_size_bytes = key_size_bytes(key);
+
+        let mut der = LcCBB::new(key_size_bytes + (key_size_bytes * 4));
```
justsmth, line 145: NP: this is just 5 times `key_size_bytes` -- but maybe you broke it out for clarity regarding the comment above? (Also a couple of spelling mistakes in the comment.)
```
let mut der = LcCBB::new(5 * key_size_bytes);
```

justsmth (approved):

aws-lc-rs/src/rsa/encoding.rs
skmcgrail: Really need to look for a spell check extension for vscode 😬.

Yeah I have it broken out just to make it clear, but I can optimize it to your suggestion and leave a comment.

justsmth (approved):

andrewhop (approved):

PR #309, opened 2024-01-12
A working FIPS build for Windows

justsmth:
### Issues:
N/A

### Description of changes: 
* A working FIPS build for Windows.
* This change uses a Windows batch script to execute `vcvarsall.bat` and then display all of the environment variables that were set.
  * Our build-script parses environment variables from the output of the script and sets those variables to be available for CMake.

### Call-outs:
N/A

### Testing:
* Test succeeded on my Windows host.

### DIFF
--- a/aws-lc-fips-sys/CMakeLists.txt
+++ b/aws-lc-fips-sys/CMakeLists.txt
@@ -18,9 +18,10 @@ if (BUILD_SHARED_LIBS AND FIPS)
     # as cmake crate will postfix the C/CXX flags after our disablement nullifying them.
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-function-sections -fno-data-sections")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-function-sections -fno-data-sections")
-endif ()
+endif()
 
 add_subdirectory(aws-lc aws-lc EXCLUDE_FROM_ALL)
+
 if (BUILD_LIBSSL)
     add_definitions(-DAWS_LC_RUST_INCLUDE_SSL)
 endif()
@@ -38,6 +39,7 @@ set(FINAL_ARTIFACTS_DIRECTORY ${CMAKE_BINARY_DIR}/artifacts)
 # location to find the artifacts cross-platform.
 set_my_target_properties(
         ARCHIVE_OUTPUT_DIRECTORY ${FINAL_ARTIFACTS_DIRECTORY}
+        RUNTIME_OUTPUT_DIRECTORY ${FINAL_ARTIFACTS_DIRECTORY}
         LIBRARY_OUTPUT_DIRECTORY ${FINAL_ARTIFACTS_DIRECTORY})
 
 # Based on https://stackoverflow.com/a/7750816 as some generators, like MSVC, will try to prefix the output directory
@@ -46,12 +48,14 @@ foreach (OUT_NAME ${CMAKE_CONFIGURATION_TYPES})
     string(TOUPPER ${OUT_NAME} OUT_NAME)
     set_my_target_properties(
             ARCHIVE_OUTPUT_DIRECTORY_${OUT_NAME} ${FINAL_ARTIFACTS_DIRECTORY}
+            RUNTIME_OUTPUT_DIRECTORY_${OUT_NAME} ${FINAL_ARTIFACTS_DIRECTORY}
             LIBRARY_OUTPUT_DIRECTORY_${OUT_NAME} ${FINAL_ARTIFACTS_DIRECTORY})
 endforeach ()
 
 if (BORINGSSL_PREFIX)
     if (MSVC)
         set(TARGET_PREFIX "${BORINGSSL_PREFIX}")
+        set_my_target_properties(IMPORT_PREFIX ${TARGET_PREFIX})
     else()
         set(TARGET_PREFIX "lib${BORINGSSL_PREFIX}")
     endif()
--- a/aws-lc-fips-sys/builder/cmake_builder.rs
+++ b/aws-lc-fips-sys/builder/cmake_builder.rs
@@ -3,6 +3,7 @@
 
 use crate::OutputLib::{Crypto, RustWrapper, Ssl};
 use crate::{target, target_arch, target_os, target_vendor, test_command, OutputLibType};
+use std::collections::HashMap;
 use std::env;
 use std::ffi::OsStr;
 use std::path::PathBuf;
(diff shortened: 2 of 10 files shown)


aws-lc-fips-sys/builder/cmake_builder.rs
```diff
@@ -85,9 +90,18 @@ impl CmakeBuilder {
             } else {
                 cmake_cfg.define("CMAKE_BUILD_TYPE", "release");
             }
-        } else if target_os() != "windows" {
+        } else if target_os() == "windows" {
             // The Windows/FIPS build rejects "debug" profile
             // https://github.com/aws/aws-lc/blob/main/CMakeLists.txt#L656
+            cmake_cfg.define("CMAKE_BUILD_TYPE", "relwithdebinfo");
+            cmake_cfg.generator("Ninja");
```
skmcgrail, line 97: Is Ninja explicitly required on Windows? Does the MSVC generator not work?

aws-lc-fips-sys/CMakeLists.txt
```diff
@@ -52,8 +52,15 @@ endforeach ()
 if (BORINGSSL_PREFIX)
     if (MSVC)
         set(TARGET_PREFIX "${BORINGSSL_PREFIX}")
+        set_my_target_properties(IMPORT_PREFIX ${TARGET_PREFIX})
     else()
         set(TARGET_PREFIX "lib${BORINGSSL_PREFIX}")
     endif()
-    set_my_target_properties(PREFIX ${TARGET_PREFIX})
+    set_my_target_properties(PREFIX ${TARGET_PREFIX} )
```
skmcgrail, line 59:
```suggestion
    set_my_target_properties(PREFIX ${TARGET_PREFIX})
```

aws-lc-fips-sys/builder/cmake_builder.rs
justsmth: Unfortunately so. I've not been able to get the MSVC generator working. The binary the MSVC generator produces for the FIPS build doesn't pass the bcm "self test".

aws-lc-fips-sys/CMakeLists.txt
```diff
@@ -52,8 +52,15 @@ endforeach ()
...
 endif ()
+
+if (BUILD_LIBSSL)
+    install(TARGETS rust_wrapper crypto ssl DESTINATION .)
+else()
+    install(TARGETS rust_wrapper crypto DESTINATION .)
+endif()
```
skmcgrail, line 66: Doesn't the build script have some logic to also copy the `include` headers into this path? I guess what does this produce, cause I don't think it inherits the install artifact behavior defined in AWS-LC right?

justsmth: * The DLLs were not being found at runtime unless they were in the `build/artifacts`.
    * This is the we set for: `cargo:rustc-link-search=native=`).
* The DLLS were only being placed `build/artifacts` when I use the "install" target.
    * If I left off the `DESTINATION` parameters, it would place the DLLs under `build/artifacts/bin`, where they weren't being found.

aws-lc-fips-sys/builder/cmake_builder.rs
```diff
@@ -137,11 +156,31 @@ impl CmakeBuilder {
...
+        }
+        Ok(map)
+    }
 }
 
 impl crate::Builder for CmakeBuilder {
     fn check_dependencies(&self) -> Result<(), String> {
         let mut missing_dependency = false;
+        if target_os() == "windows" && !test_ninja_command() {
+            eprintln!("Missing dependency: Ninja is required for FIPS on Windows.");
```
WillChilds-Klein, line 181: is this true for non-FIPS as well? looks like we don't check FIPSness on L180

aws-lc-fips-sys/builder/main.rs
```diff
@@ -116,11 +115,25 @@ fn target_platform_prefix(name: &str) -> String {
...
 
-fn test_command(executable: &OsStr, args: &[&OsStr]) -> bool {
-    if let Ok(output) = Command::new(executable).args(args).output() {
-        return output.status.success();
+pub(crate) struct TestCommandResult {
+    output: Box<str>,
+    status: bool,
+}
+
+fn test_command(executable: &OsStr, args: &[&OsStr]) -> TestCommandResult {
```
WillChilds-Klein, line 123: nit: can we de-duplicate this builder logic between aws-lc-fips-sys and aws-lc-sys?

aws-lc-sys/CMakeLists.txt
```diff
@@ -39,12 +41,14 @@ foreach (OUT_NAME ${CMAKE_CONFIGURATION_TYPES})
...
     set_my_target_properties(
             ARCHIVE_OUTPUT_DIRECTORY_${OUT_NAME} ${FINAL_ARTIFACTS_DIRECTORY}
+            RUNTIME_OUTPUT_DIRECTORY_${OUT_NAME} ${FINAL_ARTIFACTS_DIRECTORY}
             LIBRARY_OUTPUT_DIRECTORY_${OUT_NAME} ${FINAL_ARTIFACTS_DIRECTORY})
 endforeach ()
 
 if (BORINGSSL_PREFIX)
     if (MSVC)
         set(TARGET_PREFIX "${BORINGSSL_PREFIX}")
+        set_my_target_properties(IMPORT_PREFIX ${TARGET_PREFIX})
```
WillChilds-Klein, line 51: why do you have to set this on windows only?

aws-lc-fips-sys/builder/cmake_builder.rs
justsmth: The `aws-lc-fips-sys` crate only gets built for FIPS.

aws-lc-sys/CMakeLists.txt
justsmth: An `IMPORT_PREFIX` is only needed for Windows. Windows typically expects a separate "import library" (*.lib) corresponding to a shared library (*.dll) to be provided for consumers to link against. So, the (*.dll) shared library is not used for linking at compile time, it's only used at runtime.

skmcgrail (approved):

aws-lc-sys/include/rust_wrapper.h
```diff
@@ -20,9 +23,9 @@ extern "C" {
 // The following functions are wrappers over inline functions and macros in
 // BoringSSL, which bindgen cannot currently correctly bind. These wrappers
 // ensure changes to the functions remain in lockstep with the Rust versions.
-int ERR_GET_LIB_RUST(uint32_t packed_error);
-int ERR_GET_REASON_RUST(uint32_t packed_error);
-int ERR_GET_FUNC_RUST(uint32_t packed_error);
+AWS_LC_SYS_EXPORT int ERR_GET_LIB_RUST(uint32_t packed_error);
```
samuel40791765, line 26: Can we use `OPENSSL_EXPORT` from `"openssl/base.h"`? Seems like it would come in handy, so we wouldn't need duplicate definitions in both `rust_wrapper.h`s

justsmth: Is that intended for use outside the library? It might work if we define `BORINGSSL_IMPLEMENTATION` and `BORINGSSL_SHARED_LIBRARY` prior to its include.

samuel40791765: hmm I was under the impression that these were already defined if we built with AWS-LC. 
* https://github.com/aws/aws-lc/blob/eaa19c77d000175da90da7b4ef22ffe24dfc2a2f/INCORPORATING.md?plain=1#L72-L77

Not a blocker though, just curious if it was reusable

samuel40791765 (approved):

PR #310, opened 2024-01-12
Add CI test with musl

justsmth:
### Issues:
N/A

### Description of changes: 
* Add test using musl to our CI.

### Call-outs:
N/A

### Testing:
N/A


skmcgrail (approved):

aws-lc-fips-sys/builder/bindgen.rs
```diff
@@ -90,7 +90,8 @@ const PRELUDE: &str = r"
     clippy::too_many_lines,
     clippy::unreadable_literal,
     clippy::used_underscore_binding,
-    clippy::useless_transmute
+    clippy::useless_transmute,
+    clippy::pub_underscore_fields
```
andrewhop, line 94: Is there no way to share this common config/options for clippy between files? This seems really annoying to update. 

Also looking at this change what is introducing a public field with an underscore?

justsmth: > Also looking at this change what is introducing a public field with an underscore?

Our bindings for AWS-LC are automatically generated from its include files using [rust-bindgen](https://github.com/rust-lang/rust-bindgen). For a variety of reasons, the code produced by "bindgen" violates Rust's conventions for symbol naming and other Rust best-practices. 

In this case, the generated code hasn't change. Instead, [clippy::pub_underscore_fields](https://rust-lang.github.io/rust-clippy/master/index.html#/pub_underscore_fields) is a new lint being added to clippy for the upcoming 1.77.0 Rust release.

justsmth: > Is there no way to share this common config/options for clippy between files? This seems really annoying to update.

This is the first time we've had to modify this `#[allow(...)]` since [Sean integrated the sys-crates into this repo](https://github.com/aws/aws-lc-rs/pull/75) in February of last year.  There's really only two places in each sys-crate (so four places total), [the `PRELUDE`](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-sys/builder/bindgen.rs#L76) and [lib.rs](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-sys/src/lib.rs#L37) -- the other locations are generated code, so they have what's provided by the `PRELUDE`.

skmcgrail (approved):

Issue #297, opened 2023-12-14
Remove CMake requirement for builds


camshaft:
### Problem:

CMake is not a ubiquitous build tool; for example it is not automatically installed with the [build-essential](https://packages.ubuntu.com/focal/build-essential). By requiring CMake, `aws-lc-rs` puts additional build requirements on environments, which will potentially hinder adoption.

### Solution
Instead, the library should have a build system that uses more ubiquitous tools. I think `make` would be a step in the right direction, but I think _ideally_ the only requirement should be a C compiler and assembler. This is the approach that was taken in [s2n-tls-sys](https://github.com/aws/s2n-tls/blob/main/bindings/rust/s2n-tls-sys/build.rs) which makes it quite easy to build.

justsmth:
Related to #152

justsmth:
As of release [v1.7.0](https://github.com/aws/aws-lc-rs/releases/tag/v1.7.0), CMake is no longer required to build **aws-lc-rs** for Linux/MacOS on x86-64/aarch64. For these platforms, only the minimal Rust & C toolchains should be required for the build to succeed.

I'm going to close this issue now. Feel free to open another issue to request extending support for our no-cmake build to other platforms (e.g., Windows).

Issue #152, opened 2023-06-13
Support for building aws-lc-rs with just rustc 


rcoh:
### Problem:
We'd love to include aws-lc-rs in the Rust SDK but feel that the current build requirements are too onerous for most customers to use it as the default.

Consider using `build.rs` to replace `CMake` to compile `aws-lc-rs`. Perhaps we could also use the Rust compiler to achieve the static linking achieved with the Go compiler currently?

skmcgrail:
This is potentially doable for the (non-fips) build of aws-lc, as that build doesn't require the use of the Go compiler.

The challenge here would be specific to the static FIPS build, which is what requires the Go compiler. In particular the Go compiler is used to remove relocation entries from the FIPS module boundary of the resulting assembly code. This is done specifically by [delocate](https://github.com/aws/aws-lc/blob/main/util/fipstools/delocate/delocate.go), and more information can be found [here](https://github.com/aws/aws-lc/tree/main/util/fipstools/delocate).

pronebird:
I concur. I've inherited aws-lc-rs through rustls, now I have to walk around 3 platforms we have and tell everyone how to install CMake which is sub-optimal.

justsmth:
> I concur. I've inherited aws-lc-rs through rustls, now I have to walk around 3 platforms we have and tell everyone how to install CMake which is sub-optimal and would rather avoid.

My highest priority right now is eliminating the need for CMake when building for our primary platforms (Linux/MacOS on x86-64/aarch64). Progress is being made. You can watch [this PR](https://github.com/aws/aws-lc-rs/pull/317) for the latest updates.

justsmth:
As of release [v1.7.0](https://github.com/aws/aws-lc-rs/releases/tag/v1.7.0), CMake is no longer required to build **aws-lc-rs** for Linux/MacOS on x86-64/aarch64. For these platforms, only the minimal Rust & C toolchains should be required for the build to succeed.

I'm going to close this issue now. Feel free to open another issue to request extending support for our no-cmake build to other platforms (e.g., Windows).

PR #317, opened 2024-01-24
No CMake required for non-FIPS

justsmth:
### Issues:
Addresses Issues:
* #152 
* #297 

### Description of changes: 
* Allow for non-FIPS build requiring only a C compiler.
* Extended existing bindgen automation to also collect require information for no-CMake build.
* Cleaned up build logic around platform detection.
* Re-aligned aws-lc-fips-sys builder/main.rs with aws-lc-sys.

### Call-outs:
* The change allows the build to be performed w/o CMake on the same platforms for which we pre-generate the bindings: `i686-unknown-linux-gnu`, `x86_64-unknown-linux-gnu`, `aarch64-unknown-linux-gnu`, `x86_64-unknown-linux-musl`, `aarch64-unknown-linux-musl`, `x86_64-apple-darwin`, `aarch64-apple-darwin`
* The `AWS_LC_SYS_CMAKE_BUILDER` environment variable can be used to force the build to use, or not use, CMake.


aws-lc-sys/builder/cc/aarch64-apple-darwin.toml
```diff
@@ -0,0 +1,314 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
camshaft, line 1: Not really sure this is worth the toml/serde dependency... Why not just generate a rust file and `include!` it? You can even have it create a `struct` value as well:

```rust
static CRYPTO_LIB: Library = Library {
    name: "crypto",
    flags: &[],
    sources: &[...],
};
```

If you want better scoping your could also use the `#[path]` attribute on a module:

```rust
#[path = "./aarch64-apple-darwin.rs"]
mod libraries;
```

camshaft:
One thing to note that we need to replicate (and maintain) the build-time checks here:

https://github.com/aws/aws-lc/blob/37708f52446ee727ebeea6beaba75baf0588593b/CMakeLists.txt#L426-L430
https://github.com/aws/aws-lc/blob/37708f52446ee727ebeea6beaba75baf0588593b/CMakeLists.txt#L311-L317

aws-lc-sys/builder/cc/aarch64-apple-darwin.toml
justsmth: One reason to not generate this information in Rust syntax is that AWS-LC source file information could be useful beyond just building our (Rust-specific) sys-crates. For example, we could potentially generate TOML files like these as part of an AWS-LC release for use by other build processes or static/dynamic-analysis.

That said, updating the shell script to generate Rust-syntax should be fairly easy, and elimination of the toml/serde dependencies could reduce "maintenance costs" long term. I'll look into it.

camshaft: > For example, we could potentially generate TOML files like these as part of an AWS-LC release for use by other build processes or static/dynamic-analysis.

That should still be possible with the rust files, too. Even if you just parse it with `syn` in your analysis script (and maybe that could produce the toml files). I just would prefer reducing the number of dependencies here, since that usually causes the most problems for maintenance.

aws-lc-fips-sys/builder/main.rs
```diff
@@ -130,39 +130,42 @@ fn prefix_string() -> String {
...
+    stdout: Box<str>,
     executed: bool,
+    status: bool,
 }
 
 fn test_command(executable: &OsStr, args: &[&OsStr]) -> TestCommandResult {
-    if let Ok(result) = Command::new(executable).args(args).output() {
-        let output = String::from_utf8(result.stdout)
+    if let Ok(mut result) = Command::new(executable).args(args).output() {
+        result.stderr.truncate(4112);
```
camshaft, line 149: why 4112?

scripts/build/collect_build_src.sh
```diff
@@ -0,0 +1,227 @@
...
+    for FILE in "${@}"; do
+        FILE_PATH="${SYS_CRATE_DIR}/aws-lc/${FILE}"
+        if [[ ! -f "${FILE_PATH}" ]]; then
+            echo File does not exist: "${FILE_PATH}"
+            exit 1
+        fi
+    done
+}
+
+function generate_toml() {
```
camshaft, line 152: it's no longer toml :)

aws-lc-sys/builder/cc_builder/aarch64_apple_darwin.rs
```diff
@@ -0,0 +1,317 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+// Wed Apr  3 12:53:57 UTC 2024
+
+use crate::cc_builder::Library;
+
+pub(super) const CRYPTO_LIBRARY: Library = Library {
+    name: "crypto",
+    flags: &[],
```
camshaft, line 9: Are these supposed to have values? or is it for future use?

aws-lc-fips-sys/builder/main.rs
justsmth: No reason in particular. It was 2112, then thought that might not be enough. I will make it a named constant to remove the "magic" from it.

aws-lc-sys/builder/cc_builder/aarch64_apple_darwin.rs
justsmth: I thought I might end up needing to specify flags that could vary by platform, but that's not been needed yet. I can remove the attribute if needed.

camshaft: I think it's fine to keep. I just wasn't sure if there were values supposed to be here on at least one platform. Maybe a comment for now?

camshaft (dismissed):

aws-lc-sys/builder/cc_builder.rs
samuel40791765, line 1: Probably worth mentioning in this file that this should be updated and maintained when any relevant build options in AWS-LC's CMake file are changed.

aws-lc-sys/builder/cc_builder.rs
```diff
@@ -0,0 +1,284 @@
...
+        if OutputLibType::Dynamic == self.output_lib_type {
+            // https://github.com/rust-lang/cc-rs/issues/594
+            return Err("CcBuilder only supports static builds".to_string());
+        }
+
+        if PlatformConfig::default_for(&target()).is_none() {
+            return Err(format!("Platform not supported: {}", target()));
+        }
+
+        if Some(true) == env_var_to_bool("CARGO_FEATURE_SSL") {
```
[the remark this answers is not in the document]
justsmth: Th `CARGO_FEATURE_SSL` environment variable is set by cargo when our "ssl" feature is enabled. It's documented here under "[Environment variables Cargo sets for build scripts](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts)".

Cross.toml
```diff
@@ -4,5 +4,5 @@ dockerfile = "./docker/linux-cross/Dockerfile"
 [build.env]
 passthrough = [
     "AWS_LC_SYS_INTERNAL_NO_PREFIX", "AWS_LC_RUST_INTERNAL_BINDGEN", "AWS_LC_FIPS_SYS_INTERNAL_NO_PREFIX", "GOPROXY",
-    "AWS_LC_FIPS_SYS_STATIC", "AWS_LC_SYS_STATIC"
+    "AWS_LC_FIPS_SYS_STATIC", "AWS_LC_SYS_STATIC", "AWS_LC_SYS_CMAKE_BUILDER", "AWS_LC_SYS_CC_SRC_COLLECTOR"
```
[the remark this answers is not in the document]
justsmth: I originally had expected to need to pass additional build options to CMake in order to have the build generate sufficient debug information from which we could identify all of the source files used.  However, it's turned out to not be needed so far.  I'd like to leave it listed here in case its needed when we expand the set of platforms supported by this mechanism.

aws-lc-sys/builder/main.rs
```diff
@@ -126,39 +128,42 @@ fn prefix_string() -> String {
...
     #[allow(dead_code)]
-    output: Box<str>,
+    stderr: Box<str>,
+    #[allow(dead_code)]
+    stdout: Box<str>,
+    executed: bool,
     status: bool,
 }
 
+const MAX_CMD_OUTPUT_SIZE: usize = 1 << 12;
```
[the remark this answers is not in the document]
justsmth: The intention is to avoid collecting excessive output from a build tool (i.e., I don't want to dump megabytes of text onto the console). I figure 4k of output is sufficient to diagnose most failures for the tools that we're using in our build. So far it's been enough for my development.

scripts/build/collect_build_src.sh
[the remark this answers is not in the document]
justsmth: Added comments/documentation in latest commit.

aws-lc-sys/builder/cc_builder.rs
justsmth: Added comments/documentation in latest commit.

aws-lc-sys/builder/cc_builder.rs
```diff
@@ -0,0 +1,297 @@
...
+        if let Some(prefix) = &self.build_prefix {
+            cc_build.compile(format!("{}_crypto", prefix.as_str()).as_str());
+        } else {
+            cc_build.compile(lib.name);
+        }
+    }
+
+    // This performs basic checks of compiler capabilities and sets an appropriate flag on success.
+    // This should be kept in alignment with the checks performed by AWS-LC's CMake build.
+    // See: https://github.com/search?q=repo%3Aaws%2Faws-lc%20check_compiler&type=code
```
camshaft, line 177: Not required for this PR but something to think about long term: what mechanism do we need to ensure these compiler checks stay in sync with the AWS-LC CMake build? In s2n-tls we went with the approach of scanning the features directory in each build system, which automatically picks up new features (or removes unused ones). Maybe we should do something similar here?

https://github.com/aws/s2n-tls/pull/3921

aws-lc-sys/builder/cc_builder.rs
```diff
@@ -0,0 +1,297 @@
...
+            output_lib_type,
+        }
+    }
+
+    fn create_builder(&self) -> cc::Build {
+        let mut cc_build = cc::Build::default();
+        cc_build
+            .out_dir(&self.out_dir)
+            .flag("-std=c99")
+            .flag("-Wno-unused-parameter")
```
camshaft, line 104: why just this flag and not any of the others? do we want to keep all of these in sync with CMake? do we want to emit warnings like these in the rust build?

camshaft (approved):

pronebird:
When you say FIPS, what do you refer to? Federal Information Processing Standards? Apologize for being silly, not familiar with the abbreviation.

justsmth:
Yeah, "FIPS" stands for "Federal Information Processing Standards". We perform a special build process for AWS-LC when the "fips" feature is enabled. There is more information and links in this blog post about it: https://aws.amazon.com/blogs/security/aws-lc-is-now-fips-140-3-certified/

aws-lc-sys/Cargo.toml
```diff
@@ -54,6 +54,7 @@ bindgen = ["dep:bindgen"] # Generate the bindings on the targetted platform as a
 cmake = "0.1.48"
```
pronebird, line 54: Does cmake have to be a part of build-dependencies?

PR #327, opened 2024-01-29
Improve how vcvarsall is located

justsmth:
### Issues:
N/A

### Description of changes: 
* For the Windows/FIPS build, add fallback logic for locating `vcvarsall.bat` when `vswhere.exe` fails.
* Increase verbosity of failures.

### Call-outs:
N/A

### Testing:
* I'm still not sure of the root cause for the [build failure seen here](https://github.com/rustls/rustls/actions/runs/7694846379/job/20966481181?pr=1732#step:7:205).  However, this change should provide more logging to diagnose the issue.

### DIFF
--- a/aws-lc-fips-sys/builder/printenv.bat
+++ b/aws-lc-fips-sys/builder/printenv.bat
@@ -2,6 +2,7 @@
 setlocal EnableDelayedExpansion
 for /f "usebackq tokens=* delims=" %%i in (`"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath`) do (
     set "VS_PATH=%%i\VC\Auxiliary\Build\"
+    echo FOUND: "!VS_PATH!"
     goto FoundVS
 )
 
--- a/aws-lc-fips-sys/builder/printenv.bat
+++ b/aws-lc-fips-sys/builder/printenv.bat
@@ -1,13 +1,22 @@
 @echo off
 setlocal EnableDelayedExpansion
+
+set "TOP_DIR=%ProgramFiles(x86)%\Microsoft Visual Studio"
 for /f "usebackq tokens=* delims=" %%i in (`"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath`) do (
-    set "VS_PATH=%%i\VC\Auxiliary\Build\"
-    echo FOUND: "!VS_PATH!"
-    goto FoundVS
+    set "TOP_DIR=%%i"
+    echo VS Installation: "!TOP_DIR!"
+    if exist "!TOP_DIR!\VC\Auxiliary\Build\vcvarsall.bat" (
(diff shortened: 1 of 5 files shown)


skmcgrail (dismissed):

aws-lc-fips-sys/builder/printenv.bat
```diff
@@ -1,7 +1,35 @@
 @echo off
+setlocal EnableDelayedExpansion
 for /f "usebackq tokens=* delims=" %%i in (`"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath`) do (
-    set "VS_PATH=%%i"
+    set "VS_PATH=%%i\VC\Auxiliary\Build\"
```
WillChilds-Klein, line 4: we can assume that this directory exists?

aws-lc-fips-sys/builder/printenv.bat
```diff
@@ -1,7 +1,35 @@
...
+    echo FOUND: "!VS_PATH!"
+    goto FoundVS
 )
-call "%VS_PATH%\VC\Auxiliary\Build\vcvarsall.bat" x64
 
-set
\ No newline at end of file
+echo Visual Studio installation not found using vswhere. Searching in default directories...
+for /R "%ProgramFiles(x86)%\Microsoft Visual Studio" %%a in (vcvarsall.bat) do (
+    if exist "%%~fa" (
```
WillChilds-Klein, line 11: what does `~f` do in this expansion?

aws-lc-fips-sys/builder/printenv.bat
```diff
@@ -1,7 +1,35 @@
...
+        set "VS_PATH=%%~dpa"
```
WillChilds-Klein, line 12: should we set/override the VS_PATH to this or append to it?

aws-lc-fips-sys/builder/printenv.bat
justsmth: This is the standard location of the script for [VS 2017 and VS 2019](https://learn.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-170#vcvarsall) (and VS 2022 apparently) but it doesn't appear to be guaranteed:
> ... Then, use CD again to change to the subdirectory that contains the configuration-specific command files. For Visual Studio 2019 and Visual Studio 2017, use the VC\Auxiliary\Build subdirectory. For Visual Studio 2015, use the VC subdirectory.

aws-lc-fips-sys/builder/printenv.bat
justsmth: It expands the variable to a fully-qualified path name. I found documentation of this and other modifiers (like the `dp` used on the next line) [here](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/for).

skmcgrail (approved):

WillChilds-Klein (approved):

PR #330, opened 2024-01-30
Upgrade the version of the `untrusted` crate

joshtriplett:
### Description of changes: 
This upgrades the version of the `untrusted` crate to the latest 0.9, to avoid
duplicate dependencies when building aws-lc-rs in a project with other
dependencies on `untrusted`.

### Call-outs:
A few deprecated functions take `untrusted::Input` as a parameter directly. This PR should wait for a new major version of aws-lc-rs that drops those deprecated functions. (Alternatively, it'd be possible to have an `untrusted07` feature flag that continues depending on `untrusted` 0.7 for the sole purpose of the API of those deprecated funtions, but that doesn't seem worthwhile here.)

As far as I can tell, no non-deprecated APIs would be affected, which means future upgrades to untrusted would not have this issue.

### Testing:
Builds and passes `cargo test`, with and without the `ring-io` and `ring-sig-verify` feature flags set.

### DIFF
--- a/aws-lc-rs-testing/Cargo.toml
+++ b/aws-lc-rs-testing/Cargo.toml
@@ -15,7 +15,7 @@ asan = ["aws-lc-rs/asan"]
 
 [dependencies]
 aws-lc-rs = { version = "1.0", path = "../aws-lc-rs", features = ["ring-sig-verify", "unstable"] }
-untrusted = { version = "0.7" }
+untrusted = { version = "0.9" }
 
 [dev-dependencies]
 paste = "1.0.13"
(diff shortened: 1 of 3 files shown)


justsmth:
As a possible workaround, you could disable default features for `aws-lc-rs` and only add back the `aws-lc-sys` feature. (The `untrusted` crate is only needed when the `ring-io` feature is enabled.)

joshtriplett:
@justsmth Unfortunately, rustls is one of the crates that depends on it with default features enabled.

joshtriplett:
I've submitted https://github.com/rustls/rustls/pull/1768 to attempt to fix that.

justsmth:
[Downstream PR](https://github.com/rustls/rustls/pull/1768) merged. Closing this PR.

PR #335, opened 2024-01-31
feat: export sys variables from aws-lc-rs crate

camshaft:
### Description of changes: 

This change exports the underlying `*-sys` crate in the `aws-lc-rs` crate. This is needed for a couple of reasons:

* In s2n-tls, we currently depend on the `aws-lc-sys = "0.12"`. In s2n-quic, we depend on `aws-lc-rs = ">=1.5"`. When a new `aws-lc-rs` version is published, then we get conflicting resolutions and end up with 2 copies of `aws-lc-sys`. Symbol prefixing is currently not working well enough to prevent conflicts.
* If crates depend on `aws-lc-sys` directly, they'll have to duplicate the `fips`/`non-fips` feature switching logic that's already present in `aws-lc-rs`. This logic is _a bit_ complicated so it'd be nice to have it implemented in a single location.

With this change, s2n-tls will instead depend on `aws-lc-rs = "1"`. This will avoid issues of preventing it using the latest aws-lc-sys crates on newer releases. It'll also make it easier to pick up `fips` builds, since we rely on the logic in `aws-lc-rs`.

### Call-outs:
I did some minor refactoring to the build scripts to try and make it a bit clearer.

### Testing:
I've expanded the `sys-testing` crate to include the new functionality.


aws-lc-rs/build.rs
```diff
@@ -2,28 +2,74 @@
...
-        .filter(|x| **x)
-        .count();
-
-    if at_least_one_count < 1 {
-        eprint!(
-            "one of the following features must be specified: \
-        aws-lc-sys, non-fips, or fips."
+    // if `fips` is enabled, then use that
+    let sys_crate = if cfg!(feature = "fips") {
+        "aws-lc-fips-sys"
```
justsmth, line 15: I think it could just check for the "aws-lc-fips-sys" feature here, since ["fips" depends on "aws-lc-fips-sys"](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-rs/Cargo.toml#L43)

aws-lc-rs/build.rs
```diff
@@ -2,28 +2,74 @@
...
+    } else if cfg!(feature = "aws-lc-sys") {
+        "aws-lc-sys"
+    } else {
+        panic!(
```
justsmth, line 19: One problem consumers have is that our build requires the consumer to specify which sys crate to use. Instead, I think it would be better [automatically select](
https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-cfg) the "aws-lc-sys" when neither was explicitly selected.
```
   let sys_crate = if cfg!(feature = "aws-lc-fips-sys") {
        "aws-lc-fips-sys"
    } else if cfg!(feature = "aws-lc-sys") {
        "aws-lc-sys"
    } else {
        println!("cargo:rustc-cfg=feature=\"aws-lc-sys\"]");
        "aws-lc-sys"
    };
```

aws-lc-rs/build.rs
```diff
@@ -2,28 +2,74 @@
...
+        "DEP_AWS_LC_FIPS_"
+    } else {
+        "DEP_AWS_LC_"
+    };
+
+    let mut selected = String::default();
+    let mut candidates = vec![];
+
+    // search through the DEP vars and find the selected sys crate version
+    for (name, value) in std::env::vars() {
```
justsmth, line 38: I know it's less efficient, but I think it would be easier to read if it would first identify the selected version, then try to export the variables.
```
    let mut selected = String::default();
    // search through the DEP vars and find the selected sys crate version
    for (name, _) in std::env::vars() {
        // we're still looking for a selected prefix
        if let Some(version) = name.strip_prefix(prefix) {
            if let Some(version) = version.strip_suffix("_INCLUDE") {
                // we've found the selected version so update it
                selected = format!("{prefix}{version}_");
                break;
            }
        }
    }
    assert!(!selected.is_empty(), "missing {prefix} include");

    // process all environment vars
    for (name, value) in std::env::vars() {
        if let Some(var) = name.strip_prefix(&selected) {
            eprintln!("cargo:rerun-if-env-changed={name}");
            let var = var.to_lowercase();
            println!("cargo:{var}={value}");
        }
    }
```

sys-testing/Cargo.toml
```diff
@@ -6,10 +6,13 @@ publish = false
 
 [features]
 default = ["aws-lc-sys"]
```
justsmth, line 8: Instead of setting a default feature here, could we just fallback to "aws-lc-sys" in the build script?

aws-lc-rs/build.rs
camshaft: Hmm I'm not sure that'll actually work:

> Note that this does not affect Cargo’s dependency resolution. This cannot be used to enable an optional dependency, or enable other Cargo features.

aws-lc-rs/build.rs
camshaft: works for me

aws-lc-rs/build.rs
camshaft: that's actually _literally_ what i had initially. but i do worry that we're iterating over this twice, which may be quite large in some environments and make build times quite a bit worse. what i ended up with _should_ do it in a single pass (assuming it's alphabetical), which seems not nearly as bad.

sys-testing/Cargo.toml
camshaft: As mentioned above, I don't think we can use that to pull in optional dependencies.

justsmth (approved):

aws-lc-rs/Cargo.toml
```diff
@@ -2,6 +2,8 @@
 name = "aws-lc-rs"
 authors = ["AWS-LibCrypto"]
 version = "1.6.1"
+# this crate re-exports whatever sys crate that was selected
+links = "aws_lc_rs_1_6_1"
```
skmcgrail, line 6: Let's make this `aws_libcrypto_1_6_1`.  I don't want to confuse `aws_lc_rs` being a "native library".
"

sys-testing/build.rs
```diff
@@ -4,31 +4,46 @@
...
+    select_dep!("aws-lc-fips-sys");
+
+    assert_eq!(
+        deps.len(),
+        1,
+        "exactly one dependency is allowed at a time, got {deps:?}"
+    );
+
+    let dep = deps.pop().unwrap();
+    let aws_lc_sys_links = get_package_links_property(&format!("../{dep}/Cargo.toml"));
```
skmcgrail, line 28:
```suggestion
    let dep_links = get_package_links_property(&format!("../{dep}/Cargo.toml"));
```

sys-testing/Cargo.toml
```diff
@@ -6,10 +6,13 @@ publish = false
```
skmcgrail, line 6: Should we rename this crate to `links-testing`?. Not really specific to `sys` anymore.

aws-lc-rs/Cargo.toml
camshaft: Yeah the only way to export build variables from this crate is by having a `links` attribute. I'm fine to call it whatever, though.

justsmth: How about `aws_lc_rs_1_6_1_libcrypto`?

camshaft: talked offline with @justsmth. we think `aws_lc_rs_1_6_1_sys` communicates best what's happening here.

justsmth (approved):

skmcgrail (approved):

PR #351, opened 2024-02-16
Add check for sys crate publish; Script cleanup

justsmth:
### Description of changes: 
* Add check to sys-crate publishing script to verify expected prefix.
* Fix how generation scripts check for macOS.
* Remove generation script expectation that we're on main.
* Other cleanup in generation scripts.

### Testing:
Tested on macos and linux hosts

### DIFF
--- a/scripts/generate/_test_supported_builds.sh
+++ b/scripts/generate/_test_supported_builds.sh
@@ -21,24 +21,6 @@ IGNORE_MACOS=0
 RELATIVE_CRATE_PATH=""
 CRATE_TEST_SCRIPT="${SCRIPT_DIR}/_crate_test_build.sh"
 
-while getopts c:f option; do
-  case $option in
-  c)
-    RELATIVE_CRATE_PATH="${OPTARG}"
-    ;;
-  f)
-    GENERATE_FIPS=1
-    ;;
-  m)
-    IGNORE_MACOS=1
-    ;;
-  ?)
-    usage
-    exit 1
-    ;;
-  esac
-done
-
 if [[ -z "${RELATIVE_CRATE_PATH}" ]]; then
   echo "Relative crate path must be provided"
   exit 1
@@ -51,8 +33,8 @@ pushd "${REPO_ROOT}" &>/dev/null
 pids=''
 if [[ "${GENERATE_FIPS}" -eq 0 ]]; then
   ### Test crate on Mac
-  IS_MACOS_HOST=$(check_running_on_macos [[ $IGNORE_MACOS -eq 0 ]])
-  if [[ $IS_MACOS_HOST -eq 0 ]]; then
+  IS_MACOS_HOST=$(check_running_on_macos ${IGNORE_MACOS})
+  if [[ $IS_MACOS_HOST -eq 1 ]]; then
     ${CRATE_TEST_SCRIPT} -c "${RELATIVE_CRATE_PATH}" &
   else
     echo Script is not running on macOS.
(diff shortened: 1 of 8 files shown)


scripts/generate/_test_supported_builds.sh
```diff
@@ -21,24 +21,6 @@ IGNORE_MACOS=0
...
-    ;;
-  m)
-    IGNORE_MACOS=1
-    ;;
-  ?)
-    usage
-    exit 1
-    ;;
-  esac
-done
```
skmcgrail, line 40: I'd like to keep this concept, the reason I wrote it like this was to make it easier to allow us to test individual parts of the process. Otherwise there is a some top-level script that sets a bunch of environment variables, and then it is unclear what the individual scripts actually need / use. Thus if you just want to invoke this script to test changes, you need to know what combination of environment variables need to be set. In the previous design a script could be explicit about the options it expects to be passed in.

scripts/publish/_publish_tools.sh
```diff
@@ -24,6 +24,11 @@ function publish_options {
   done
 }
 
+# Finds the version of the crate based on current working directory
+function crate_version_prefix {
+  grep -E '^version =' ./Cargo.toml | sed -e 's/.*"\([0-9]*\)\.\([0-9]*\)\.\([0-9]*\)"/\1_\2_\3/'
```
skmcgrail, line 29: `${REPO_ROOT}"/scripts/tools/cargo-dig.rs -v` will do this without the need for the regex. Makes it consistent with how we get the version in other spots of the automation.

smittals2 (dismissed):

scripts/publish/publish-aws-lc-sys.sh
```diff
@@ -7,14 +7,26 @@ set -e
...
+CRATE_PREFIX="aws_lc_${CRATE_VERSION_PREFIX}"
+EXPECTED_MACRO_LINE="#define BORINGSSL_PREFIX ${CRATE_PREFIX}"
+PREFIX_INCLUDE_PATH="${CRATE_DIR}"/generated-include/openssl/boringssl_prefix_symbols_asm.h
+
+if ! grep "${EXPECTED_MACRO_LINE}" "${PREFIX_INCLUDE_PATH}"; then
+  echo
+  echo ERROR: Expected prefix macro not found in: "${PREFIX_INCLUDE_PATH}"
+  echo "${EXPECTED_MACRO_LINE}"
+  exit 1
+fi
```
samuel40791765, line 28: This might be worth abstracting to `scripts/publish/_publish_tools.sh`

scripts/publish/_publish_tools.sh
```diff
@@ -24,6 +24,41 @@ function publish_options {
...
 }
 
+# Finds the version of the crate based on current working directory
+function crate_version_prefix {
+  local REPO_ROOT
+  REPO_ROOT=$(git rev-parse --show-toplevel)
+  "${REPO_ROOT}"/scripts/tools/cargo-dig.rs -v | sed -e 's/\([0-9]*\)\.\([0-9]*\)\.\([0-9]*\)/\1_\2_\3/'
+}
+
+function sanity_check_sys_crate {
```
skmcgrail, line 34: Might be worth validating that this information is correct too. Feel free to either include it in this PR or as a follow-up.

```
[package.metadata.aws-lc-sys]
commit-hash = "c2b14dfb82e1a4d0f2260e05cf07c3a864e52d99"
```

skmcgrail (dismissed):

samuel40791765 (dismissed):

skmcgrail (approved):

samuel40791765 (approved):

PR #369, opened 2024-03-08
Auto-generation of aws-lc-sys v0.14.0

justsmth:
### Description of changes: 
* Update aws-lc-sys to 0.14.0.
* Align aws-lc-sys with [AWS-LC v1.22.0](https://github.com/aws/aws-lc/releases/tag/v1.22.0)
  * It was previously aligned to [AWS-LC v1.20.0](https://github.com/aws/aws-lc/releases/tag/v1.20.0)
* Update aws-lc-rs to 1.6.3.
* Pre-generated bindings for 3 more platforms: `aarch64-apple-darwin`, `aarch64-unknown-linux-musl` and `x86_64-unknown-linux-musl`
* Github Workflow can be seen here: https://github.com/aws/aws-lc-rs/actions/runs/8206912705

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

PR #372, opened 2024-03-19
Auto-generated bindings for aws-lc-fips-sys v0.12.6

justsmth:
### Description of changes: 
* Update `aws-lc-fips-sys` to 0.12.6 to align with [AWS-LC-FIPS 2.0.9](https://github.com/aws/aws-lc/releases/tag/AWS-LC-FIPS-2.0.9)
* Auto-generation of bindings for `aws-lc-fips-sys`.
* Provide pregenerated bindings for 4 more platforms: "x86_64-apple-darwin", "aarch64-apple-darwin", "aarch64-unknown-linux-musl" and "x86_64-unknown-linux-musl"
* Added a CI job to build/test FIPS build with LTO enabled.

### Testing:
Bindings generated by this workflow run: https://github.com/aws/aws-lc-rs/actions/runs/8346539433


camshaft (dismissed):

Mark-Simulacrum:
> Defaults to dynamic build for FIPS (except for "musl" or when AWS_LC_FIPS_SYS_STATIC is set in environment).

Is there a reason for this documented somewhere? In general static builds are fairly idiomatic for Rust so this sounds surprising to me.

Instead of conditioning on musl I'd probably recommend crt-static target feature (see https://rust-lang.github.io/rfcs/1721-crt-static.html for some docs), that is probably more in line with matching libc linkage across lots of platforms. (Note that there's a long-standing ask from musl upstream that Rust stops statically linking by default, since that's not what you want on musl distros like Alpine).

justsmth:
> 
> Is there a reason for this documented somewhere? In general static builds are fairly idiomatic for Rust so this sounds surprising to me.

There are a few reasons for this that are fairly specific to the need for FIPS compliance:
* The FIPS standard requires a "self-test" at runtime to verify a hash of the contents of the FIPS module. The [concern with a static build](https://github.com/aws/aws-lc-rs/pull/369#discussion_r1526249484) is that this hash might get broken by link-time-optimizations (LTO).
* A static build of the FIPS module is only supported on Linux for `x86-64` and `aarch64`. The reason for this is [complicated](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/FIPS.md#static-build), but it relates to our need to eliminate relocations (due to having position-independent code) in order for the FIPS module contents hash to not get changed. We call the process of eliminating relocations the "delocation", and it requires parsing and restructuring of assembly code.  This "delocation" logic is [currently implemented in Go](https://github.com/aws/aws-lc/blob/main/util/fipstools/delocate/delocate.go), which is why our "FIPS" build requires Go.
* Thus, we support the shared build of the FIPS modules across many more platforms.

justsmth:
To avoid potentially breaking downstream builds/deployment, the default build of `aws-lc-fips-sys` on Linux (x86-64 and aarch64) remains static in this PR.

scripts/build/collect_symbols.sh
```diff
@@ -26,43 +36,69 @@ done
...
+  grep -v -E '^_Z' | grep -v 'BORINGSSL_bcm_' | grep -v 'BORINGSSL_integrity_test'
+}
+
+function filter_macho_symbols() {
+  grep -E '^_' | sed -e 's/^_\(.*\)/\1/'
 }
 
 function find_libcrypto() {
-  find "${REPO_ROOT}/target" -type f -name "libcrypto.a" | grep "aws-lc-sys"
+  find "${REPO_ROOT}/target" -type f \( -name "lib*crypto.a" -o -name "lib*crypto.so" -o -name "lib*crypto.dylib" \) | grep "${CRATE_NAME}"
```
skmcgrail, line 72: Actually now that I think about it this technically doesn't filter it down to one directory with the grep. As the target directories will have the crate name with a suffix of some sort of hash.

justsmth: Correct.  This script would likely only succeed after a "clean" build was performed.

skmcgrail (approved):

camshaft (approved):

PR #377, opened 2024-04-01
ENV vars for OPENSSL_NO_ASM

justsmth:
### Issues:
* Provides workaround for #376 (or for other platforms where AWS-LC assembly fails to build)

### Description of changes: 
* AWS-LC contains optimized assembly language implementations for many primitive cryptographic operations. As a fall-back, these operations also have a C-implementation provided.
* For most platforms, CMake correctly detects whether/which assembly files should be built. However, CMake's detection is sometimes inadequate or the build environment lacks an assembler (such as NASM), and the `OPENSSL_NO_ASM` flag must be passed to avoid build failures.
* The `OPENSSL_NO_ASM` flag may also be useful for debugging failing cryptographic operations.

### Call-outs:
* The C-implementations are significantly slower than their assembly-language counterparts.
* The `OPENSSL_NO_ASM` flag is only allowed for un-optimized/debug builds.

### Testing:
* Updated CI.

### DIFF
--- a/aws-lc-sys/builder/cmake_builder.rs
+++ b/aws-lc-sys/builder/cmake_builder.rs
@@ -3,7 +3,8 @@
 
 use crate::OutputLib::{Crypto, RustWrapper, Ssl};
 use crate::{
-    execute_command, target, target_arch, target_env, target_os, target_vendor, OutputLibType,
+    cargo_env, execute_command, is_no_asm, target, target_arch, target_env, target_os,
+    target_vendor, OutputLibType,
 };
 use std::env;
 use std::ffi::OsStr;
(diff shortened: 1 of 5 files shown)


aws-lc-sys/builder/cmake_builder.rs
```diff
@@ -101,6 +104,15 @@ impl CmakeBuilder {
         cmake_cfg.define("DISABLE_PERL", "ON");
         cmake_cfg.define("DISABLE_GO", "ON");
 
+        if Some(true) == env_var_to_bool("AWS_LC_SYS_NO_ASM") {
+            let opt_level = cargo_env("OPT_LEVEL");
+            if opt_level == "0" {
+                cmake_cfg.define("OPENSSL_NO_ASM", "1");
+            } else {
+                panic!("AWS_LC_SYS_NO_ASM only allowed for debug builds!")
```
skmcgrail, line 112: What are we accomplishing then? It seems disingenuous to artificially increase platform support with this option, only allow it for debug builds?

justsmth: I see the goals that this achieves as being: (1) Facilitating improved debugging/tracing of cryptographic operations at lower levels, and (2) Unblocking use of our library in development environments on platforms we don't (or can't) fully support currently (such as Windows/ARM64).

dkostic (approved):

skmcgrail (approved):

Issue #396, opened 2024-04-17
Support streaming API for CBC


fgaud:
### Problem:

I would like to be able to use CBC in a streaming fashion. Unfortunately, currently the CBC encryptor always puts padding after an encryption operation and expects padding during a decryption operation.

OpenSSL solves that problem with finalize methods (for ex. https://docs.rs/openssl/latest/openssl/envelope/struct.Open.html#method.finalize)

### Solution:

There are two possible options

* Add CBC support to [EncryptingKey](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/cipher/struct.EncryptingKey.html) and [DecryptingKey](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/cipher/struct.DecryptingKey.html) types, and have the caveat that the inputs must be divisible by the block size, and leave padding decisions to the caller.

* Introduce a StreamingPaddedBlockEncryptingKey and StreamingPaddedBlockDecryptingKey that implements the streaming and padding concepts.

The later would be nicer but more work

* **Does this change any public APIs?** No it add more APIs
* **Which algorithm(s) will this impact?** CBC

### Requirements / Acceptance Criteria:

What must a solution address in order to solve the problem? How do we know the solution is complete?

Ideally we can encrypt/decrypt in a streaming fashion.

* **RFC links:** N/A
* **Related Issues:** None
* **Will the Usage Guide or other documentation need to be updated?** Maybe depending on the solution
* **Testing:** How will this change be tested? Call out new integration tests, functional tests, or particularly
  interesting/important unit tests. Probably through unit tests

### Out of scope:

Is there anything the solution will intentionally NOT address?

justsmth:
FYI -- this API is now available in [v1.8.0](https://github.com/aws/aws-lc-rs/releases/tag/v1.8.0).

PR #410, opened 2024-04-26
Streaming API for CBC and CTR cipher modes

justsmth:
## Issues
* Resolves: #396

## Description
* New structs `StreamingEncryptionKey` and `StreamingDecryptionKey` supporting operations with AES128 and AES256 in CTR and CBC (w/ PKCS7 padding) modes
  * The `update` and `final` functions return a `BufferUpdate` struct indicating where the provided output buffer was modified.
* Added example to `cipher` module documentation.
* Added tests.

### DIFF
--- a/aws-lc-rs/src/cipher.rs
+++ b/aws-lc-rs/src/cipher.rs
@@ -83,6 +83,53 @@
 //! # }
 //! ```
 //!
+//! ### AES-128 CBC Streaming Cipher
+//!
+//! ```rust
+//! # use std::error::Error;
+//! #
+//! # fn main() -> Result<(), Box<dyn Error>> {
+//! use aws_lc_rs::cipher::{StreamingDecryptingKey, StreamingEncryptingKey, UnboundCipherKey, AES_128};
+//!
+//! let original_message = "This is a secret message!".as_bytes();
+//!
+//! let key_bytes: &[u8] = &[
+//!     0xff, 0x0b, 0xe5, 0x84, 0x64, 0x0b, 0x00, 0xc8, 0x90, 0x7a, 0x4b, 0xbf, 0x82, 0x7c, 0xb6,
+//!     0xd1,
+//! ];
+//!
+//! // Encrypt
+//! let mut ciphertext_buffer = vec![0u8; original_message.len() + AES_128.block_len()];
+//! let ciphertext_slice = ciphertext_buffer.as_mut_slice();
+//!
+//! let key = UnboundCipherKey::new(&AES_128, key_bytes)?;
+//! let mut encrypting_key = StreamingEncryptingKey::cbc_pkcs7(key)?;
+//! let written_slice = encrypting_key.update(original_message, ciphertext_slice)?;
+//! let written_len = written_slice.len();
+//! let remaining_slice = &mut ciphertext_slice[written_len..];
+//! let (context, written_slice) = encrypting_key.finish(remaining_slice)?;
+//! let ciphertext_len = written_len + written_slice.len();
+//! let ciphertext = &ciphertext_slice[0..ciphertext_len];
+//!
+//! // Decrypt
+//! let mut plaintext_buffer = vec![0u8; ciphertext_len + AES_128.block_len()];
+//! let plaintext_slice = plaintext_buffer.as_mut_slice();
+//!
+//! let key = UnboundCipherKey::new(&AES_128, key_bytes)?;
+//! let mut decrypting_key = StreamingDecryptingKey::cbc_pkcs7(key, context)?;
+//! let written_slice = decrypting_key.update(ciphertext, plaintext_slice)?;
+//! let written_len = written_slice.len();
+//! let remaining_slice = &mut plaintext_slice[written_len..];
+//! let written_slice = decrypting_key.finish(remaining_slice)?;
+//! let plaintext_len = written_len + written_slice.len();
+//! let plaintext = &plaintext_slice[0..plaintext_len];
+//!
+//! assert_eq!(original_message, plaintext);
+//! #
+//! # Ok(())
+//! # }
+//! ```
+//!
 //! ## Constructing a `DecryptionContext` for decryption.
 //!
 //! ```rust
@@ -142,8 +189,10 @@ pub(crate) mod block;
 pub(crate) mod chacha;
 pub(crate) mod key;
 mod padded;
+mod streaming;
 
 pub use padded::{PaddedBlockDecryptingKey, PaddedBlockEncryptingKey};
+pub use streaming::{StreamingDecryptingKey, StreamingEncryptingKey};
 
 use crate::buffer::Buffer;
 use crate::error::Unspecified;
@@ -280,7 +329,9 @@ impl Algorithm {
         &self.id
     }
 
-    const fn block_len(&self) -> usize {
+    /// The block length of this cipher algorithm.
+    #[must_use]
+    pub const fn block_len(&self) -> usize {
         self.block_len
     }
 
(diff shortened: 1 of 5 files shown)


aws-lc-rs/src/cipher.rs
```diff
@@ -905,4 +956,47 @@ mod tests {
         "eca7285d19f3c20e295378460e8729",
         "b5098e5e788de6ac2f2098eb2fc6f8"
     );
+
+    #[test]
+    fn streaming_cipher() {
```
camshaft, line 961: isn't this test copy/pasted from the doc test?

aws-lc-rs/src/cipher.rs
```diff
@@ -280,7 +329,9 @@ impl Algorithm {
         &self.id
     }
 
-    const fn block_len(&self) -> usize {
+    /// The block length of this cipher algorithm.
+    #[must_use]
```
camshaft, line 333: It's usually a good idea to say why the caller should use the result https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute

```suggestion
    #[must_use = "message goes here"]
```

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,527 @@
...
+    /// If the key is not valid for the cipher algorithm
+    pub fn cbc_pkcs7(key: UnboundCipherKey) -> Result<Self, Unspecified> {
+        let context = key.algorithm().new_encryption_context(OperatingMode::CBC)?;
+        Self::less_safe_cbc_pkcs7(key, context)
+    }
+
+    /// CBC cipher mode
+    /// # Errors
+    /// If the key is not valid for the cipher algorithm
+    pub fn less_safe_cbc_pkcs7(
```
camshaft, line 156: why is this less safe? Most of these methods could use additional documentation with more details

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,527 @@
...
+use crate::ptr::{LcPtr, Pointer};
+use aws_lc::{
+    EVP_CIPHER_CTX_new, EVP_CIPHER_iv_length, EVP_CIPHER_key_length, EVP_DecryptFinal_ex,
+    EVP_DecryptInit_ex, EVP_DecryptUpdate, EVP_EncryptFinal_ex, EVP_EncryptInit_ex,
+    EVP_EncryptUpdate, EVP_CIPHER_CTX,
+};
+use std::ptr::null_mut;
+
+/// A cipher encryption key for streaming encryption operations.
+pub struct StreamingEncryptingKey {
```
camshaft, line 17: Generally you don't need to prefix the struct with the module name (see https://rust-lang.github.io/rust-clippy/master/#/module_name_repetitions)
```suggestion
pub struct EncryptingKey {
```

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,527 @@
...
+    }
+
+    /// Encrypt the input and return the output.
+    /// # Errors
+    /// Returns an error if the output buffer is too small.
+    pub fn update<'a>(
+        &mut self,
+        input: &[u8],
+        output: &'a mut [u8],
+    ) -> Result<&'a [u8], Unspecified> {
```
camshaft, line 72: The `Ok(&[u8])` return value should be documented, since it's a bit weird

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,527 @@
...
+        } {
+            return Err(Unspecified);
+        }
+        let outlen: usize = outlen.try_into()?;
+        Ok(&output[0..outlen])
+    }
+
+    /// Finish the encryption and return the output.
+    /// # Errors
+    /// Returns an error if the output buffer is too small.
```
camshaft, line 96: This should document what is the minimum length

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,527 @@
...
+        Self::new(key, OperatingMode::CBC, context)
+    }
+}
+
+/// A cipher decryption key for streaming encryption operations.
+pub struct StreamingDecryptingKey {
+    algorithm: &'static Algorithm,
+    mode: OperatingMode,
+    cipher_ctx: LcPtr<EVP_CIPHER_CTX>,
+}
```
camshaft, line 169: nit
```suggestion
}

```

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,527 @@
...
+            algorithm,
+            mode,
+            cipher_ctx,
+            context,
+        })
+    }
+
+    /// Encrypt the input and return the output.
+    /// # Errors
+    /// Returns an error if the output buffer is too small.
```
camshaft, line 67: It also returns an error if it's too big (i.e. if it exceeds i32::MAX)

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,527 @@
...
+    /// # Errors
+    /// If the key is not valid for the cipher algorithm
+    pub fn ctr(key: UnboundCipherKey, context: DecryptionContext) -> Result<Self, Unspecified> {
+        Self::new(key, OperatingMode::CTR, context)
+    }
+
+    /// CBC cipher mode
+    /// # Errors
+    /// If the key is not valid for the cipher algorithm
+    pub fn cbc_pkcs7(
```
WillChilds-Klein, line 273: no padding-less option?

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,527 @@
...
+        }
+
+        let mut outlen: i32 = output.len().try_into()?;
+        let inlen: i32 = input.len().try_into()?;
+        if 1 != unsafe {
+            EVP_DecryptUpdate(
+                self.cipher_ctx.as_mut_ptr(),
+                output.as_mut_ptr(),
+                &mut outlen,
+                input.as_ptr(),
```
WillChilds-Klein, line 229: what if `input` and `output` overlap?

camshaft: > what if input and output overlap?

this is impossible with rust's aliasing rules (at least not without `unsafe`)

justsmth:
Some CI jobs are failing due to the the "nightly" Rust compiler panicking. The issue is reported here: https://github.com/rust-lang/rust/issues/125474

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,840 @@
...
+    pub fn cbc_pkcs7(
+        key: UnboundCipherKey,
+        context: DecryptionContext,
+    ) -> Result<Self, Unspecified> {
+        Self::new(key, OperatingMode::CBC, context)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
skmcgrail, line 364: Can you add tests to the existing cipher FIPS module indicator tests to ensure we have coverage for this new type?

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,840 @@
...
+            <usize>::try_from(unsafe { EVP_CIPHER_iv_length(*cipher) }).unwrap()
+        );
+
+        if 1 != unsafe {
+            EVP_EncryptInit_ex(
+                cipher_ctx.as_mut_ptr(),
+                *cipher,
+                null_mut(),
+                key_bytes.as_ptr(),
+                iv.as_ptr(),
```
skmcgrail, line 85: Might be worth leaving a comment behind to indicate that AWS-LC effectively copies the key and initalization vector and keeps it within the EVP context. So the references we pass here are allowed to drop after this call.

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,786 @@
...
+                self.cipher_ctx.as_mut_ptr(),
+                output.as_mut_ptr(),
+                &mut outlen,
+                input.as_ptr(),
+                inlen,
+            )
+        } {
+            return Err(Unspecified);
+        }
+        let outlen: usize = outlen.try_into()?;
```
skmcgrail, line 138: Is it worth asserting that outlen writen was actually less then or equal to `output.len()`?

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,786 @@
...
+        if 1 != indicator_check!(unsafe {
+            EVP_EncryptFinal_ex(
+                self.cipher_ctx.as_mut_ptr(),
+                output.as_mut_ptr(),
+                &mut outlen,
+            )
+        }) {
+            return Err(Unspecified);
+        }
+        let outlen: usize = outlen.try_into()?;
```
skmcgrail, line 169: Same note here on whether we should assert `outlen` being less then or equal to `output.len()`.

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,786 @@
...
+    }
+}
+
+/// A key for streaming decryption operations.
+pub struct StreamingDecryptingKey {
+    algorithm: &'static Algorithm,
+    mode: OperatingMode,
+    cipher_ctx: LcPtr<EVP_CIPHER_CTX>,
+}
+impl StreamingDecryptingKey {
```
skmcgrail, line 246: Just leaving a comment reminder to determine whether we should assert the `outlen` written check here as indicated above for update and finish on this type.

skmcgrail (commented):
Just some final comments that you can decide to take or not.

skmcgrail (approved):

WillChilds-Klein (approved):

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -0,0 +1,790 @@
...
+    pub fn update<'a>(
+        &mut self,
+        input: &[u8],
+        output: &'a mut [u8],
+    ) -> Result<BufferUpdate<'a>, Unspecified> {
+        let min_outsize = input
+            .len()
+            .checked_add(self.algorithm().block_len())
+            .ok_or(Unspecified)?
+            - 1;
```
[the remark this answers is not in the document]
justsmth: That might be true for most/all(?) cipher modes, but I believe the API contract/documentation doesn't provide any guarantee that it will necessarily be anything less than (`input_len` + `block_len` - 1). So I went with the most conservative restriction for it.

PR #418, opened 2024-05-06
Cleanup CI and build configs

justsmth:
### Description of changes: 
* Remove last remaining usage of `actions-rs` GitHub action.
* Fix riscv cross-compilation -- download latest to build cross, then download toolchain for target.
* Bump MSRV to 1.63 -- [Rustls is now at 1.63](https://github.com/rustls/rustls/blob/main/rustls/Cargo.toml#L5) as well.
* Ignoring the `unexpected_cfgs`. We can't properly handle these until we're at MSRV 1.77.
* Remove various erroneous reference to non-existent `cfg` values.

### DIFF
--- a/aws-lc-rs/src/agreement/ephemeral.rs
+++ b/aws-lc-rs/src/agreement/ephemeral.rs
@@ -158,15 +158,14 @@ mod tests {
             &mut k,
             &mut u,
         );
-
-        if cfg!(feature = "slow_tests") {
-            expect_iterated_x25519(
-                "7c3911e0ab2586fd864497297e575e6f3bc601c0883c30df5f4dd2d24f665424",
-                10_000..1_000_000,
-                &mut k,
-                &mut u,
-            );
-        }
+        /*
+               expect_iterated_x25519(
+                   "7c3911e0ab2586fd864497297e575e6f3bc601c0883c30df5f4dd2d24f665424",
+                   10_000..1_000_000,
+                   &mut k,
+                   &mut u,
+               );
+        */
     }
 
     #[test]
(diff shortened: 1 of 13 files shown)


skmcgrail (approved):

aws-lc-rs/src/agreement/ephemeral.rs
```diff
@@ -158,15 +158,14 @@ mod tests {
...
-                &mut u,
-            );
-        }
+        /*
+               expect_iterated_x25519(
+                   "7c3911e0ab2586fd864497297e575e6f3bc601c0883c30df5f4dd2d24f665424",
+                   10_000..1_000_000,
+                   &mut k,
+                   &mut u,
+               );
```
WillChilds-Klein, line 167: why disable this test? is it too slow?

aws-lc-rs/tests/hmac_test.rs
```diff
@@ -16,11 +16,8 @@ fn hmac_tests() {
         let output = test_case.consume_bytes("Output");
 
         let algorithm = {
-            let digest_alg = match digest_alg {
-                Some(digest_alg) => digest_alg,
-                None => {
-                    return Ok(());
-                } // Unsupported digest algorithm
+            let Some(digest_alg) = digest_alg else {
```
WillChilds-Klein, line 19: ooh, nice. TIL you can skip the if in an if/let

aws-lc-rs/src/agreement/ephemeral.rs
justsmth: Yeah, really slow.  This was never executing anyways due to being conditioned on a non-existent feature.  However, I'm not opposed too putting a guard around such tests so that they can still execute, but only when we're prepared to wait.

PR #434, opened 2024-06-05
feat: support openharmony platform

southorange0929:
### Description of changes: 

When using cmake for building, make sure to include the cmake files and parameters that openharmony needs to ensure a successful build.

### Call-outs:
Point out areas that need special attention or support during the review process. Discuss architecture or design changes.

### Testing:
How is this change tested (unit tests, fuzz tests, etc.)? Are there any testing steps to be verified by the reviewer?

### DIFF
--- a/aws-lc-fips-sys/builder/cmake_builder.rs
+++ b/aws-lc-fips-sys/builder/cmake_builder.rs
@@ -3,7 +3,7 @@
 
 use crate::OutputLib::{Crypto, RustWrapper, Ssl};
 use crate::{
-    cargo_env, execute_command, is_no_asm, option_env, target, target_arch, target_os,
+    cargo_env, execute_command, is_no_asm, target, target_arch, target_env, target_os,
     target_vendor, OutputLibType,
 };
 use std::collections::HashMap;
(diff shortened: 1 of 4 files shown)


southorange0929:
> Are there any tools (similar to [cross-rs](https://github.com/cross-rs/cross) or [cargo-xwin](https://github.com/rust-cross/cargo-xwin)) that would allow us to test compilation to an OpenHarmony target in our CI?

No, in fact, there is currently no very suitable way to run code whose target architecture is openharmony, even rust or openharmony officially does not have it yet. When there is a suitable way, I will try to add the corresponding CI to support

aws-lc-fips-sys/builder/cmake_builder.rs
```diff
@@ -150,6 +150,17 @@ impl CmakeBuilder {
                 cmake_cfg.define("CMAKE_OSX_ARCHITECTURES", "arm64");
             }
         }
+
+        if target_env() == "ohos" {
+            let ndk = env::var("OHOS_NDK_HOME").expect("Try to get ohos ndk home failed.");
```
justsmth, line 155: This concern applies to the other change as well...

* Instead of panicking the build, should this be conditioned on the required environment variable being set.
```
if target_env() == "ohos" && env::var("OHOS_NDK_HOME").is_some() {
```
* If the environment already has a `CMAKE_TOOLCHAIN_FILE` set, I think that would take precedence over the `CMAKE_TOOLCHAIN_FILE` being set here. (This change will conflict with #431, so we'll need to consider this during a merge.)

southorange0929: You are right, our code is more robust now. Thanks

skmcgrail (approved):

samuel40791765 (approved):

PR #488, opened 2024-08-05
cipher: clarify the EncryptingKey docs

cpu:
### Issues:
N/A

### Description of changes: 

While working with the `aws_lc_rs::cipher` module downstream in https://github.com/rustls/rustls/pull/2066 we noticed the documentation here could use a small touch-up.

Previously the `aws_lc_rs::cipher::EncryptingKey` type's `encrypt()` and `less_safe_encrypt()` functions were documented as returning a reference to the encrypted data. However, the implementation uses an `in_out: &mut [u8]` argument for the encrypted data and **returns** a `DecryptionContext`, not a ref. to ciphertext.

Similarly, it feels important to call out that the `DecryptionContext` when using `encrypt()` contains a randomly generated IV. This makes usage clearer and also provides good counter-point to the disclaimer on `less_safe_encrypt()` about the provenance (and safety) of the IV being unknown.

### Call-outs:
None that I can think of :-)

### Testing:
N/A - docs only update.


justsmth (approved):
Thanks!

skmcgrail (approved):
Thanks!

Issue #364, opened 2024-03-06
Remove NASM requirement for windows builds


justsmth:
See: aws/aws-lc#1477

justsmth:
There's now a draft PR regarding this issue here: https://github.com/aws/aws-lc-rs/pull/491

Please let us any thoughts/concerns you have on the approach taken for this.

justsmth:
This is resolved with the release of [aws-lc-rs v1.9.0](https://github.com/aws/aws-lc-rs/releases/tag/v1.9.0).

camshaft:
Should we close https://github.com/aws/aws-lc/issues/1477 as well?

PR #491, opened 2024-08-06
Pregeneration of NASM build artifacts

justsmth:
### Issues:
Addresses #364

### Description of changes: 
* Provide prebuilt NASM artifacts for build environments that lack a NASM assembler.

### Call-outs:
* Usage of NASM artifacts is an "opt-in". It is determined as follows (subject to further discussion):
  * `AWS_LC_SYS_PREBUILT_NASM=1` - Prebuilt artifacts used only if NASM was not found.
  * `AWS_LC_SYS_PREBUILT_NASM=0` - NASM is required.
  * Unset - NASM is required.

### Testing:
* Added the "prebuilt NASM verification" step that disassembles newly built NASM objects. It compares their disassembly to the disassembly of the NASM objects already in the repo.  The CI job fails if they differ.


djc:
Have you considered using a Cargo feature instead of an environment variable to make this work? That seems like a solution with better ergonomics for most downstream Rust applications.

justsmth:
> Have you considered using a Cargo feature instead of an environment variable to make this work? That seems like a solution with better ergonomics for most downstream Rust applications.

We considered using a feature for this. The concern is that we want the decision of whether its acceptable to use the pre-built binaries to be made by the end consumer within their build environment. Because features are only "additive" and features can be enabled by any crate that's part of a build, it seems like the wrong mechanism. 

As a feature, the consumer would essentially be required to "delegate" the decision of whether to use NASM pre-built binaries to any/every crate in their dependency tree. One possible compromise might be to expose a feature that specifies what to default to when `AWS_LC_SYS_PREBUILT_NASM` is not set in the environment -- but this also seems to violate the nature/purpose of "features" as the environment will be "subtracting" a "feature" that a dependent crate indicated that it needed.

We considered both "environment variables" and "features" as mechanisms for enabling this. Is there another option that would also enable intermediate crates (i.e., those part of the dependency tree between us an the consumer) to influence what the default choice is?

djc:
> We considered both "environment variables" and "features" as mechanisms for enabling this. Is there another option that would also enable intermediate crates (i.e., those part of the dependency tree between us and the end consumer) to influence what the default choice is?

No, I don't think so. Your rationale makes sense to me, unfortunately Cargo doesn't support anything like this right now (see https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618 for some discussion -- and if you think there is appetite for sponsoring work on this at AWS, I'd be interested in discussing that!). As a downstream application builder I would still prefer being able to control this via features, maybe just with documentation guidance that libraries should not touch this and potentially keeping a way to override it via the environment variable?

justsmth:
@djc -- Thanks for your feedback!  

We are still discussing this issue internally. We may follow up in a subsequent PR to adjust the build environment conditions that determine whether the pre-built NASM objects are allowed to be used.

aws-lc-sys/builder/prebuilt-nasm.bat
```diff
@@ -0,0 +1,21 @@
+@echo off
```
skmcgrail, line 1: Missing copyright header

aws-lc-sys/builder/main.rs
```diff
@@ -386,6 +388,10 @@ fn is_no_asm() -> bool {
     unsafe { AWS_LC_SYS_NO_ASM }
 }
 
+fn allow_prebuilt_nasm() -> Option<bool> {
+    unsafe { AWS_LC_SYS_PREBUILT_NASM }
+}
```
skmcgrail, line 393: Any reason to not just do:

```suggestion
fn allow_prebuilt_nasm() -> bool {
    unsafe { AWS_LC_SYS_PREBUILT_NASM } == Some(true)
}
```

.github/workflows/tests.yml
```diff
@@ -205,6 +206,40 @@ jobs:
...
+        shell: bash
+        run: AWS_LC_SYS_PREBUILT_NASM=0 cargo test --tests -p aws-lc-rs --release --no-default-features --features aws-lc-sys
+      - name: Collect NASM outputs
+        shell: bash
+        run: ./scripts/build/collect_nasm_obj.sh
+      - name: Flag any NASM changes
+        shell: bash
+        run: |
+          git add .
+          git diff --cached --exit-code HEAD -- aws-lc-sys/builder/prebuilt-nasm/*.txt
```
andrewhop, line 241: Double checking my understanding: if there is a difference this will exit unsuccessfully and fail the PR?

aws-lc-sys/builder/cmake_builder.rs
```diff
@@ -298,10 +311,14 @@ impl crate::Builder for CmakeBuilder {
...
+            if target_arch() == "x86_64"
+                && !is_no_asm()
+                && !test_nasm_command()
+                && Some(true) != allow_prebuilt_nasm()
+            {
                 eprintln!(
-                    "Consider setting `AWS_LC_SYS_NO_ASM` in the environment for development builds.\
-                See User Guide about the limitations: https://aws.github.io/aws-lc-rs/index.html"
+                    "Consider setting `AWS_LC_SYS_PREBUILT_NASM` in the build environment.\
+                See User Guide: https://aws.github.io/aws-lc-rs/index.html"
```
andrewhop, line 321: Deep link to the windows building page?

aws-lc-sys/builder/prebuilt-nasm.bat
andrewhop, line 1: What does this script do? Why can't the rust program that call this just find all the nasm files itself?

justsmth: What this script does is imitate the part of NASM during the build. When called, it identifies the NASM object corresponding to the indicated source file, and places it into the output directory. We configure Cmake to treat this script as if it were NASM.

aws-lc-sys/builder/main.rs
justsmth: The return value here needs to represent 3 possibilities: (1) the environment variable is set to *true*, (2) the environment variable is set to *false*, and (3) the environment variable is not set.

.github/workflows/tests.yml
justsmth: Yeah, the `--exit-code` option is set so that it will provide the correct exit statis:
```
❯ git help diff | grep -A 2 --regexp='^\s*\-\-exit-code'
       --exit-code
           Make the program exit with codes similar to diff(1). That is, it
           exits with 1 if there were differences and 0 means no differences.
```

aws-lc-sys/builder/cmake_builder.rs
justsmth: I thought about that, but there's some refactoring I want to do to the User Guide so I'm hesitant to deep link.  In this case, the most relevant information might actually be on the "Resources" page where we list the environment variables that can affect the build: https://aws.github.io/aws-lc-rs/resources.html

skmcgrail (approved):

andrewhop (approved):

PR #498, opened 2024-08-16
Support for compressed and X.509 public key format

skmcgrail:
### Description of changes: 
* Support for serializing ECDSA and ECDH keys in compressed and uncompressed EC point octet format, and support for X.509 SubjectPublicKeyInfo format.
* Support for parsing ECDSA and ECDH keys that may be in compressed, uncompressed, or X.509 format. First attempting to parse as X.509, and then falling back to EC point format before failing the parse attempt.
* X2559 supports X.509 as well, will follow up Ed2559 in a separate PR since that requires a bit more refactor work and wanted to simplify this PR a bit.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

PR #506, opened 2024-08-19
Broader Ed25519 Key Format Support

skmcgrail:
### Description of changes: 
* Similar goal to https://github.com/aws/aws-lc-rs/pull/498
* Support parsing of raw ED25519 public keys or X.509 SubjectPublicInfo when performing signature validation
* Support for X.509 SubjectPublicInfo serialization for ED25519 PublicKey using AsDer trait.
* Add support for the AsDer for PKCS#8 v1 and v2 for Ed25519KeyPair for consistency with our serialization trait design.


aws-lc-rs/src/ed25519.rs
```diff
@@ -269,24 +338,58 @@ impl Ed25519KeyPair {
...
+            ED25519_PRIVATE_KEY_LEN,
+            ED25519_PRIVATE_KEY_SEED_LEN + ED25519_PUBLIC_KEY_LEN
+        );
+        let mut out_len: usize = private_key.len();
+        if 1 != unsafe {
+            EVP_PKEY_get_raw_private_key(
+                *evp_pkey.as_const(),
+                private_key.as_mut_ptr(),
+                &mut out_len,
+            )
```
justsmth, line 369: Instead of holding on to the "raw bytes" private key in the `Ed25519KeyPair` struct, why not just recompute it on demand?  Maintaining the secret in the struct isn't needed, and it seems unlikely to be needed very often since an `Ed25519KeyPair` can only be constructed from its serialized form.

skmcgrail: Because `Ed25519KeyPair::seed` awkwardly returns `Seed` which has a lifetime parameter bound to `Ed25519KeyPair` that it uses to reference the private key bytes. If you recall our conversation prior I believe we were in agreement that there was concerns with changing that lifetime parameter?

justsmth (approved):

justsmth (dismissed):

samuel40791765 (dismissed):

justsmth (approved):

samuel40791765 (approved):

PR #509, opened 2024-08-28
Update aws-lc-sys, align w/ AWS-LC v1.34.2

justsmth:
### Description of changes: 
* Aligns aws-lc-sys v0.21.0 w/ [AWS-LC v1.34.2](https://github.com/aws/aws-lc/releases/tag/v1.34.2)
  * Adds support for a `AWS_LC_SYS_C_STD` environment variable to determine which C standard to use.
  * For platforms that build using "cc" (instead of CMake), the build defaults to using the C11 standard.

### Call-outs:
* Automated workflow is here: https://github.com/aws/aws-lc-rs/actions/runs/10598652101
* Only [first](https://github.com/aws/aws-lc-rs/pull/509/commits/762461f219108b1976596e2235eb75317a998410) commit is manual.
* The C11 build doesn't contain the `CRYPTO_refcount_dec_and_test_zero` or `CRYPTO_refcount_inc` symbols. We need these symbols to be prefixed for C99 build. Thus, for some targets in the symbol collection step of the automation are being forced to use a C99 build.


skmcgrail (approved):

aws-lc-sys/builder/cmake_builder.rs
```diff
@@ -135,6 +135,15 @@ impl CmakeBuilder {
...
         }
+        match requested_c_std() {
+            CStdRequested::C99 => {
+                cmake_cfg.define("CMAKE_C_STANDARD", "99");
+            }
+            CStdRequested::C11 => {
+                cmake_cfg.define("CMAKE_C_STANDARD", "11");
+            }
+            CStdRequested::None => {}
+        }
```
andrewhop, line 146: How does the aws-lc-rs' CMake build work? Does it run aws-lc's regular CMakelists.txt? If so why not let it figure out the C standard to use?

aws-lc-sys/builder/main.rs
```diff
@@ -314,13 +314,39 @@ trait Builder {
...
+    None,
+}
+
+impl CStdRequested {
+    fn from_env() -> Self {
+        if let Some(val) = option_env("AWS_LC_SYS_C_STD") {
+            let cstd = match val.as_str() {
+                "99" => CStdRequested::C99,
+                "11" => CStdRequested::C11,
+                _ => CStdRequested::None,
```
andrewhop, line 330: If it's something we can't parse should this be a hard error? If someone set `AWS_LC_SYS_C_STD=c11` it would hit none here, and then could they actually get c99, or vice versa.

aws-lc-sys/builder/cmake_builder.rs
justsmth: In most build environments, we will not be setting `CMAKE_C_STANDARD`. This logic would only affect the build when `AWS_LC_SYS_C_STD` has been set to "99" or "11" in the build environment, which should rarely be needed.

aws-lc-sys/builder/main.rs
justsmth: Yeah, I thought about matching against a RegExp here to add more flexibility, but I'm not sure it's worth the trouble.

I see this environment variable as something that will rarely be needed. We will default to C11 for most consumers (which should work for any build host that's been setup sometime in the last decade).

zh-jq:
Is it possible to update `bindgen` to 0.70 for `aws-lc-sys` v0.21.0?

andrewhop (approved):

Issue #495, opened 2024-08-13
dynamic linking on mac causes linking failures when running cargo installed executable.


rmullin-apptio:
### Problem:

It seems that static linking for aws-lc when compiled in fips mode is only available on Linux. When compiled on Mac M1, the library is dynamically linked. The program can be run through `cargo run`, but if it's installed using `cargo install`, the libraries aren't installed with it, so it crashes on startup:

```
» ./rust-fips-test                                                                                                                                                                  
dyld[53187]: Library not loaded: @rpath/libaws_lc_fips_0_12_11_crypto.dylib
  Referenced from: <A3DAE490-6F88-3CA7-9D27-464C9314D5D8> /Users/ryan/src/rust-fips-test/target/debug/rust-fips-test
  Reason: no LC_RPATH's found
[1]    53187 abort      ./rust-fips-test
```

### Solution:

Right now the only way I've found to get this to work is to manually copy the dylib files into the same directory as the output executable and use install_name_tool to add @loader_path to the output executable's rpath:

```
cp build/aws-lc-fips-sys-2da5c751fdbe11cd/out/build/artifacts/libaws_lc_fips_0_12_11_crypto.dylib .
cp build/aws-lc-fips-sys-2da5c751fdbe11cd/out/build/artifacts/libaws_lc_fips_0_12_11_rust_wrapper.dylib .
install_name_tool -add_rpath @loader_path ./the-executable
```

After that it works. I'm unsure why this restriction exists outside of linux so am unsure if it's possible to statically link the fips library. At the very least if this is necessary it may be worthwhile to add documentation addressing this, and what the best approach is for installing fips versions of rust programs built with this library.

* **Does this change any public APIs?** No
* **Which algorithm(s) will this impact?** None

### Requirements / Acceptance Criteria:

Provide documentation on the optimal way to install fips enabled rust programs on non-linux platforms.

* **RFC links:** N/A
* **Related Issues:** It appears this behavior was added in [this PR](https://github.com/aws/aws-lc-rs/pull/173)
* **Will the Usage Guide or other documentation need to be updated?** Yes
* **Testing:** How will this change be tested? Call out new integration tests, functional tests, or particularly
  interesting/important unit tests. N/A.

justsmth:
Hello!

Yeah, unfortunately this is a limitation of the AWS-LC FIPS build, which currently can only provide a static library on the Linux platform.  

[...]

PR #511, opened 2024-08-29
Add Feature "prebuilt-nasm"; Update documentation

justsmth:
### Description of changes: 
* Introduce "prebuilt-nasm" feature [as requested in previous PR](https://github.com/aws/aws-lc-rs/pull/491#issuecomment-2275067463)
  * Added documentation about this feature.
* Numerous documentation fixes/updates
* Update User Guide
* Addresses missing documentation identified in: #495 

### Testing:
* Added CI test for new feature and environment variable.


aws-lc-rs/README.md
```diff
@@ -70,6 +70,30 @@ the pre-generated bindings. This feature requires `libclang` to be installed. Se
...
+
+For Windows x86 and x86-64, the AWS-LC assembly code requires NASM to build. On these platforms, if a NASM assembler is
+found during the build process, then it *will* be used to compile the assembly files. However, if a NASM assembler is
+not found, then the build may fail except in the following cases:
+
+* The `AWS_LC_SYS_PREBUILT_NASM` environment variable is found and has a value of "1".
+* `AWS_LC_SYS_PREBUILT_NASM` is *not found* in the environment and the "prebuilt-nasm" feature has been enabled.
+
+If the above cases apply, then our prebuilt NASM objects may be used for the build. To prevent our prebuilt NASM
+objects from being used, please set `AWS_LC_SYS_PREBUILT_NASM=0` in your build environment to prevent their use.
```
darylmartin100, line 83: or install NASM

aws-lc-rs/README.md
```diff
@@ -70,6 +70,30 @@ the pre-generated bindings. This feature requires `libclang` to be installed. Se
 [requirements](https://rust-lang.github.io/rust-bindgen/requirements.html)
 for [rust-bindgen](https://github.com/rust-lang/rust-bindgen)
 
+##### prebuilt-nasm
+
+For Windows x86 and x86-64, the AWS-LC assembly code requires NASM to build. On these platforms, if a NASM assembler is
+found during the build process, then it *will* be used to compile the assembly files. However, if a NASM assembler is
+not found, then the build may fail except in the following cases:
```
darylmartin100, line 77: I would add a sentence recommending that people install NASM instead of using this flag as an easy fix.

aws-lc-rs/README.md
```diff
@@ -70,6 +70,30 @@ the pre-generated bindings. This feature requires `libclang` to be installed. Se
 [requirements](https://rust-lang.github.io/rust-bindgen/requirements.html)
 for [rust-bindgen](https://github.com/rust-lang/rust-bindgen)
 
+##### prebuilt-nasm
```
darylmartin100, line 73: nit:  Should the flag be use-prebuilt-nasm?

darylmartin100, line 73: I think we should document that the feature flag is honored for all indirect dependencies of the crate that enables it.  We should caution against use without understanding the impact for all dependencies.

justsmth: I think a more accurate name would be "allow-prebuilt-nasm", since we will still use the NASM assembler if found. I prefer the shorter "prebuilt-nasm", but I could likely be convinced otherwise.

darylmartin100: sounds good to me

aws-lc-rs/src/lib.rs
```diff
@@ -71,6 +71,41 @@
...
+//!
+//! ## About prebuilt NASM objects
+//!
+//! Our prebuilt NASM objects are generated using the same automation that produces our pregenerated bindings. See our
+//! [GitHub workflow configuration](https://github.com/aws/aws-lc-rs/blob/main/.github/workflows/sys-bindings-generator.yml).
+//! The prebuilt NASM objects are checked into our repository
+//! and are [available for inspection](https://github.com/aws/aws-lc-rs/tree/main/aws-lc-sys/builder/prebuilt-nasm).
+//! For each PR submitted,
+//! [our CI verifies](https://github.com/aws/aws-lc-rs/blob/8fb6869fc7bde92529a5cca40cf79513820984f7/.github/workflows/tests.yml#L209-L241)
+//! that the NASM objects newly built from source match the NASM objects currently in the repository.
```
skmcgrail, line 107:
```suggestion
//! Prebuilt NASM objects are generated using automation similar to the crate provided pregenerated bindings. See the repositories
//! [GitHub workflow configuration](https://github.com/aws/aws-lc-rs/blob/main/.github/workflows/sys-bindings-generator.yml) for more information.
//! The prebuilt NASM objects are checked into the repository
//! and are [available for inspection](https://github.com/aws/aws-lc-rs/tree/main/aws-lc-sys/builder/prebuilt-nasm).
//! For each PR submitted,
//! [CI verifies](https://github.com/aws/aws-lc-rs/blob/8fb6869fc7bde92529a5cca40cf79513820984f7/.github/workflows/tests.yml#L209-L241)
//! that the NASM objects newly built from source match the NASM objects currently in the repository.
```

aws-lc-rs/src/lib.rs
```diff
@@ -71,6 +71,41 @@
...
+//! by enabling this feature, it is enabled for crates within the same build.
+//!
+//! # Use of prebuilt NASM objects
+//!
+//! For Windows x86 and x86-64, the AWS-LC assembly code requires NASM to build. On these platforms,
+//! we recommend that you install [the NASM assembler](https://www.nasm.us/). If a NASM assembler is
+//! found during the build process, then it *will* be used to compile the assembly files. However,
+//! if a NASM assembler is not found, then the (non-"fips") build may fail except in the following case:
+//!
+//! * You are building for `x86-64` and either:
```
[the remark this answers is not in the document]
justsmth: We are only providing prebuilt NASM objects for x86-64, so this condition won't hold for `x86`.

aws-lc-rs/src/lib.rs
```diff
@@ -71,6 +71,41 @@
...
+//!    * The `AWS_LC_SYS_PREBUILT_NASM` environment variable is found and has a value of "1" (or the value is empty); OR
```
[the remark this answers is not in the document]
justsmth: Currently our logic checks for "falsey" values [("0", "n", "off")](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-sys/builder/main.rs#L77-L95) and treats any other value as having set the variable to true.

I'm ok with this. What are your thoughts? One option would be to also check for "truthy" values (e.g., "1", "y", and "on") -- and treat any value that's neither "truthy" nor "falsey" the same as if the variable was unset.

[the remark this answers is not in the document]
skmcgrail: In a perfect world I would say we error when the environment variable is set to an unsupported/unknown value. But that creates issues with different library versions and expectations. I'd personally learn towards treating unknown values as the variable being unset? That way it just sticks to the default behavior of the crate as if the variable didn't affect anything due to it being invvalid.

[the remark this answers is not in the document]
justsmth: Updated in the latest commit.

skmcgrail (approved):

aws-lc-rs/README.md
```diff
@@ -70,6 +70,42 @@ the pre-generated bindings. This feature requires `libclang` to be installed. Se
...
+Use of prebuilt NASM objects is prevented if either of the following conditions are true:
+* The NASM assembler is detected in the build environment
+* `AWS_LC_SYS_PREBUILT_NASM` environment variable is set with a value of `0`
+
+Be aware that [features are additive](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification);
+by enabling this feature, it is enabled for crates within the same build.
+
+## Use of prebuilt NASM objects
+
+For Windows x86 and x86-64, NASM is required for assembly code compilation. On these platforms,
```
andrewhop, line 87: Does windows arm work with our arm assembly without nasm?

aws-lc-sys/builder/main.rs
```diff
@@ -77,21 +77,28 @@ fn option_env<N: AsRef<str>>(name: N) -> Option<String> {
...
+            eprintln!("Parsed: {name}=false");
+            return Some(false);
         }
-    } else {
-        None
+        if env_var_value.starts_with(|c: char| c.is_ascii_digit())
+            || env_var_value.starts_with('y')
+            || env_var_value.starts_with("on")
+            || env_var_value.starts_with('t')
+        {
```
andrewhop, line 95: Should this handle `1` as true like you do for false up above?

aws-lc-rs/README.md
justsmth: For Windows/ARM we use the clang-cl (part of MSVC) which can handle assembly code, so there's no need for a separate assembler.

aws-lc-sys/builder/main.rs
justsmth: I thought about only accepting "1" for `true`, but then thought it's also reasonable to accept "2" as `true`. Which led me to thinking that any non-zero digit could reasonably be considered `true` if being parsed into a `boolean` value.

andrewhop (approved):

PR #580, opened 2024-11-08
Generate aws-lc-fips-sys with openssl/ocsp.h

skmcgrail:

### DIFF
--- a/aws-lc-fips-sys/Cargo.toml
+++ b/aws-lc-fips-sys/Cargo.toml
@@ -1,8 +1,8 @@
 [package]
 name = "aws-lc-fips-sys"
 description = "AWS-LC is a general-purpose cryptographic library maintained by the AWS Cryptography team for AWS and their customers. This is the FIPS validated version of AWS-LC."
-version = "0.12.13"
-links = "aws_lc_fips_0_12_13"
+version = "0.13.0"
+links = "aws_lc_fips_0_13_0"
 authors = ["AWS-LC"]
 edition = "2021"
 repository = "https://github.com/aws/aws-lc-rs"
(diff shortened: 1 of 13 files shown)


aws-lc-fips-sys/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
 name = "aws-lc-fips-sys"
 description = "AWS-LC is a general-purpose cryptographic library maintained by the AWS Cryptography team for AWS and their customers. This is the FIPS validated version of AWS-LC."
-version = "0.12.13"
-links = "aws_lc_fips_0_12_13"
+version = "0.13.0"
```
justsmth, line 4: Since this only adds a new header file (which seems unlikely to break backwards compatibility), I think we should make the version `0.12.14`.

aws-lc-fips-sys/builder/main.rs
```diff
@@ -476,6 +527,8 @@ fn main() {
         println!("cargo:libssl={}_ssl", prefix_string());
     }
 
+    println!("cargo:conf={}", OSSL_CONF_DEFINES.join(","));
```
justsmth, line 530: Where will this be used?

skmcgrail: The rust-openssl crate largely uses these for toggling APIs that should be available.

skmcgrail: BoringSSL emits these as well, I could see them being generally useful for integrators of the sys crates that won't to be able to inter-opt between the different offerings.

samuel40791765 (approved):

PR #582, opened 2024-11-10
Implement TryInto<PublicEncryptingKey> for PublicKeyComponents

### LINKED ISSUE #581: Support importing RSA parameters manually for encryption with Pkcs1PublicEncryptingKey
### Security issue notifications

If you discover a potential security issue in AWS-LC for Rust we ask that you notify AWS Security via our
[vulnerability reporting page](http://aws.amazon.com/security/vulnerability-reporting/). Please do **not** create a
public github issue, if in doubt contact AWS security first.

### Problem:

At the moment it seems there is no way to get a `Pkcs1PublicEncryptingKey` with a custom `n` and `e`.
There already exists `PublicKeyComponents::verify` but this only handles signature verification, so having something for encryption does not seem out of scope.

### Solution:

I think the simplest and ideal solution is to allow getting a `PublicEncryptingKey` from a `PublicKeyComponents`. This would best be done by adding a `build_encrypting_key ()` helper. This should be trivial as all that would be required is to call `build_rsa()` and pass the resulting `EVP_PKEY` to `PublicEncryptingKey::new`.

* **Does this change any public APIs?** Yes, would ideally allow converting a `PublicKeyComponents` to a `PublicEncryptingKey`
* **Which algorithm(s) will this impact?** RSA

* **RFC links:** N/A
* **Related Issues:** N/A
* **Will the Usage Guide or other documentation need to be updated?** Yes
* **Testing:** I am not sure about the testing requirements, from my eyes it seems trivial to implement.

### Out of scope:

Enforcing the secure usage of these apis would likeley be out of scope, if the user wants to manually supply RSA parameters its likely for some more low-level operation.

Vadmeme:
### Issues:
Resolves #581 

### Description of changes: 
Adds `impl TryInto<PublicEncryptingKey> for PublicKeyComponents`.

### Call-outs:
N/A

### Testing:
Im not sure how to implement unit tests for this change would need a reviewer to assist with this. I tested them in my personal application and they functioned as expected.

### DIFF
--- a/aws-lc-rs/src/rsa/key.rs
+++ b/aws-lc-rs/src/rsa/key.rs
@@ -14,13 +14,13 @@ use crate::io;
 use crate::ptr::ConstPointer;
 use crate::{
     digest::{self},
-    rsa::{PublicEncryptingKey},
     encoding::{AsDer, Pkcs8V1Der},
     error::{KeyRejected, Unspecified},
     fips::indicator_check,
     hex,
     ptr::{DetachableLcPtr, LcPtr},
     rand,
+    rsa::{PublicEncryptingKey},
     sealed::Sealed,
 };
 #[cfg(feature = "fips")]
--- a/aws-lc-rs/src/rsa/key.rs
+++ b/aws-lc-rs/src/rsa/key.rs
@@ -20,7 +20,7 @@ use crate::{
     hex,
     ptr::{DetachableLcPtr, LcPtr},
     rand,
-    rsa::{PublicEncryptingKey},
+    rsa::PublicEncryptingKey,
     sealed::Sealed,
 };
 #[cfg(feature = "fips")]
@@ -367,7 +367,7 @@ impl PublicKey {
     }
 }
 
-/// Low-level API for the verification of RSA signatures.
+/// Low-level API for RSA public keys.
 ///
 /// When the public key is in DER-encoded PKCS#1 ASN.1 format, it is
 /// recommended to use `aws_lc_rs::signature::verify()` with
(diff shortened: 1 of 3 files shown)


skmcgrail (approved):

aws-lc-rs/src/rsa/key.rs
```diff
@@ -432,6 +433,16 @@ where
...
 
+    /// Builds a `PublicEncryptingKey` from the public key components.
+    ///
+    /// # Errors
+    /// `error::Unspecified` if the key failed to verify.
+    pub fn build_encrypting_key(&self) -> Result<PublicEncryptingKey, Unspecified> {
+        let rsa = self.build_rsa()?;
+
+        PublicEncryptingKey::new(rsa)
+    }
```
justsmth, line 444: It might be more idiomatic (more useful?) to implement this through the `TryInto` trait instead of an explicit function?

skmcgrail (approved):

skmcgrail (approved):

aws-lc-rs/src/rsa/key.rs
```diff
@@ -456,6 +457,22 @@ where
     }
 }
 
+impl<B> TryInto<PublicEncryptingKey> for PublicKeyComponents<B>
```
justsmth, line 460: I'm not sure it's needed, but we could potentially have a non-consuming implementation:
```
impl<B> TryInto<PublicEncryptingKey> for &PublicKeyComponents<B>
...
``` 
Its use would be more cumbersome: `TryInto::try_into(&public_key)?` or `(&public_key).try_into()?`.

justsmth (approved):

PR #585, opened 2024-11-12
Support for AES CFB (128-bit) mode

### LINKED ISSUE #575: [Feature Request] Support CFB mode for AES
### Problem:

The [TPM2 specification Part 1: Architecture 24.4 Symmetric Encrypt](https://trustedcomputinggroup.org/resource/tpm-library-specification/) requires the use of an AES 128 bit key using CFB mode. Currently CFB mode is not exposed in aws-ls-rs which blocks us from using it for the particular use case.

### Requirements:

- Encryption using an AES 128 bit key with CFB mode
- The IV vector needs to be configurable (in the use case it will be set to all zero)

skmcgrail:
### Issues:
Resolves #575

### Description of changes: 
Adds support for AES cipher feedback 128-bit mode (CFB-128). I was going to also support CFB 1-bit and CFB 8-bit modes but have a smaller concern around being able to support that with the fips feature enabled currently. Need to investigate that further, so we may come back at a later date and add support for those additional bit-mode variants. For now this resolves the requester's needs.

### DIFF
--- a/aws-lc-rs/src/cipher.rs
+++ b/aws-lc-rs/src/cipher.rs
@@ -134,7 +134,7 @@
 //! # Ok(())
 //! # }
 //! ```
-//! 
+//!
 //! ### AES-128 CFB 128-bit mode
 //!
 //! ```rust
@@ -671,13 +671,12 @@ fn encrypt(
     let block_len = algorithm.block_len();
 
     match mode {
-        // TODO: Hopefully support CFB1, and CFB8
-        OperatingMode::CTR | OperatingMode::CFB128 => {}
-        _ => {
+        OperatingMode::CBC => {
             if (in_out.len() % block_len) != 0 {
                 return Err(Unspecified);
             }
         }
+        _ => {}
     }
 
     match mode {
@@ -706,13 +705,12 @@ fn decrypt<'in_out>(
     let block_len = algorithm.block_len();
 
     match mode {
-        // TODO: Hopefully support CFB1, and CFB8
-        OperatingMode::CTR | OperatingMode::CFB128 => {}
-        _ => {
+        OperatingMode::CBC => {
             if (in_out.len() % block_len) != 0 {
                 return Err(Unspecified);
             }
         }
+        _ => {}
     }
 
     match mode {
@@ -741,7 +739,7 @@ fn encrypt_aes_ctr_mode(
         SymmetricCipherKey::Aes128 { enc_key, .. } | SymmetricCipherKey::Aes256 { enc_key, .. } => {
             enc_key
         }
-        _ => return Err(Unspecified),
+        _ => unreachable!(),
     };
 
     let mut iv = {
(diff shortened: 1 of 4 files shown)


mariusknaust:
I've tried it with the use case described in #575 and it works as expected 👍🏻 Thanks a lot!

aws-lc-rs/src/cipher.rs
```diff
@@ -737,8 +818,79 @@ fn decrypt_aes_cbc_mode<'in_out>(
...
+        OperatingMode::CFB128 => aes_cfb128_decrypt,
+        _ => {
+            // this indicates a programming error and shouldn't happen
+            return Err(Unspecified);
+        }
+    };
+
+    cfb_decrypt(key, &mut iv, in_out);
+
+    iv.zeroize();
```
justsmth, line 887: NP: The IV is not secret, so no need to zeroize (but I think we do likewise in many other places).

aws-lc-rs/src/cipher.rs
```diff
@@ -632,7 +706,8 @@ fn decrypt<'in_out>(
     let block_len = algorithm.block_len();
 
     match mode {
-        OperatingMode::CTR => {}
+        // TODO: Hopefully support CFB1, and CFB8
+        OperatingMode::CTR | OperatingMode::CFB128 => {}
         _ => {
             if (in_out.len() % block_len) != 0 {
                 return Err(Unspecified);
```
justsmth, line 713: NP: I think it would be better to make CBC mode explicit, and the default case the no-op (or make all modes explicit?):
```
    match mode {
        OperatingMode::CBC => {
            if (in_out.len() % block_len) != 0 {
                return Err(Unspecified);
            }
        }
        _ => {}
    }
```

justsmth (dismissed):
I have a few nit-picks, but no significant concerns.

aws-lc-rs/src/cipher.rs
skmcgrail: It's a pattern followed elsewhere in the existing code, but agree it's technically not considered secret.

justsmth (approved):

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -631,6 +668,61 @@ mod tests {
...
+            let key = UnboundCipherKey::new(&AES_256, &key.clone()).unwrap();
+            StreamingEncryptingKey::cfb128(key).unwrap()
+        };
+        let decrypting_key_creator = |decryption_ctx: DecryptionContext| {
+            let key = UnboundCipherKey::new(&AES_256, &key.clone()).unwrap();
+            StreamingDecryptingKey::cfb128(key, decryption_ctx).unwrap()
+        };
+
+        for i in 13..=21 {
+            for j in 124..=131 {
```
samuel40791765, line 687: Sorry if I've asked this before and I've noticed this is used in the other tests as well. 
What do these pair of numbers mean and should we abstract them into variables?

samuel40791765 (approved):

PR #597, opened 2024-11-15
AES ECB mode support

### LINKED ISSUE #573: Decrypting AES ECB payload
### Security issue notifications

If you discover a potential security issue in AWS-LC for Rust we ask that you notify AWS Security via our
[vulnerability reporting page](http://aws.amazon.com/security/vulnerability-reporting/). Please do **not** create a
public github issue, if in doubt contact AWS security first.

### Problem:

I have incoming payloads that for historical purposes are encrypted using AES in ECB mode (this is the default AES implementation for Java). From what I've seen, the crate uses ecb internally through the aws-lc library, but the only externally exposed implementations are CTR and CBC. Is there any way to expose that algorithm or use a workaround that

### Solution:

A description of the possible solution in terms of AWS-LC for Rust architecture. Highlight and explain any potentially
controversial design decisions taken.

* **Does this change any public APIs?** If yes, explain.
Yes, this would effect the public api for AES. I assume the reason this implementation was left as internal only is due to the lower security compared to AES with CBC.

* **Which algorithm(s) will this impact?**
AES

Either a support for ECB mode, or a workaround for using ECB mode with the existing API.

* **RFC links:** Links to relevant RFC(s)
* **Related Issues:** Link any relevant issues
* **Will the Usage Guide or other documentation need to be updated?**
If the API were to change, it should include strong warnings against using this algorithm.
* **Testing:** How will this change be tested? Call out new integration tests, functional tests, or particularly
  interesting/important unit tests.

### Out of scope:

Is there anything the solution will intentionally NOT address?

skmcgrail:
### Issues:
Resolves #573

### Description of changes: 
* Adds support for ECB mode with pkcs7 padding
* Adds support for ECB mode with pkcs7 padding using the streaming interface
* Adds support for ECB mode without padding (requires input/output to be multiple of block length).

### Call-outs:
* Streaming mode doesn't support ECB without padding as the EVP interface doesn't have a way to toggle off the behavior for this mode.

### DIFF
--- a/aws-lc-rs/src/cipher.rs
+++ b/aws-lc-rs/src/cipher.rs
@@ -230,38 +230,35 @@ pub use streaming::{BufferUpdate, StreamingDecryptingKey, StreamingEncryptingKey
 
 use crate::buffer::Buffer;
 use crate::error::Unspecified;
-use crate::fips::indicator_check;
 use crate::hkdf;
 use crate::hkdf::KeyType;
 use crate::iv::{FixedLength, IV_LEN_128_BIT};
 use crate::ptr::ConstPointer;
 use aws_lc::{
-    AES_cbc_encrypt, AES_cfb128_encrypt, AES_ctr128_encrypt, AES_ecb_encrypt, EVP_aes_128_cbc,
-    EVP_aes_128_cfb128, EVP_aes_128_ctr, EVP_aes_128_ecb, EVP_aes_256_cbc, EVP_aes_256_cfb128,
-    EVP_aes_256_ctr, EVP_aes_256_ecb, AES_DECRYPT, AES_ENCRYPT, AES_KEY, EVP_CIPHER,
+    EVP_aes_128_cbc, EVP_aes_128_cfb128, EVP_aes_128_ctr, EVP_aes_128_ecb, EVP_aes_256_cbc,
+    EVP_aes_256_cfb128, EVP_aes_256_ctr, EVP_aes_256_ecb, EVP_CIPHER,
 };
 use core::fmt::Debug;
 use key::SymmetricCipherKey;
-use zeroize::Zeroize;
 
 /// The number of bytes in an AES 128-bit key
-pub const AES_128_KEY_LEN: usize = 16;
+pub use crate::cipher::aes::AES_128_KEY_LEN;
 
 /// The number of bytes in an AES 256-bit key
-pub const AES_256_KEY_LEN: usize = 32;
+pub use crate::cipher::aes::AES_256_KEY_LEN;
 
 const MAX_CIPHER_KEY_LEN: usize = AES_256_KEY_LEN;
 
 /// The number of bytes for an AES-CBC initialization vector (IV)
-pub const AES_CBC_IV_LEN: usize = 16;
+pub use crate::cipher::aes::AES_CBC_IV_LEN;
 
 /// The number of bytes for an AES-CTR initialization vector (IV)
-pub const AES_CTR_IV_LEN: usize = 16;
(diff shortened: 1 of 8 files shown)


aws-lc-rs/src/cipher.rs
```diff
@@ -873,6 +941,82 @@ fn decrypt_aes_cfb_mode<'in_out>(
...
+    indicator_check!(unsafe {
+        AES_ecb_encrypt(in_out.as_ptr(), in_out.as_mut_ptr(), key, AES_ENCRYPT);
+    });
+}
+
+fn aes_ecb_decrypt(key: &AES_KEY, in_out: &mut [u8]) {
+    indicator_check!(unsafe {
+        AES_ecb_encrypt(in_out.as_ptr(), in_out.as_mut_ptr(), key, AES_DECRYPT);
+    });
+}
```
justsmth, line 1018: Should we migrate calls to [`cipher::aes::encrypt_block_aes`](https://github.com/aws/aws-lc-rs/blob/c2612bf996810d0397b52b6bf706815f36e83332/aws-lc-rs/src/cipher/aes.rs#L25) to call here instead?

aws-lc-rs/src/cipher/padded.rs
```diff
@@ -85,8 +85,17 @@ impl PaddedBlockEncryptingKey {
...
+        Self::new(key, OperatingMode::CBC, PaddingStrategy::PKCS7)
+    }
+
+    /// Constructs a new `PaddedBlockEncryptingKey` cipher with electronic code book (ECB) mode.
+    /// Plaintext data is padded following the PKCS#7 scheme.
+    ///
+    /// # Errors
+    /// * [`Unspecified`]: Returned if there is an error constructing a `PaddedBlockEncryptingKey`.
+    pub fn ecb_pkcs7(key: UnboundCipherKey) -> Result<Self, Unspecified> {
+        Self::new(key, OperatingMode::ECB, PaddingStrategy::PKCS7)
```
justsmth, line 98: Danger message in the documentation here?

aws-lc-rs/src/cipher/padded.rs
```diff
@@ -191,8 +200,22 @@ impl PaddedBlockDecryptingKey {
...
+    ///
+    // # FIPS
+    // Use this function with an `UnboundCipherKey` constructed with one of the following algorithms:
+    // * `AES_128`
+    // * `AES_256`
+    //
+    /// # Errors
+    /// * [`Unspecified`]: Returned if there is an error constructing the `PaddedBlockDecryptingKey`.
+    pub fn ecb_pkcs7(key: UnboundCipherKey) -> Result<Self, Unspecified> {
+        Self::new(key, OperatingMode::ECB, PaddingStrategy::PKCS7)
```
justsmth, line 218: Danger message?

justsmth (approved):

aws-lc-rs/src/cipher/aes.rs
```diff
@@ -3,32 +3,342 @@
...
+    };
+
+    {
+        let mut in_out_iter = in_out.chunks_exact_mut(AES_BLOCK_LEN);
+
+        for block in in_out_iter.by_ref() {
+            aes_ecb_decrypt(key, block);
+        }
+
+        // This is a sanity check hat should not fail. We validate in `decrypt` that in_out.len() % block_len == 0 for
```
samuel40791765, line 252:
```suggestion
        // This is a sanity check that should not fail. We validate in `decrypt` that in_out.len() % block_len == 0 for
```

samuel40791765 (approved):

Issue #599, opened 2024-11-16
ffile-prefix-map argument not quoted


s-mayrh:
### Problem:

I tried to build `trunk` on windows-gnu toolchain using the `rustls` feature flag. This added the `aws-lc-sys` crate which tries to test the C compiler via `cmake`, but fails because the argument `-ffile-prefix-map` is not properly quoted when there are backslashes and spaces in the path.

See the `running: "cmake.exe"` arguments below where the `ffile-prefix-map` argument is correctly quoted (except the space) but `cmake` seems to have removed the quotation when passed to `cc.exe`.  I was about to report this to CMake, but their rules say
> If you are having trouble building a specific third-party project
> that uses CMake, ask for help in that project's forums first.

#### Relevant details
AWS-LC for Rust versions or commit: aws-lc-sys@0.23.0

System information: 
 * CPU architecture: x86-64
 * CPU name: Intel Atom
 * OS: Windows 10
 * CMake: 3.31.0
 * Toolchain: stable-x86_64-pc-windows-gnu with MSYS2

Build log:
```text
[..]
[...]

PR #608, opened 2024-11-20
Handle spaces in source/build paths

justsmth:
#### Addresses:
* #599
#### Related:
* aws/aws-lc#2005

### Description of changes: 
* Fix handling of spaces in the source/build paths.
* Revert recent MSRV violations
* Moved several compiler-version tests to its own workflow.

### Testing:
New tests added.

### DIFF
--- a/aws-lc-sys/builder/cc_builder.rs
+++ b/aws-lc-sys/builder/cc_builder.rs
@@ -188,7 +188,7 @@ impl CcBuilder {
             env::set_var("CFLAGS", cflags);
         }
 
-        if target_os() == "macos" && target_arch() == "x86_64" {
+        if target_os() == "macos" {
             // This compiler error has only been seen on MacOS x86_64:
             // ```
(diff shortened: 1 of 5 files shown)


skmcgrail (approved):

aws-lc-sys/builder/cc_builder.rs
```diff
@@ -188,7 +188,7 @@ impl CcBuilder {
             env::set_var("CFLAGS", cflags);
         }
 
-        if target_os() == "macos" && target_arch() == "x86_64" {
+        if target_os() == "macos" {
             // This compiler error has only been seen on MacOS x86_64:
```
samuel40791765, line 192: Does this comment need updating

justsmth: It's still a true statement, but I think the reason for it is that for MacOS 13 we only use x86-64 runners and not ARM64.

samuel40791765 (approved):

PR #622, opened 2024-12-05
Support for AES 192

justsmth:
### Issues:
Addresses: CryptoAlg-2678

### Description of changes: 
Add support for AES-192:
* aead::AES_192_GCM
* cipher::AES_192: CTR, CBC, ECB and CFB modes

### Testing:
* The AES_192_GCM tests use the NIST KATs
* For cipher, test KAT values obtained from invoking same operation in OpenSSL. 


aws-lc-rs/src/cipher.rs
```diff
@@ -347,10 +355,14 @@ define_cipher_context!(DecryptionContext, EncryptionContext);
...
+    Aes128 = 0,
+
+    // Discriminant explicitly set due to changing implicit discriminant would break semver
+    /// AES 192-bit
+    Aes192 = 2,
 
     /// AES 256-bit
-    Aes256,
+    Aes256 = 1,
 }
```
skmcgrail, line 366: I feel like this add more overhead then just living with the fact that they aren't sorted in-order by AES key size? It's an aesthetic thing that arguably a documentation system could handle.

skmcgrail (approved):

samuel40791765 (approved):
LGTM; double checked that we weren't exposing AES-192 for TLS since there aren't any existing ciphers using it.

Issue #571, opened 2024-10-24
Invalid numeric argument '/Wextra'  with MSVC, how to build with Clang?


theoparis:
### Problem:

I am trying to build facebook's buck2 project with updated dependencies with Clang on Windows, and aws-lc-sys seems to be trying to use msvc cl.exe instead of clang.exe which causes errors.

While it would be great if this msvc issue can be fixed, is there any way to force aws-lc-sys to use `clang.exe` or `clang-cl.exe` directly from LLVM instead of msvc cl.exe?

#### Relevant details
AWS-LC for Rust versions or commit: 92ffba9c29a34b22

System information: for linux, below info can be collected by running `uname -srvmp`
 * CPU architecture: (x86, x86-64, ARMv7...) x86_64
 * CPU name: (Xeon Platinum 8000, AMD EPYC 7000...) Ryzen 5600X
 * OS: (Ubuntu 20.04, Windows Server 2019...) Windows 10

Build log:
 * The log tells compiler and version.
```
  CMake Error at C:/toolchain/share/cmake-3.31/Modules/CMakeTestCCompiler.cmake:67 (message):
    The C compiler

      "C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools/VC/Tools/MSVC/14.41.34120/bin/Hostx64/x64/cl.exe"

    is not able to compile a simple test program.

    It fails with the following output:

      Change Dir: 'C:/Users/theo/Documents/buck2/target/release/build/aws-lc-sys-92ffba9c29a34b22/out/build/CMakeFiles/CMakeScratch/TryCompile-ey9jyx'

      Run Build Command(s): "C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools/MSBuild/Current/Bin/amd64/MSBuild.exe" cmTC_04b65.vcxproj /p:Configuration=Debug /p:Platform=x64 /p:VisualStudioVersion=17.0 /v:n
      MSBuild version 17.11.9+a69bbaaf5 for .NET Framework
      Build started 10/23/2024 11:03:23 PM.

      Project "C:\Users\theo\Documents\buck2\target\release\build\aws-lc-sys-92ffba9c29a34b22\out\build\CMakeFiles\CMakeScratch\TryCompile-ey9jyx\cmTC_04b65.vcxproj" on node 1 (default targets).
      PrepareForBuild:
        Creating directory "cmTC_04b65.dir\Debug\".
[...]

PR #644, opened 2024-12-29
Avoid appending empty path to artifact directory.

jdm:
### Issues:
Related to https://github.com/aws/aws-lc-rs/issues/571#issuecomment-2453515258

The build script for aws-lc-sys appends a library search path to the final linker command line. On Windows environments when building with lld-link.exe, the inclusion of an empty directory at the end of this path can cause a build failure.

The problematic path looks like this: `/LIBPATH:C:\\\\a\\\\servo\\\\servo\\\\target\\release\\build\\aws-lc-sys-d847be10163c7448\\out\\build\\artifacts\\`

The presence of the `\\` at the end of the path causes the subsequent character to be treated as an escaped character, leading to a failure to parse the linker argument that follows this.

### Description of changes:
Since the final component of the path is always empty, we can remove it without any behaviour change.

### Testing:
Verified this change by building Servo on Windows with my branch of aws-lc-sys. Before the change it failed to build; after the change it builds successfully.

### DIFF
--- a/aws-lc-fips-sys/builder/cmake_builder.rs
+++ b/aws-lc-fips-sys/builder/cmake_builder.rs
@@ -60,10 +60,6 @@ fn find_cmake_command() -> Option<OsString> {
     }
 }
 
-fn get_platform_output_path() -> PathBuf {
-    PathBuf::new()
-}
-
 impl CmakeBuilder {
     pub(crate) fn new(
         manifest_dir: PathBuf,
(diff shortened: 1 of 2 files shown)


justsmth:
Thanks for the PR! 🚀

justsmth (approved):

skmcgrail (approved):

PR #653, opened 2025-01-09
Improve Ed25519 doctest; improve rand test

justsmth:
### Description of changes: 
* Update Ed25519 doc-test to use the "generate()` function for key generation instead of `generate_pkcs8v1()`.
* Increase the number of "random" values collected so that mean and variance should always remain within the asserted bounds.

### DIFF
--- a/aws-lc-rs/src/rand.rs
+++ b/aws-lc-rs/src/rand.rs
@@ -172,7 +172,8 @@ mod tests {
 
     #[test]
     fn test_secure_random_fill() {
-        let mut random_array = [0u8; 173];
+        // Collect enough random values so that the assertions below should never fail again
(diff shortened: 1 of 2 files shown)


skmcgrail (approved):

aws-lc-rs/src/rand.rs
```diff
@@ -172,7 +172,8 @@ mod tests {
 
     #[test]
     fn test_secure_random_fill() {
-        let mut random_array = [0u8; 173];
+        // Collect enough random values so that the assertions below should never fail again
+        let mut random_array = [0u8; 1009];
```
WillChilds-Klein, line 176: anything significant about 1009?

WillChilds-Klein (approved):

PR #657, opened 2025-01-13
Support FIPS build w/ CPU Jitter

justsmth:
### Issues:
Addresses: CryptoAlg-2814

### Description of changes: 
* Detects environment variable `AWS_LC_FIPS_SYS_CPU_JITTER_ENTROPY` indicating that AWS-LC should be built to use cpu jitter entropy.
* Adds functions to allow consumers to verify cpu jitter entropy is in use at runtime.

### Call-outs:
* The required `FIPS_is_entropy_cpu_jitter` function is not yet available on the FIPS branch.  So the check currently only verifies that the "cpu_jitter_entropy" configuration was successfully detected by the build scripts.
* This PR depends on: #649

### Testing:
* Added script/CI that verifies that the `fips_cpu_jitter_entropy()` assertion succeeds.

### DIFF
--- a/aws-lc-rs/src/lib.rs
+++ b/aws-lc-rs/src/lib.rs
@@ -233,14 +233,42 @@ pub fn fips_mode() {
 /// Indicates whether the underlying implementation is FIPS.
 ///
 /// # Errors
-/// Return an error if the underlying implementation is not FIPS, otherwise ok
+/// Return an error if the underlying implementation is not FIPS, otherwise Ok.
 pub fn try_fips_mode() -> Result<(), &'static str> {
     init();
-    unsafe {
-        match FIPS_mode() {
-            1 => Ok(()),
-            _ => Err("FIPS mode not enabled!"),
-        }
+    match unsafe { FIPS_mode() } {
+        1 => Ok(()),
+        _ => Err("FIPS mode not enabled!"),
+    }
+}
+
+#[cfg(feature = "fips")]
+/// Panics if the underlying implementation is not using CPU jitter entropy, otherwise it returns.
+///
+/// # Panics
+/// Panics if the underlying implementation is not using CPU jitter entropy.
+pub fn fips_cpu_jitter_entropy() {
+    try_fips_cpu_jitter_entropy().unwrap();
+}
+
+/// Indicates whether the underlying implementation is FIPS.
+///
+/// # Errors
+/// Return an error if the underlying implementation is not using CPU jitter entropy, otherwise Ok.
+pub fn try_fips_cpu_jitter_entropy() -> Result<(), &'static str> {
+    init();
+    // TODO: Delete once FIPS_is_entropy_cpu_jitter() available on FIPS branch
+    // https://github.com/aws/aws-lc/pull/2088
+    #[cfg(feature = "fips")]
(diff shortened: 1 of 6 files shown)


aws-lc-rs/src/lib.rs
```diff
@@ -292,12 +320,17 @@ mod tests {
     #[test]
     fn test_fips() {
         assert!({ crate::try_fips_mode().is_err() });
+        assert!({ crate::try_fips_cpu_jitter_entropy().is_err() });
     }
 
     #[test]
     // FIPS mode is disabled for an ASAN build
```
andrewhop, line 327: Just curious is there some issue the Rust compiler/AWS-LC FIPS/ASAN? ASAN works with the FIPS build of the C library.

scripts/tests/assert_cpu_jitter_entropy.rs
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env -S cargo +nightly -Zscript
```
andrewhop, line 1: Just wondering: why is this a standalone script and not just a normal unit test?

aws-lc-rs/src/lib.rs
```diff
@@ -233,14 +233,42 @@ pub fn fips_mode() {
...
+}
+
+#[cfg(feature = "fips")]
+/// Panics if the underlying implementation is not using CPU jitter entropy, otherwise it returns.
+///
+/// # Panics
+/// Panics if the underlying implementation is not using CPU jitter entropy.
+pub fn fips_cpu_jitter_entropy() {
+    try_fips_cpu_jitter_entropy().unwrap();
+}
```
andrewhop, line 252: Is this normal in Rust to include two functions: one that returns the result and another that unwraps it for the caller?

scripts/tests/assert_cpu_jitter_entropy.rs
justsmth: I wanted to simulate a consumer who depends on aws-lc-rs setting the `AWS_LC_FIPS_SYS_CPU_JITTER_ENTROPY` environment variable then expecting jitter entropy to be available. This script provides that setup in a compact form. (We also have couple of unit tests for coverage.)

aws-lc-rs/src/lib.rs
justsmth: With ASAN enabled [`FIPS_mode` returns 0](https://github.com/aws/aws-lc/blob/94daeb22e78465e53f97417b4c7397dbf2f6b969/crypto/fipsmodule/self_check/fips.c#L24) instead of 1.

aws-lc-rs/src/lib.rs
justsmth: * This change is just following the pattern established w/ `fips_mode()` and `try_fips_mode()`.
* It's fairly typical to name a function "try_xxx" when the operation is fallible. 
* Given that a consumer's compliance status might depend upon FIPS being enabled, having a function that panics the process (and hopefully fails what otherwise would've been a bad/non-compliant deployment) is a convenience.

skmcgrail (approved):

andrewhop (approved):

PR #663, opened 2025-01-20
Implement `from_seed_unchecked` for `Ed25519KeyPair`

### LINKED ISSUE #662: Construct `Ed25519KeyPair` from just a raw private key seed
### Problem:

`Ed25519KeyPair` cannot be constructed from just a raw private key seed. The only option is to encode the key to PKCS#8 and call `from_pkcs8`. There's also `from_seed_and_public_key`, but it requires the caller to already know the public key ahead of time, which in our use case is not.

### Solution:

Implement `from_seed_unchecked`, like `ring` does.

* **RFC links:** Links to relevant RFC(s)
* **Related Issues:** Link any relevant issues
* **Will the Usage Guide or other documentation need to be updated?**
* **Testing:** How will this change be tested? Call out new integration tests, functional tests, or particularly
  interesting/important unit tests.

### Out of scope:

Is there anything the solution will intentionally NOT address?

paolobarbolini:
### Issues:
Resolves #662

### Description of changes: 

`Ed25519KeyPair` currently cannot be constructed from a raw private key seed. This copies the `ring` `from_seed_unchecked` API which makes it possible to do so.

### Call-outs:

I've moved most of the `from_seed_and_public_key` implementation to `from_seed_unchecked` and had `from_seed_and_public_key` call it. I'm not sure if it's ok for `from_seed_and_public_key` to do more work before rejecting the seed, in case the provided public key doesn't match the one derived from the seed.

### Testing:

I've added a test for the new function together with the pre-existing test for `from_seed_and_public_key`. I've also tested my application and confirmed the `from_seed_unchecked` API works the same as the `from_seed_unchecked` unchecked function from `ring`.

### DIFF
--- a/aws-lc-rs/src/ed25519.rs
+++ b/aws-lc-rs/src/ed25519.rs
@@ -360,7 +360,13 @@ impl Ed25519KeyPair {
 
     /// Constructs an Ed25519 key pair from the private key seed `seed`.
     ///
-    /// It is recommended to use `Ed25519KeyPair::from_pkcs8()` instead.
+    /// It is recommended to use `Ed25519KeyPair::from_pkcs8()` instead. If the public key is
+    /// available, prefer to use `Ed25519KeyPair::from_seed_and_public_key()` as it will verify
+    /// the validity of the key pair.
+    ///
+    /// CAUTION: Both an Ed25519 seed and its public key are 32-bytes. If the bytes of a public key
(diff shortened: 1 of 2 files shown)


justsmth (approved):
Thanks for the PR!

skmcgrail (approved):

PR #669, opened 2025-01-22
Cleanup EVP_PKEY serialization

justsmth:
### Description of changes: 
* Remove duplicated logic around usage of `EVP_marshal_private_key`, `EVP_parse_private_key`, `EVP_marshal_public_key` and `EVP_parse_public_key`.
* Change all usage of `EC_GROUP*` to `const`.
* Consolidate EC SEC1 and RFC 5915 parsing/marshaling into `ec:encoding` module.

### Call-outs:
N/A


aws-lc-rs/src/ec/key_pair.rs
```diff
@@ -185,7 +183,7 @@ impl EcdsaKeyPair {
         alg: &'static EcdsaSigningAlgorithm,
         private_key: &[u8],
     ) -> Result<Self, KeyRejected> {
-        let evp_pkey = encoding::unmarshal_der_to_private_key(private_key, alg.id.nid())?;
+        let evp_pkey = ec::encoding::rfc5915::parse_rfc5915_private_key(private_key, alg.id.nid())?;
```
skmcgrail, line 186: Maybe remove the slight name stuttering, `rfc5915` is already in the module name, so I would remove it from the function name.

aws-lc-rs/src/ec/key_pair.rs
```diff
@@ -303,10 +301,7 @@ impl AsBigEndian<EcPrivateKeyBin<'static>> for PrivateKey<'_> {
     /// # Errors
     /// `error::Unspecified` if serialization failed.
     fn as_be_bytes(&self) -> Result<EcPrivateKeyBin<'static>, Unspecified> {
-        let buffer = encoding::marshal_private_key_to_buffer(
-            self.0.algorithm.id.private_key_size(),
-            &self.0.evp_pkey.as_const(),
-        )?;
+        let buffer = encoding::sec1::marshal_sec1_private_key(&self.0.evp_pkey)?;
```
skmcgrail, line 304: Maybe same thing here for `sec1`?

aws-lc-rs/src/evp_pkey.rs
```diff
@@ -158,6 +159,79 @@ impl LcPtr<EVP_PKEY> {
...
+    // pub(crate) fn marshal_raw_public_key(&self) -> Result<Vec<u8>, Unspecified> {
+    //     let mut size = 0;
+    //     if 1 != unsafe { EVP_PKEY_get_raw_public_key(*self.as_const(), null_mut(), &mut size) } {
+    //         return Err(Unspecified);
+    //     }
+    //     let mut buffer = vec![0u8; size];
+    //     let buffer_size = self.marshal_raw_public_to_buffer(&mut buffer)?;
+    //     debug_assert_eq!(buffer_size, size);
+    //     Ok(buffer)
+    // }
```
skmcgrail, line 197: Accidentally left in or intentional?

justsmth: It was intentional as my current draft of ML-DSA logic is using it, but I'm not it will be needed.  I'll remove this until we know it's needed.

aws-lc-rs/src/ec/key_pair.rs
justsmth: I'm conflicted about removing the "rfc5915" from the function name as I like that it helps prevent future refactorings from potentially confusing two different functions both named `parse_private_key`. In the latest commit, I did some cleanup to remove the redundant use of "rfc5915" name in the logic here. Let's discuss this more on the next commit.

skmcgrail (dismissed):

skmcgrail (approved):

aws-lc-rs/src/evp_pkey.rs
```diff
@@ -69,53 +57,99 @@ impl LcPtr<EVP_PKEY> {
...
+        let mut cbs = cbs::build_CBS(bytes);
+        // Also checks the validity of the key
+        let evp_pkey = LcPtr::new(unsafe { EVP_parse_private_key(&mut cbs) })
+            .map_err(|()| KeyRejected::invalid_encoding())?;
+        evp_pkey
+            .id()
+            .eq(&evp_pkey_type)
+            .then_some(evp_pkey)
+            .ok_or(KeyRejected::wrong_algorithm())
     }
```
samuel40791765, line 153: Should we consider following the same pattern as `mod sec` and `mod rfc5915` and move these functions to another `mod rfc5208` in `encoding.rs`? It seems like they belong in the same realm at first glance, but I do see how this makes things handy for `EVP_PKEY`.

samuel40791765 (approved):
Just a minor non-blocking comment about placing the encoding functions in the same file/module, but  awesome change overall.

Issue #658, opened 2025-01-14
Building on alpine


ulrichard:
### Problem:

After upgrading some dependencies, aws-lc-rs  is pulled in as a sub-dependency, and I am now getting the following error messages:

```
   Compiling aws-lc-sys v0.24.1
The following warnings were emitted during compilation:

warning: aws-lc-sys@0.24.1: Building with: CMake
warning: aws-lc-sys@0.24.1: Symbol Prefix: Some("aws_lc_0_24_1")
warning: aws-lc-sys@0.24.1: CMAKE environment variable set: cmake
warning: aws-lc-sys@0.24.1: Generating bindings - external bindgen. Platform: x86_64-alpine-linux-musl

error: failed to run custom build command for `aws-lc-sys v0.24.1`

Caused by:
  process didn't exit successfully: `/home/satoshi/target/debug/build/aws-lc-sys-b95fd8f0c5480cc1/build-script-main` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=AWS_LC_SYS_NO_PREFIX
  cargo:rerun-if-env-changed=AWS_LC_SYS_PREGENERATING_BINDINGS
  cargo:rerun-if-env-changed=AWS_LC_SYS_EXTERNAL_BINDGEN
  cargo:rerun-if-env-changed=AWS_LC_SYS_NO_ASM
  cargo:rerun-if-env-changed=AWS_LC_SYS_CFLAGS
  cargo:rerun-if-env-changed=AWS_LC_SYS_PREBUILT_NASM
  cargo:rerun-if-env-changed=AWS_LC_SYS_C_STD
  cargo:rerun-if-env-changed=AWS_LC_SYS_CMAKE_BUILDER
  cargo:rerun-if-env-changed=AWS_LC_SYS_STATIC
  cargo:rerun-if-env-changed=CMAKE
  cargo:warning=Building with: CMake
  cargo:warning=Symbol Prefix: Some("aws_lc_0_24_1")
  cargo:rerun-if-env-changed=CMAKE
  cargo:warning=CMAKE environment variable set: cmake
  cargo:warning=Generating bindings - external bindgen. Platform: x86_64-alpine-linux-musl

  --- stderr
  Consider installing the bindgen-cli: `cargo install --force --locked bindgen-cli`
  See our User Guide for more information about bindgen:https://aws.github.io/aws-lc-rs/index.html
  Failure invoking external bindgen! External bindgen command failed.
[...]

PR #675, opened 2025-01-29
Improve support for alpine linux

justsmth:
### Issues:
Addresses: #658

### Description of changes: 
* Build scripts now map the `???-alpine-linux-musl` targets to the `???-unknown-linux-musl` targets that we support.
* Adds CI and docker image for testing builds on Alpine Linux.

### Testing:
CI Job added,

### DIFF
--- a/.github/docker_images/alpine-3.20/Dockerfile
+++ b/.github/docker_images/alpine-3.20/Dockerfile
@@ -1,5 +1,9 @@
 FROM alpine:3.20
 
+VOLUME ["/aws_lc_rs"]
+
+WORKDIR /
+
 RUN apk add \
 	alpine-sdk \
 	bash \
@@ -28,3 +32,8 @@ ENV CARGO_HTTP_MULTIPLEXING=false
 #    cargo install --locked bindgen-cli && \
 #    rustup component add rustfmt clippy && \
 #    rm ./rustup.sh
+
+COPY aws_lc_rs_build.sh /
+COPY entry.sh /
+
+ENTRYPOINT ["/entry.sh"]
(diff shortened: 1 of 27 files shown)


aws-lc-sys/builder/cc_builder.rs
```diff
@@ -394,12 +394,12 @@ impl crate::Builder for CcBuilder {
...
 
-        if PlatformConfig::default_for(&target()).is_none() {
-            return Err(format!("Platform not supported: {}", target()));
+        if PlatformConfig::default_for(&effective_target()).is_none() {
+            return Err(format!("Platform not supported: {}", effective_target()));
         }
 
         if Some(true) == env_var_to_bool("CARGO_FEATURE_SSL") {
-            return Err(format!("libssl not supported: {}", target()));
+            return Err(format!("cc_builder for libssl not supported"));
```
smittals2, line 402: Do we not want to print out the target in the err message here anymore?

.github/docker_images/alpine-3.20/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+	build-base \
+	busybox-suid \
+	clang-dev \
+	cargo \
+	curl \
+	cmake \
+	openssl-dev
+
+ARG UID
+RUN adduser -u $UID -S -s /bin/sh -G abuild satoshi
```
smittals2, line 20: satoshi nice

smittals2 (dismissed):

.github/docker_images/alpine-3.20/Dockerfile
justsmth: 🤣

aws-lc-sys/builder/cc_builder.rs
justsmth: In this case, the target doesn't matter. I suspect I had copied the error directly above this and left the target in the error message, even though it's not relevant.

smittals2 (approved):

ulrichard:
I'm happy to report that this PR solves the problem I ran into with #658
Thanks a lot.

skmcgrail (approved):

Issue #569, opened 2024-10-21
FIPS fails to build with Clang 19


jkalez:
### Problem:

`cargo check` fails when using clang 19 for FIPS builds. AFAICT it's because AWS-LC is pinned to a specific commit ([ec94d74](https://github.com/aws/aws-lc/tree/ec94d74a19b5a0aa738b436a95bb06ff87fc7ba9)), which has a cast operation in it that newer versions of clang rejects. I did try upgrading to [AWS-LC-FIPS-2.0.17](https://github.com/aws/aws-lc/releases/tag/AWS-LC-FIPS-2.0.17), but it appears the issue is still there and `cargo check` still fails.

#### Relevant details
AWS-LC for Rust versions or commit: (6b1bce0...)

System information: for linux, below info can be collected by running `uname -srvmp`
 * CPU architecture: x86-64
 * CPU name: Xeon CascadeLake
 * OS: Ubuntu 24.10

Build log:
 * The log tells compiler and version.
```text
root@504baee641ea:~/aws-lc-rs# cargo check -p aws-lc-rs --features fips
   Compiling aws-lc-fips-sys v0.12.13 (/root/aws-lc-rs/aws-lc-fips-sys)
   Compiling aws-lc-sys v0.22.0 (/root/aws-lc-rs/aws-lc-sys)
The following warnings were emitted during compilation:

warning: aws-lc-fips-sys@0.12.13: Building with: CMake
warning: aws-lc-fips-sys@0.12.13: Symbol Prefix: Some("aws_lc_fips_0_12_13")

error: failed to run custom build command for `aws-lc-fips-sys v0.12.13 (/root/aws-lc-rs/aws-lc-fips-sys)`

Caused by:
  process didn't exit successfully: `/root/aws-lc-rs/target/debug/build/aws-lc-fips-sys-4f564eb018bf155b/build-script-main` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=AWS_LC_FIPS_SYS_NO_PREFIX
  cargo:rerun-if-env-changed=AWS_LC_FIPS_SYS_PREGENERATING_BINDINGS
  cargo:rerun-if-env-changed=AWS_LC_FIPS_SYS_EXTERNAL_BINDGEN
  cargo:rerun-if-env-changed=AWS_LC_FIPS_SYS_NO_ASM
  cargo:rustc-cfg=x86_64_unknown_linux_gnu
[...]

PR #678, opened 2025-01-31
Avoid fips MacOS failure due to -Woverriding-option

justsmth:
### Context
Some users experiencing build failures relating to the following error:
```
  clang: error: overriding '-mmacosx-version-min=14.7' option with '--target=arm64-apple-macosx15.1' [-Werror,-Woverriding-option]
  make[2]: *** [aws-lc/crypto/fipsmodule/CMakeFiles/bcm_library.dir/bcm.c.o] Error 1
```

### Description of changes: 
* This change applies the same fix for `-Woverriding-option` as we had done for `-Woverriding-t-option`.

### Call-outs:
N/A

### Testing:
A user experiencing this problem confirmed this change resolved it.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

PR #682, opened 2025-02-03
aws-lc-fips-sys: Default to clang compiler

michelhe:
### Issues:
Addresses #569 

### Description of changes: 
The underlying aws-lc library is broken when built with FIPS support on GCC-14, which comes by default with many Linux distributions.

When this crate is used as a sub-dependency, the users' only workaround is to force all C/C++ code on other crates to also be compiled with clang by setting the CC/CXX environment variables.

This change circumvents the issue by setting the CC/CXX environment variables in the crate build script.

This is a temporary workaround until the underlying issue is resolved: https://github.com/aws/aws-lc/issues/2010

### Testing:
 Tested the builds with and without setting CC and CXX.

### DIFF
--- a/aws-lc-fips-sys/builder/cmake_builder.rs
+++ b/aws-lc-fips-sys/builder/cmake_builder.rs
@@ -93,6 +93,16 @@ impl CmakeBuilder {
     const GOCACHE_DIR_NAME: &'static str = "go-cache";
     #[allow(clippy::too_many_lines)]
     fn prepare_cmake_build(&self) -> cmake::Config {
+        // Set the compiler to clang if CC/CXX are not set
+        // TODO: this is a workaround until aws-lc fixes build on GGC-14
+        // See https://github.com/aws/aws-lc/issues/201
+        if env::var("CC").is_err() {
+            env::set_var("CC", "clang");
+        }
+        if env::var("CXX").is_err() {
+            env::set_var("CXX", "clang++");
+        }
+
         env::set_var(
             "GOCACHE",
             self.out_dir.join(Self::GOCACHE_DIR_NAME).as_os_str(),


aws-lc-fips-sys/builder/cmake_builder.rs
```diff
@@ -93,6 +93,16 @@ impl CmakeBuilder {
     const GOCACHE_DIR_NAME: &'static str = "go-cache";
     #[allow(clippy::too_many_lines)]
     fn prepare_cmake_build(&self) -> cmake::Config {
+        // Set the compiler to clang if CC/CXX are not set
+        // TODO: this is a workaround until aws-lc fixes build on GGC-14
+        // See https://github.com/aws/aws-lc/issues/201
+        if env::var("CC").is_err() {
+            env::set_var("CC", "clang");
+        }
```
justsmth, line 101: If clang is not available on the system, then this would likely cause a build failure. 

I think switching the compiler on the user is something that should only be done when (1) we know clang is available in the environment, and (2) the switch is necessary to avoid an inevitable build failure. Even in that case, quietly switching which compiler is used bothers me a little. Would it be better to instead instruct the user to set the environment variable themselves?

michelhe: The problem with setting the env vars globally is that it imposes change to the build of other crates that also use the 'cc' crate.

I.e what if a crate "foo" has a bug when building with clang and now I'm deadlocked and cannot build both aws-lc-sys-rs and foo with the same global compiler env var?


I agree that failing the build if clang does not exist may not be nice for the user.

I can't find a way to tell the cc crate to "prefer clang if installed and env vars are not set"

justsmth: I added a workaround to address this on [another PR](https://github.com/aws/aws-lc-rs/pull/678): https://github.com/aws/aws-lc-rs/pull/678/commits/81c9f00d5afa4c30e634c500eb6556254e51311c

michelhe: Ah, wonderful, we can close this one and I'll patiently await for it to be released :)

michelhe:
Workaround (special env vars to affect compiler) merged in #678 , this is no longer needed

PR #690, opened 2025-02-06
Unstable API for ML-DSA

justsmth:
### Description of changes: 
* Adds unstable API for support of MLDSA.

### Call-outs:
* For the signature operation, I decided to have the caller provide the mutable buffer.  The required buffer length is provided by `PqdsaSigningAlgorithm::signature_len()`. This is similar to [RSA's signature function](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/rsa/struct.KeyPair.html#method.sign).
* `PqdsaPrivateKey` object can serialize to either PKCS8 or raw bytes.
* `PqdsaPublicKey` object can serialize to X.509 or raw bytes.

### Testing:
New tests added.  


aws-lc-rs/src/encoding.rs
```diff
@@ -82,3 +84,12 @@ pub trait AsBigEndian<T> {
     /// Returns Unspecified if serialization fails.
     fn as_be_bytes(&self) -> Result<T, crate::error::Unspecified>;
 }
+
+/// Trait for values that can be serialized into a raw format
+pub trait AsRawBytes<T> {
```
skmcgrail, line 89: I seem to recall for ML-KEM we did this with the [`AsBigEndian`](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/encoding/trait.AsBigEndian.html) trait?

justsmth: Since the value encoded for MLDSA is not technically "big endian" (which seems most apt to integer encodings), the use of the `AsBigEndian` trait for this seemed wrong. I thought an `AsRawBytes` trait would be most appropriate and likely to be re-used in the future.

aws-lc-rs/src/pqdsa/key_pair.rs
```diff
@@ -0,0 +1,372 @@
...
+            evp_pkey,
+            pubkey,
+        })
+    }
+
+    /// Constructs a key pair from a raw private key.
+    ///
+    /// # Errors
+    /// Returns `Unspecified` if the key is not valid for the specified signing algorithm.
+    pub fn from_raw_private_key(
```
skmcgrail, line 124: I'm confused. What's the purpose of having both `from_raw_private_key` and `from_raw_seed`? Aren't they both the same? They both are just calling the same `EVP_PKEY_pqdsa_new_raw_private_key`. The documentation on seed says that the seeds are the same length, but then `parse_raw_private_key` implementation has an alternative length there. Is one of these functions implemented incorrectly?

aws-lc-rs/src/pqdsa/key_pair.rs
```diff
@@ -0,0 +1,372 @@
...
+        Ok(Pkcs8V1Der::new(
+            self.0
+                .evp_pkey
+                .marshal_rfc5208_private_key(pkcs8::Version::V1)?,
+        ))
+    }
+}
+
+impl AsRawBytes<PqdsaPrivateKeyRaw<'static>> for PqdsaPrivateKey<'_> {
+    fn as_raw_bytes(&self) -> Result<PqdsaPrivateKeyRaw<'static>, Unspecified> {
```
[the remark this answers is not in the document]
justsmth: It's currently used for `PqdsaPrivateKey`, but I'd like for the `AsRawBytes` trait to be potentially usable for values that are not necessarily "keys" (like "seeds", "IVs", etc.).

aws-lc-rs/src/pqdsa/key_pair.rs
```diff
@@ -0,0 +1,372 @@
...
+    ///
+    /// It is recommended to use `PqdsaKeyPair::from_pkcs8()` instead. If the public key is
+    /// available, prefer to use `PqdsaKeyPair::from_seed_and_public_key()` as it will verify
+    /// the validity of the key pair.
+    ///
+    /// CAUTION: MLDSA-44, MLDSA-65 and MLDSA-87 all use a 32-byte seed.
+    ///
+    /// # Errors
+    /// Returns `Unspecified` if the key is not valid for the specified signing algorithm.
+    pub fn from_raw_seed(
```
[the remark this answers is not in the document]
justsmth: I removed this function in the latest commit.

aws-lc-rs/src/pqdsa/key_pair.rs
justsmth: `from_raw_seed` is removed in the latest commit.

aws-lc-rs/src/pqdsa/key_pair.rs
```diff
@@ -0,0 +1,284 @@
...
+        let evp_pkey = evp_key_pqdsa_generate(algorithm.0.id.nid())?;
+        let pubkey = PublicKey::from_private_evp_pkey(&evp_pkey)?;
+        Ok(Self {
+            algorithm,
+            evp_pkey,
+            pubkey,
+        })
+    }
+
+    /// Parses a PKCS#8 key from the specified bytes.
```
samuel40791765, line 82: Minor nit: Should we align the documentation for `from_pkcs8` and `from_raw_private_key` to be more similar? "Constructs a key pair from ... " is a great accurate description.

aws-lc-rs/src/pqdsa/key_pair.rs
```diff
@@ -0,0 +1,284 @@
...
+    ///
+    /// # Errors
+    /// Returns `Unspecified` if serialization fails.
+    pub fn to_pkcs8(&self) -> Result<Document, Unspecified> {
+        Ok(Document::new(
+            self.evp_pkey.marshal_rfc5208_private_key(Version::V1)?,
+        ))
+    }
+
+    /// Uses this key to sign the message provided.
```
samuel40791765, line 128: Nit: Could be more clearer on what's being returned and where the signature is written to on success

aws-lc-rs/tests/pqdsa_test.rs
```diff
@@ -0,0 +1,85 @@
...
+            let public_key = test_case.consume_bytes("PUBLIC");
+            let message = test_case.consume_bytes("MESSAGE");
+            let signature = test_case.consume_bytes("SIGNATURE");
+            let context = test_case.consume_bytes("CONTEXT");
+            let expected_result = test_case.consume_bool("RESULT");
+
+            if !context.is_empty() {
+                // skip test
+                return Ok(());
+            }
```
samuel40791765, line 42: What's the reason for skipping tests without empty `context`s? Would we unintentionally skipping test cases if we imported files with populated `context`s?

aws-lc-rs/src/pqdsa.rs
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+
+    pub(crate) const fn signature_size_bytes(&self) -> usize {
+        match self {
+            Self::MLDSA_44 => 2420,
+            Self::MLDSA_65 => 3309,
+            Self::MLDSA_87 => 4627,
+        }
+    }
```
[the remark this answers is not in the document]
justsmth: It's likely that AWS-LC will eventually need to expose these constants for consumers, but I don't think it's necessary for integration into `aws-lc-rs`. Once the constants are exposed and available in the `aws-lc-sys` bindings, we should switch the code here over to use them.

aws-lc-rs/tests/pqdsa_test.rs
justsmth: Many of the ML-DSA KATs require a specific context to be set. This is useful for internally testing the AWS-LC implementation. However, AWS-LC doesn't (yet) expose the APIs necessary for consumers (like `aws-lc-rs`) to specify a context and then match our answer. So in `aws-lc-rs`, we can only compare our results against KATs where the context is empty.

Originally I kept all of the KATs in our data files, but I've since removed then. I forgot to also remove this condition from the test.

samuel40791765 (approved):

aws-lc-rs/src/unstable/kem.rs
```diff
@@ -10,6 +10,7 @@ pub use crate::kem::{ML_KEM_1024, ML_KEM_512, ML_KEM_768};
 
 #[non_exhaustive]
 #[derive(Clone, Copy, Debug, PartialEq)]
+/// TODO: Remove
```
skmcgrail, line 13: I assume you don't want this in the public documentation on docs.rs as a TODO?

skmcgrail (approved):

PR #692, opened 2025-02-10
Fix cmake build type when opt_level is "s"

### LINKED ISSUE #691: Building for Windows (x86_64-pc-windows-msvc) with opt-level="s" causes link errors
### Problem:

I am building a rust application that utilizes `aws-lc-rs` and `tonic` in Windows.
My application is set to compile with static crt.

I tried to optimize the binary size by changing opt-size to "s", but received this link error:

Image

I was able to reproduce it by building `aws-lc-rs` with `opt-level="s"` and env `RUSTFLAGS="-C target-feature=+crt-static"`.

#### Relevant details
AWS-LC for Rust versions or commit: current master (5af71cc)

I suspect the root cause relies in this logic that chooses the build type based on the opt-level. In MSVC, this directly affects the chosen C runtime, which probably causes the link error above.

elad-solomon:
### Issues:
Resolves #691

### Description of changes: 
Solves the link issue by removing the code that selected the build type. 
This is redundant since cmake-rs does it as well, and in opt-level="s" appearently it chooses a different build type (MinRelSize vs RelWithDebInfo). 

This caused a mismatch in CRT and failed the link.

The logic in cmake-rs is seems more comprehensive and correct, so it's best to let it do its job for us:
https://github.com/rust-lang/cmake-rs/blob/fd56c5a6b4ecda8815c863eb5b12d7b3f0391197/src/lib.rs#L146C1-L147C1

### DIFF
--- a/aws-lc-fips-sys/builder/cmake_builder.rs
+++ b/aws-lc-fips-sys/builder/cmake_builder.rs
@@ -122,33 +122,28 @@ impl CmakeBuilder {
 
         let cc_build = cc::Build::new();
         let opt_level = cargo_env("OPT_LEVEL");
-        if opt_level.ne("0") {
-            if opt_level.eq("1") || opt_level.eq("2") {
-                cmake_cfg.define("CMAKE_BUILD_TYPE", "relwithdebinfo");
-            } else {
-                cmake_cfg.define("CMAKE_BUILD_TYPE", "release");
-                // TODO: Due to the nature of the FIPS build (e.g., its dynamic generation of
-                // assembly files and its custom compilation commands within CMake), not all
-                // source paths are stripped from the resulting binary.
-                emit_warning(
-                    "NOTICE: Build environment source paths might be visible in release binary.",
-                );
-                let parent_dir = self.manifest_dir.parent();
-                if parent_dir.is_some() && (target_family() == "unix" || target_env() == "gnu") {
-                    let parent_dir = parent_dir.unwrap();
+        if !["0", "1", "2"].contains(&opt_level.as_str()) {
+            // TODO: Due to the nature of the FIPS build (e.g., its dynamic generation of
+            // assembly files and its custom compilation commands within CMake), not all
+            // source paths are stripped from the resulting binary.
+            emit_warning(
+                "NOTICE: Build environment source paths might be visible in release binary.",
+            );
+            let parent_dir = self.manifest_dir.parent();
+            if parent_dir.is_some() && (target_family() == "unix" || target_env() == "gnu") {
+                let parent_dir = parent_dir.unwrap();
 
-                    let flag = format!("\"-ffile-prefix-map={}=\"", parent_dir.display());
+                let flag = format!("\"-ffile-prefix-map={}=\"", parent_dir.display());
+                if let Ok(true) = cc_build.is_flag_supported(&flag) {
+                    emit_warning(&format!("Using flag: {}", &flag));
+                    cmake_cfg.asmflag(&flag);
+                    cmake_cfg.cflag(&flag);
+                } else {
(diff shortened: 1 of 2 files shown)


aws-lc-sys/builder/cmake_builder.rs
```diff
@@ -108,17 +108,6 @@ impl CmakeBuilder {
...
-        if opt_level.ne("0") {
-            if opt_level.eq("1") || opt_level.eq("2") {
-                cmake_cfg.define("CMAKE_BUILD_TYPE", "relwithdebinfo");
-            } else {
-                cmake_cfg.define("CMAKE_BUILD_TYPE", "release");
-            }
-        } else {
-            cmake_cfg.define("CMAKE_BUILD_TYPE", "debug");
-        }
-
```
justsmth, line 121: Instead of removing the logic that determines the `CMAKE_BUILD_TYPE`, it would be better to update it to properly handle when [`OPT_LEVEL="s"`](https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level). I think [it should map to `CMAKE_BUILD_TYPE="MinSizeRel"`](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html).

aws-lc-fips-sys/builder/cmake_builder.rs
```diff
@@ -122,33 +122,28 @@ impl CmakeBuilder {
...
+            if parent_dir.is_some() && (target_family() == "unix" || target_env() == "gnu") {
+                let parent_dir = parent_dir.unwrap();
+
+                let flag = format!("\"-ffile-prefix-map={}=\"", parent_dir.display());
+                if let Ok(true) = cc_build.is_flag_supported(&flag) {
+                    emit_warning(&format!("Using flag: {}", &flag));
+                    cmake_cfg.asmflag(&flag);
+                    cmake_cfg.cflag(&flag);
+                } else {
+                    let flag = format!("\"-fdebug-prefix-map={}=\"", parent_dir.display());
```
justsmth, line 142: The FIPS build on Windows doesn't support static builds, so changes to this crate aren't needed.

justsmth:
I put up a commit on my fork that I think addresses your request in a more direct way: https://github.com/aws/aws-lc-rs/commit/852baac3d885e5f0e7d7a4d44a343cbe245cf7ea

If that change works for you, feel free to use that on this PR or I can post it on a separate PR.

aws-lc-sys/builder/cmake_builder.rs
elad-solomon: Thanks, that works as well

I thought it's better to remove it to be less prone to errors.
notice my note that `cmake-rs` already performs this logic, and in a more comprehensive way - https://github.com/rust-lang/cmake-rs/blob/fd56c5a6b4ecda8815c863eb5b12d7b3f0391197/src/lib.rs#L146C1-L147C1.

Don't you think it's better not to duplicate this logic?

aws-lc-fips-sys/builder/cmake_builder.rs
elad-solomon: it actually does :)

elad-solomon:
thanks @justsmth , this works as well

are you planning to create a PR from this commit to take it to upstream soon?

justsmth:
Hello @elad-solomon!

I posted that commit I suggested as a separate PR (#694), so I'll close this PR. (Feel free to comment there if needed.) I'll leave the issue (#691) open until we've got this change published in a release.  Thanks for posting a PR!

elad-solomon:
Awesome, thanks @justsmth !
looking forward for your PR #694  to be merged so I can update my application deps 🤞

PR #693, opened 2025-02-10
Ensure ConstPointers don't outlive parent

justsmth:
### Description of changes: 
* Use lifetimes to prevent a "child" object being used after its "parent" is out of scope.
* Move most non-mutating functionality from `LcPtr<EVP_PKEY>` to `ConstPointer<'_, EVP_PKEY>`

### DIFF
--- a/aws-lc-rs/src/ec.rs
+++ b/aws-lc-rs/src/ec.rs
@@ -5,8 +5,6 @@
 
 use crate::ec::signature::AlgorithmID;
 // TODO: Uncomment when MSRV >= 1.64
-use std::os::raw::c_int;
-
 #[cfg(feature = "fips")]
 use crate::aws_lc::EC_KEY_check_fips;
 #[cfg(not(feature = "fips"))]
@@ -23,6 +21,8 @@ use crate::error::{KeyRejected, Unspecified};
 use crate::fips::indicator_check;
 use crate::ptr::{ConstPointer, LcPtr};
 use crate::signature::Signature;
+use std::os::raw::c_int;
+use std::ptr::null;
 
 pub(crate) mod encoding;
 pub(crate) mod key_pair;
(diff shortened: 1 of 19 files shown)


aws-lc-rs/src/evp_pkey.rs
```diff
@@ -231,13 +182,55 @@ impl LcPtr<EVP_PKEY> {
...
+            .then_some(evp_pkey)
+            .ok_or(KeyRejected::wrong_algorithm())
+    }
+
+    #[allow(non_snake_case)]
+    pub(crate) fn create_EVP_PKEY_CTX(&self) -> Result<LcPtr<EVP_PKEY_CTX>, ()> {
+        // The only modification made by EVP_PKEY_CTX_new to `priv_key` is to increment its
+        // refcount. The modification is made while holding a global lock:
+        // https://github.com/aws/aws-lc/blob/61503f7fe72457e12d3446853a5452d175560c49/crypto/refcount_lock.c#L29
+        LcPtr::new(unsafe { EVP_PKEY_CTX_new(*self.as_mut_unsafe(), null_mut()) })
```
skmcgrail, line 232: Technically the returned context here has a lifetime bound to the `LcPtr<EVP_PKEY>` here. But you could technically make this `'static`, as the context ups the reference count on the `EVP_PKEY`. Not sure if you want to model that in this way or not.

skmcgrail: Oh disregard `LcPtr` doesn't have a lifetime on it! 👍

aws-lc-rs/src/ec.rs
```diff
@@ -5,8 +5,6 @@
 
 use crate::ec::signature::AlgorithmID;
 // TODO: Uncomment when MSRV >= 1.64
```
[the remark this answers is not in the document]
justsmth: I believe this comment related to a commented out `use core::ffi::c_int;`, which is not [available until 1.64](https://doc.rust-lang.org/beta/core/ffi/type.c_int.html).

I'll add back the commented out `use`.  When we bump MSRV (hopefully soon), we can migrate our C-types to those in the `core::ffi` module.

andrewhop (approved):

Issue #696, opened 2025-02-14
`aws-lc-sys` fails to compile for `x86_64-pc-windows-gnu` host


ognevny:
### Problem:

while working at https://github.com/msys2/MINGW-packages/pull/23391 I've got compile errors with missing boringssl includes. MSYS2 doesn't provide boringssl package. I'm sure that submodules are fetched properly and required build dependencies are installed. anyway generated headers are missed in `CARGO_TARGET_DIR`

#### Relevant details
AWS-LC for Rust versions or commit: aws-lc-sys v0.25.0

System information: can be found in github actions run below

Build log:
 https://github.com/msys2/MINGW-packages/actions/runs/13323638143/job/37212571202#step:11:1788

mati865:
In `CMakeCache.txt` there are these lines:
```
//No help, variable specified on the command line.
CMAKE_ASM_FLAGS:UNINITIALIZED= -ffunction-sections -fdata-sections -m64 --target=x86_64-pc-windows-gnu -ffunction-sections -fdata-sections -gdwarf-2 -fno-omit-frame-pointer -m64 --target=x86_64-pc-windows-gnu -std=c11 -I H:\msys64\tmp\aws-lc-rs\aws-lc-sys\generated-include -I H:\msys64\tmp\aws-lc-rs\aws-lc-sys\include -I H:\msys64\tmp\aws-lc-rs\aws-lc-sys\aws-lc\include -I H:\msys64\tmp\aws-lc-rs\aws-lc-sys\aws-lc\third_party\s2n-bignum\include -Wall -Wextra -Wno-unused-parameter -pthread -D_XOPEN_SOURCE=700 -DBORINGSSL_IMPLEMENTATION=1 -DBORINGSSL_PREFIX=aws_lc_0_26_0

//Flags used by the ASM_NASM compiler during all build types.
CMAKE_ASM_NASM_FLAGS:STRING=
```
I suppose instead of `CMAKE_ASM_FLAGS` the flags should go to `CMAKE_ASM_NASM_FLAGS`? Although I'm not that experienced with CMake, so I could be wrong here.

ognevny:
> In `CMakeCache.txt` there are these lines:

where is this file located?

ognevny:
I've workarounded it by passing `export CFLAGS="$CFLAGS -I$(cygpath -am ${srcdir}/aws-lc-rs/aws-lc-sys/aws-lc/include)"` manually before build

ognevny:
[...]

PR #702, opened 2025-02-19
Build/Test for x86_64-pc-windows-gnullvm

justsmth:
### Issues:
Addresses: #696

### Description of changes: 
* Support target: `x86_64-pc-windows-gnullvm`
* Add CI testing for target.

### Testing:
CI tests passing: https://github.com/aws/aws-lc-rs/actions/runs/14890896214/job/41822424105?pr=702


.github/workflows/cross.yml
```diff
@@ -339,3 +341,47 @@ jobs:
...
+          - { sys: clang64, env: clang-x86_64, target: x86_64-pc-windows-gnullvm, cc: clang, cxx: g++ }
+    steps:
+      - name: Install MSYS2
+        uses: msys2/setup-msys2@v2
+        id: setup_msys2
+        with:
+          msystem: ${{ matrix.sys }}
+          update: true
+          pacboy: >-
+            clang
```
ognevny, line 364:
```suggestion
            cc
            rust-bindgen
```
cc is a meta package which points to main C compiler of the environment. as rust-bindgen is used during the build, better use one from MSYS2

ognevny: oh I see it's a cross compilation. you may not add rust-bindgen then

ognevny:
@justsmth instead of `clang` you should install `cc` package. it will download a proper compilers

ognevny:
I looked into an error log. the problem is that wrong target is passed to clang. I don't know which crate does the build, but for cc-rs it was fixed in 1.1.10 version: rust-lang/cc-rs#1176

ognevny:
we currently dropping support for 32-bit environments. there are no 32-bit packages with aws-lc, so feel free to net test mingw32

justsmth:
🎉 First success: https://github.com/aws/aws-lc-rs/actions/runs/14795425060/job/41541411147

> we currently dropping support for 32-bit environments. there are no 32-bit packages with aws-lc, so feel free to net test mingw32

Ok, I'll remove that target. Thanks!

ognevny:
small thing to fix
```
Compiler family detection failed due to error: ToolNotFound: failed to find tool "D:\\a\\_temp\\msys64\\clang64\\bin\\g++.exe": The system cannot find the file specified. (os error 2) (see https://docs.rs/cc/latest/cc/#compile-time-requirements for help)
```
this can be workarounded by downloading [gcc-compat](https://packages.msys2.org/packages/mingw-w64-clang-x86_64-gcc-compat) which is also named `gcc` for simplicity. but if it is just a warning and actual compiler family is detected then there is nothing to do

ognevny:
the issue above is not reproducible in MSYS2 CI by the way. probably something related with mixed host and target (we usually do host == target build)

skmcgrail (approved):

samuel40791765 (approved):

Issue #722, opened 2025-03-07
paste - no longer maintained


TroyKomodo:
### Problem:

[RUSTSEC-2024-0436](https://rustsec.org/advisories/RUSTSEC-2024-0436) was created with the crate `paste` being declared as unmaintained.

### Solution:

Look for alternatives to the crate such as https://lib.rs/crates/concat-idents?

justsmth:
We'll look into removing our need for `paste` or migrating to another crate.  Thanks for letting us know!

wiktor-k:
Just for the record the [advisory PR](https://github.com/rustsec/advisory-db/pull/2215) contains more background, including references to [forked crates](https://github.com/rustsec/advisory-db/pull/2215#issuecomment-2709031790) which should be drop-in replacements. (I'm not endorsing anything but as a user of `aws-lc-rs` I've been hit by the advisory warning too).

justsmth:
> Just for the record the https://github.com/rustsec/advisory-db/pull/2215 contains more background, including references to https://github.com/rustsec/advisory-db/pull/2215#issuecomment-2709031790 which should be drop-in replacements. (I'm not endorsing anything but as a user of aws-lc-rs I've been hit by the advisory warning too).


As of PR #724, we've removed our dependency on "paste" for everything except for tests. It's now only a "dev-dependency", so it shouldn't impact consumers.  Hopefully, we can get a patch release with this change published later today. Thanks!

PR #723, opened 2025-03-07
[DRAFT] Migrate from paste

justsmth:
### Issues:
Addresses:
* #722 

### Context
The paste crate is [no longer maintained](https://github.com/dtolnay/paste).

### Description of changes: 
Migrate use of "paste" to [concat-idents](https://crates.io/crates/concat-idents).

### Callouts
* Usage of the macros is nearly identical, except the constructed identity could not be used in another macro call to (e.g) `stringify!` or `use_bindings!`. Adjustments made accordingly.

### DIFF
--- a/aws-lc-fips-sys/Cargo.toml
+++ b/aws-lc-fips-sys/Cargo.toml
@@ -76,7 +76,7 @@ bindgen = { version = "0.69.5", optional = true }
 bindgen = { version = "0.69.5" }
 
 [dependencies]
-paste = "1.0.11"
+concat-idents = "1.1.5"
 
 [dev-dependencies]
 # Pinned dependency to preserve MSRV: 1.60.0  <= rust-version < 1.65.0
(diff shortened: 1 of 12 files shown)


jschwe:
`concat_idents` has not received any updates in 2 years and doesn't seem to be very popular either. I honestly think that migrating from paste to concat_idents would be a larger risk, since dependencies used in proc macros need to be trust worthy due to the lack of sandboxing.

justsmth:
> concat_idents has not received any updates in 2 years and doesn't seem to be very popular either. I honestly think that migrating from paste to concat_idents would be a larger risk, since dependencies used in proc macros need to be trust worthy due to the lack of sandboxing.

I agree. I attempted a migration, in part, to see how extensive our use of paste is and how hard it might be to replace. The best way to resolve this issue might be to write our own proc macro that provides some basic functionality.

Issue #741, opened 2025-03-19
implement `encoding::AsDer<PublicKeyX509Der<'a>>` for `rsa::PublicKey`


VasiliyS:
### Problem:

For compatibility reasons I need to work with RSA signing and encryption. 

I need to be able to serialize `rsa::PublicKey` to `SubjectPublicKeyInfo`  `der` encoding, which I then pem-encode. 

I generate the signing key in my code on a periodic basis. 

The issue is that `encoding::AsDer<PublicKeyX509Der<'a>>`  Trait is not implemented for the `rsa::PublicKey`.   

Which makes me jump though the hoops:
```rust
        // using PrivateDecryptingKey as it can serialize PK to SPKI
        let rsa_sk = PrivateDecryptingKey::generate(ks)?;
        let pk_der = rsa_sk.public_key().as_der()?;
        let pem = pem::Pem::new("PUBLIC KEY", pk_der.as_ref());
        let pk_pem = pem::encode(&pem);
        let rsa_sk_der = rsa_sk.as_der()?;
        // we actully want rsa::KeyPair as it does signing and impl Sync + Send
        let rsa_kp = KeyPair::from_pkcs8(rsa_sk_der.as_ref())?;
```
### Solution:

implement `encoding::AsDer<PublicKeyX509Der<'a>>` for `rsa::PublicKey`

* **Does this change any public APIs?** If yes, explain.

No, only `AsRef` is implmented for the type, so `AsDer` should not conflict.

* **Which algorithm(s) will this impact?**

None that I'm aware of.

justsmth:
Hello!  As you saw, we merged a PR for this here: https://github.com/aws/aws-lc-rs/pull/742

I apologize for the delay in getting this change into a release of `aws-lc-rs`. We plan to have a new release published soon.  Thanks!

justsmth:
[aws-lc-rs v1.13.0](https://github.com/aws/aws-lc-rs/releases/tag/v1.13.0) is now released. The latest release should remedy the problem identified above. We will now resolve this issue.  Feel free to comment, reopen or create a new issue if you have any other problems after updating.

We appreciate your help in improving our library. Thanks! ❤️

PR #742, opened 2025-03-20
Support RsaSubjectPublicKey to/from X509; Bump version

justsmth:
### Issues:
Addresses #741

### Description of changes: 
* Support `RsaSubjectPublicKey` serialization to/from X509
* Bump the patch version to 1.12.7.

### Testing:
Added test coverage for new functions.

### DIFF
--- a/aws-lc-rs/Makefile
+++ b/aws-lc-rs/Makefile
@@ -4,6 +4,8 @@ UNAME_S := $(shell uname -s)
 
 AWS_LC_RS_COV_EXTRA_FEATURES := unstable
 
+export AWS_LC_RS_DISABLE_SLOW_TESTS := 1
+
 asan:
 # TODO: This build target produces linker error on Mac.
 # Run specific tests:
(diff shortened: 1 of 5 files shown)


skmcgrail (approved):

aws-lc-rs/Makefile
```diff
@@ -4,6 +4,8 @@ UNAME_S := $(shell uname -s)
 
 AWS_LC_RS_COV_EXTRA_FEATURES := unstable
 
+export AWS_LC_RS_DISABLE_SLOW_TESTS := 1
```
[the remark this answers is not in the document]
justsmth: Yeah, `make test` takes forever to run w/o this environment variable set.  I typically set it manually in my environment each time, but got tired of having to do that. So I decided to just add it to the Makefile.

aws-lc-rs/src/rsa/key.rs
```diff
@@ -325,6 +337,14 @@ impl AsRef<[u8]> for PublicKey {
     }
 }
 
+impl AsDer<PublicKeyX509Der<'static>> for PublicKey {
+    fn as_der(&self) -> Result<PublicKeyX509Der<'static>, Unspecified> {
+        // TODO: refactor
```
samuel40791765, line 342: What's the plan to refactor?

samuel40791765 (approved):

aws-lc-rs/src/rsa/key.rs
justsmth: Currently the `PublicKey` struct only holds the serialized (RFC8017) bytes for the public key.  I would like for `PublicKey` to instead hold a reference to the EVP_PKEY, from which we could perform this and other operations.

PR #748, opened 2025-03-25
Update to bindgen 0.71.1

rukai:
### Issues:
No issue, this seems straightforward, if its not straightforward I can raise an issue.

### Description of changes: 

Bindgen 0.70 removes the dependency on `which` and `home`.
So upgrading from 0.69 to 0.71 will reduce the number of dependencies users need to pull in and improve build times.

I think that on most platforms, in standard configurations, aws-lc-rs doesnt actually depend on bindgen.
However due to cargo's design, `bindgen`, `which` and `home` are all included in the Cargo.lock but not actually built.
This requires users to drill down into their dependencies and manually inspect feature combinations and platforms to realize that bindgens `which` and `home` deps are not actually being built.

So to ease this pain lets upgrade to bindgen 0.71.

### Call-outs:
None

### Testing:
I built it locally on mac os with `--features bindgen`, I expect CI to thoroughly test different platforms.

### DIFF
--- a/aws-lc-fips-sys/Cargo.toml
+++ b/aws-lc-fips-sys/Cargo.toml
@@ -70,10 +70,10 @@ cc = "1.0.100"
 regex = "1"
 
 [target.'cfg(all(any(target_arch = "x86_64", target_arch = "aarch64"), any(target_os = "linux", target_os = "macos"), any(target_env = "gnu", target_env = "musl", target_env = "")))'.build-dependencies]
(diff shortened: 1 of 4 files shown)


rukai:
Ah, that is unfortunate, even bindgen 1.70 has an msrv of 1.70.
Well I'll go ahead and close this pr and let y'all make the upgrade when you are ready.

PR #774, opened 2025-04-10
Use cc_builder for riscv64gc-unknown-linux-gnu

### LINKED ISSUE #735: Remove dependency on CMake and Clang for riscv64
### Problem:

Thanks for the awesome project!

Please consider [pregenerating aws-lc-sys bindings](https://github.com/aws/aws-lc-rs/tree/main/aws-lc-sys#build-support) for riscv64 like what have done for x86_64 and aarch64.

The reason behind this is to make downstream RISC-V distros' life easier. For example, Arch Linux x86_64 doesn't need the cmake and clang dependencies while Arch Linux RISC-V needs. So we Arch Linux RISC-V has to [patch PKGBUILDs](https://github.com/search?q=repo%3Afelixonmars%2Farchriscv-packages+aws-lc-sys&type=commits) to add the dependencies for every package using `aws-lc-sys`.

Follow-up of #714 

#### Relevant details

System information: for linux, below info cae collected by running `uname -srvmp`
 * CPU architecture: riscv64
 * OS: Linux
 * Rust triplet: [riscv64gc-unknown-linux-gnu](https://rust-lang.github.io/rustup-components-history/riscv64gc-unknown-linux-gnu.html)

kxxt:
### Issues:
Resolves #735 

### Description of changes: 

This patch introduces support for building riscv64gc-unknown-linux-gnu target with cc_builder, which no longer requires extra cmake dependency.

This patch also updates the CI to generate the sources list for riscv64.

### Call-outs:

- Install clang 10 in Cross.toml because the one comes with the image is too old to support riscv
- Running the tests seems too slow because cross-rs runs them with qemu-user. Would this be a concern?

### Testing:

collect_build_src.sh -t riscv64gc-unknown-linux-gnu


kxxt:
Somehow the CI already installs clang-14. But when I try to run `collect_build_src.sh -t riscv64gc-unknown-linux-gnu` locally, it tries to use clang-6. Anyway, I have dropped the `install clang-10` change. Hopefully the CI would pass.

kxxt:
The quotes around `-Wno-string-compare` seems included into `AWS_LC_SYS_CFLAGS` itself. Dropping it should fix https://github.com/aws/aws-lc-rs/actions/runs/14467854826/job/40693683702?pr=774#step:11:1200

justsmth:
To test this properly, I triggered the workflow here: https://github.com/aws/aws-lc-rs/actions/runs/14536943083

justsmth (approved):

skmcgrail (approved):

Issue #814, opened 2025-05-27
"No rule to make target" for libaws_lc_0_29_0_crypto.a


garretwilson:
(Cross-posting from awslabs/aws-sdk-rust#1298 as requested.)

### Describe the bug

Compiling a minimal Rust CLI on Windows 10 using the GNU chain, no actual AWS calls, but only a dependency to `aws-config`, results in:

> `*** No rule to make target 'aws-lc/crypto/CMakeFiles/crypto_objects.dir/chacha/chacha-x86_64.asm.obj', needed by 'artifacts/libaws_lc_0_29_0_crypto.a'.`

I have a simple Rust CLI on Windows 10 using the GNU toolchain. I have installed MSYS2 (UCRT64) and installed the toolchain:

```bash
pacman -S --needed base-devel mingw-w64-ucrt-x86_64-toolchain
```

I added the appropriate MSYS2 `bin` folder to my `PATH`.

I added the following to my `Cargo.toml` file:

```toml
[dependencies]
aws-config = { version = "1.6.3", features = ["behavior-version-latest"] }
```

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

When I run `cargo build`, I expect the build to be successful.

### Current Behavior

```
…
  CMAKE = Some("cmake")
  running: "cmake" "…\\.cargo\\registry\\src\\index.crates.io-1949cf8c6b5b557f\\aws-lc-sys-0.29.0" "-G" "MinGW Makefiles" "-DBUILD_SHARED_LIBS=0" "-DBORINGSSL_PREFIX=aws_lc_0_29_0_" "-DBORINGSSL_PREFIX_HEADERS=C:\\Users\\garret\\.cargo\\registry\\src\\index.crates.io-1949cf8c6b5b557f\\aws-lc-sys-0.29.0\\generated-include" "-DBUILD_TESTING=OFF" "-DBUILD_TOOL=OFF" "-DBUILD_LIBSSL=OFF" 
[...]

PR #815, opened 2025-05-28
[DRAFT] Improve handling of NASM failures

justsmth:
### Issues:
Addresses #814

### Description of changes: 
WIP

### Call-outs:
WIP

### Testing:
WIP

### DIFF
--- a/aws-lc-sys/builder/main.rs
+++ b/aws-lc-sys/builder/main.rs
@@ -628,7 +628,11 @@ fn has_pregenerated() -> bool {
 }
 
 fn test_nasm_command() -> bool {
-    execute_command("nasm".as_ref(), &["-version".as_ref()]).status
+    let status = execute_command("nasm".as_ref(), &["-version".as_ref()]).status;
+    if !status {
+        emit_warning("NASM command not found or failed to execute.");
+    }
+    status
 }
 
 fn prepare_cargo_cfg() {


aws-lc-sys/builder/main.rs
```diff
@@ -628,7 +628,11 @@ fn has_pregenerated() -> bool {
 }
 
 fn test_nasm_command() -> bool {
-    execute_command("nasm".as_ref(), &["-version".as_ref()]).status
+    let status = execute_command("nasm".as_ref(), &["-version".as_ref()]).status;
+    if !status {
+        emit_warning("NASM command not found or failed to execute.");
```
garretwilson, line 633: If the build will break, then this shouldn't be a warning—it should be an error. What's the point of hiding a little warning somewhere in a long output of multiple "The system cannot find the file specified" messages and other low-level messages that newcomer who just wants to add AWS SDK for Rust to the project won't have any way to decipher?

If the answer is, "well, it won't necessarily fail if the appropriate environmental variable is set to not use NASM at all", then the validation step should check those environment variables as well and not even issue a warning.

The point is that this is not a warning level issue. It's either an error or not a problem at all, based upon the other build settings.

justsmth: If NASM is required then [the build will fail](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-sys/builder/cmake_builder.rs#L424) early, while checking whether all build dependencies can be found. 

The purpose of the message above (along with [the other messages](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-sys/builder/cmake_builder.rs#L403-L410)) is only to provide more information/context when a build does fail, possibly due to other causes.

What I believe our build logic is missing is detection/avoidance of the situation (like the one [you reported](https://github.com/aws/aws-lc-rs/issues/814)) where the script we use to provide the prebuilt NASM objects is not able to execute. That still requires some research.

justsmth:
Change included in #830.

PR #826, opened 2025-06-17
[DRAFT] Init impl for sign/verify w/ digest

justsmth:
### Description of changes: 
* Start of work for supporting signing/verifying precomputed digests.

### Call-outs:
TOOO

### Testing:
TODO

### DIFF
--- a/aws-lc-rs/src/ec/key_pair.rs
+++ b/aws-lc-rs/src/ec/key_pair.rs
@@ -4,15 +4,15 @@
 // SPDX-License-Identifier: Apache-2.0 OR ISC
 
 use crate::aws_lc::{EVP_PKEY, EVP_PKEY_EC};
-use core::fmt;
-use core::fmt::{Debug, Formatter};
-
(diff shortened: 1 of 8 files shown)


aws-lc-rs/src/ec/key_pair.rs
```diff
@@ -235,6 +235,31 @@ impl EcdsaKeyPair {
             EcdsaSignatureFormat::Fixed => ec::ecdsa_asn1_to_fixed(self.algorithm.id, &out_sig)?,
         })
     }
+
+    /// Returns the signature of the digest using a random nonce.
```
skmcgrail:
```suggestion
    /// Returns the signature of the provide digest.
```

Using the term nonce here feels out of place since we don't use the term elsewhere when discussing ECDSA signatures in the library to my knowledge.

skmcgrail (approved):

skmcgrail (approved):

PR #832, opened 2025-06-30
Change 'try_sign' to pub in ED25519 module

theswiftfox:
### Description of changes: 
Currently, the only way to use ed25519 to sign a message is using the `sign` function. However, this function can panic, in the case that the signing operation failed. To allow a user of the library to try to sign a message and handle the error case (without relying on`std::panic::catch_unwind`) this change exposes the `try_sign` as public again. This was also the behaviour before the [rename of the function](https://github.com/aws/aws-lc-rs/commit/b43fca398e69a9bad85f2321cdf1af9ecd7ffaec#diff-1f32d2fbb7ef839cab8718f754391a1f2d04452131eff2eeb3c2d6b6d61b8664L218).

### Call-outs:
I added a docstring to the function, similar to the one on `sign`. However, I'm not sure if this function would need the FIPS disclaimer as well, or if that was related to the possible panic.  

### Testing:
There is no change in how the library works, only in what is exposed in the public interface. Therefore no additional testing effort / checks are necessary.  

Elena Gantner [elena.gantner@mercedes-benz.com](mailto:elena.gantner@mercedes-benz.com) on behalf of Mercedes-Benz Tech Innovation GmbH
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)

### DIFF
--- a/aws-lc-rs/src/ed25519.rs
+++ b/aws-lc-rs/src/ed25519.rs
@@ -410,7 +410,6 @@ impl Ed25519KeyPair {
     /// # Errors
     /// Returns `error::Unspecified` if the signing operation fails.
     #[inline]
-    #[must_use]
     pub fn try_sign(&self, msg: &[u8]) -> Result<Signature, Unspecified> {
         let sig_bytes = self.evp_pkey.sign(msg, None, No_EVP_PKEY_CTX_consumer)?;
 
--- a/aws-lc-rs/tests/ed25519_tests.rs
+++ b/aws-lc-rs/tests/ed25519_tests.rs
@@ -32,7 +32,7 @@ fn test_signature_ed25519() {
             let expected_sig = test_case.consume_bytes("SIG");
 
             let key_pair = Ed25519KeyPair::from_seed_unchecked(&seed).unwrap();
-            let actual_sig = key_pair.sign(&msg);
+            let actual_sig = key_pair.try_sign(&msg)?;
             assert_eq!(&expected_sig[..], actual_sig.as_ref());
 
             let key_pair = Ed25519KeyPair::from_seed_and_public_key(&seed, &public_key).unwrap();


aws-lc-rs/src/ed25519.rs
```diff
@@ -402,8 +402,13 @@ impl Ed25519KeyPair {
         Self::try_sign(self, msg).expect("ED25519 signing failed")
     }
 
+    /// Returns the signature of the message `msg`.
+    ///
```
andrewhop, line 406:
```suggestion
    ///
    // # FIPS
    // This method must not be used.
    //
```

andrewhop (approved):
Thanks for the change! I'm going to hunt down how/why this got removed.

aws-lc-rs/src/ed25519.rs
```diff
@@ -402,8 +402,16 @@ impl Ed25519KeyPair {
...
+    // # FIPS
+    // This method must not be used.
+    //
+    /// # Errors
+    /// Returns `error::Unspecified` if the signing operation fails.
     #[inline]
-    fn try_sign(&self, msg: &[u8]) -> Result<Signature, Unspecified> {
+    #[must_use]
```
justsmth: This `must_use` causes a clippy error b/c the return type `Result` is already a `must_use`, so this is redundant.

justsmth:
> Thanks for the change! I'm going to hunt down how/why this got removed.

The `try_sign` or `sign_checked` method has never been publicly exposed. (That commit with the rename is from October 2022 but our first release, 1.0, was in April 2023.) 

The reason we didn't expose it was due to the assumption that `sign` would never fail unless the environment was already compromised in some way (e.g., unable to allocate memory). And, it's possible that this is a bad assumption. Either way, exposing a `try_sign` is the right thing to do.

PR #834, opened 2025-07-02
expose algorithm used to create ecdsa keypair

soundofspace:
### Description of changes: 
Expose algorithm used to create ecdsa keypair so users who need this, can access it without also having to store and pass it around 

### DIFF
--- a/aws-lc-rs/tests/ecdsa_tests.rs
+++ b/aws-lc-rs/tests/ecdsa_tests.rs
@@ -140,7 +140,8 @@ fn ecdsa_generate_pkcs8_test() {
         println!();
         println!();
 
-        EcdsaKeyPair::from_pkcs8(alg, pkcs8.as_ref()).unwrap();
+        let key_pair = EcdsaKeyPair::from_pkcs8(alg, pkcs8.as_ref()).unwrap();
+        assert_eq!(key_pair.algorithm(), *alg);
     }
 }
 


andrewhop (dismissed):

skmcgrail (approved):

soundofspace:
Hi guys, thanks for the review.

Updated description so `assert-license` should work now.
Anything else needed from my side?

aws-lc-rs/src/ec/key_pair.rs
```diff
@@ -318,5 +324,6 @@ mod tests {
         assert_eq!(key_pair.evp_pkey, key_pair_5208.evp_pkey);
         assert_eq!(key_pair.evp_pkey, key_pair_5915.evp_pkey);
         assert_eq!(key_pair_5208.evp_pkey, key_pair_5915.evp_pkey);
+        assert_eq!(key_pair_5915.algorithm, &ECDSA_P256_SHA256_FIXED_SIGNING);
```
justsmth, line 327: This doesn't actually invoke the function that you added. I'll push a commit to update an "external" test to invoke this function.

skmcgrail (approved):

aws-lc-rs/src/ec/key_pair.rs
soundofspace: @justsmth was indeed a mistake from my end, thanks for fixing it

PR #847, opened 2025-07-20
Add support for decrypting ANSI X9.23 / ISO 10126-padded AES

tstenner:
### Description of changes: 

[EBICS](https://en.wikipedia.org/wiki/Electronic_Banking_Internet_Communication_Standard) (one of the most widespread european protocols for corporate banking) uses AES with ISO 10126 padding.

This is equivalent to PKCS7 except that all bytes but the last padding byte are generated randomly, i.e. PKCS7 padded data can be decoded as ISO 10126 but not the other way round.

This PR adds `cbc_iso10126` as a `cbc_pkcs7` alternative to `PaddedBlockDecryptingKey` for compatibility with clients using ISO 10126 (and ANSI X9.23) padding.

### Call-outs:

The only addition to the public API (`cbc_iso10126`) could be hidden behind a `#[cfg(not(feature = "fips"))]` feature gate. The same could apply to the mention of fips in the comments, but I don't have access to the FIPS lists myself.

### Testing:

Tested locally.

### DIFF
--- a/aws-lc-rs/src/cipher/padded.rs
+++ b/aws-lc-rs/src/cipher/padded.rs
@@ -49,7 +49,7 @@ impl PaddingStrategy {
         match self {
             PaddingStrategy::ISO10126 => {
                 let padding: u8 = in_out[in_out.len() - 1];
-                if padding == 0 || padding > block_len as u8 {
+                if padding == 0 || padding as usize > block_len {
                     return Err(Unspecified);
                 }
 


skmcgrail (approved):

skmcgrail:
This looks good to me, we just need to get a test case added to verify this for the decryption path.

tstenner:
I have added two unit tests that check that a previously known ciphertext and a specifically crafted padded text with semi-random bytes get unpadded properly.

skmcgrail (dismissed):

aws-lc-rs/src/cipher/padded.rs
```diff
@@ -390,7 +429,13 @@ mod tests {
 
                 let context = encrypting_key.less_safe_encrypt(&mut in_out, ec).unwrap();
 
-                assert_eq!(expected_ciphertext, in_out);
+                if ($padding == PaddingStrategy::ISO10126) {
+                    // This padding scheme is technically non-deterministic in nature if the padding is more then one
```
torben-hansen, line 433: For unknown input to the padding decoder, this assumption is true.
But for your implementation you can make a stronger assumption and testing doesn't take unknown inputs.
Hence, does one want a test to preserve current implementation behaviour or let it be general.

I have 0 strong opinion on it.

torben-hansen (dismissed):

aws-lc-rs/src/cipher/padded.rs
tstenner: If it were an implementation for a new (un)padding scheme, it would make sense to have more tests.

As it is, this is just a simple compatibility option that works for any scheme with the number of padding bytes in the last byte so I wouldn't complicate things until there's a real need to differentiate between ISO 10126, ANSI X9.23 and/or PKCS5.

tstenner:
Clippy didn't like `padding > block_len as u8`, so I changed it to `padding as usize > block_len`.

torben-hansen (approved):

PR #857, opened 2025-08-13
Support verification of P256/P384+SHA512 signatures

ctz:
### Description of changes: 
This PR adds support for `ECDSA_P256_SHA512_ASN1` and `ECDSA_P384_SHA512_ASN1`.

This is needed because in TLS1.2 offering support for `SignatureAndHashAlgorithm {signature: ecdsa, hash: sha512}` is not specific to any curve, so it implies support for  `ECDSA_P256_SHA512_ASN1`, `ECDSA_P384_SHA512_ASN1`, and `ECDSA_P521_SHA512_ASN1`.

### Call-outs:
I did not add the `_FIXED` variants for these, as (eg) `ECDSA_P256_SHA384_ASN1` already does not have a `_FIXED` counterpart.

I did not also add any `_SIGNING` counterparts because I think these weird/legacy combinations shouldn't be encouraged.

### Testing:
I have added one positive and one negative test for each algorithm, taken from wycheproof.

re https://github.com/rustls/rustls/issues/2477


justsmth:
Thanks for the PR! 

> I did not also add any _SIGNING counterparts because I think these weird/legacy combinations shouldn't be encouraged.

Yeah, this is consistent with the other "odd" combinations that we have. 

A couple of the RSAs have a `_FOR_LEGACY_USE_ONLY` appended, but that's more b/c they're no longer considered secure. Whereas combinations like (P256+SHA512) are secure, but are rarely used due to their security-level mismatch -- i.e., less efficient w/o improving security.

justsmth (approved):

PR #862, opened 2025-08-15
ParsedPublicKey for agreement

justsmth:
### Description of changes: 
Provide a ParsedPublicKey type that can be used for agreement.

### Call-outs:
The current implementation provides an indication of the format from which the public key was parsed.  I'm not sure whether that is a use-case we want to support.

### Testing:
Added tests.


skmcgrail (dismissed):

aws-lc-rs/src/agreement/ephemeral.rs
```diff
@@ -98,9 +98,9 @@ impl EphemeralPrivateKey {
 #[allow(clippy::needless_pass_by_value)]
 #[allow(clippy::missing_panics_doc)]
 #[allow(clippy::module_name_repetitions)]
-pub fn agree_ephemeral<B: AsRef<[u8]>, F, R, E>(
+pub fn agree_ephemeral<B: TryInto<ParsedPublicKey>, F, R, E>(
```
samuel40791765, line 101: Very minor, but we should probably call out the API changes done here and with `agree()` in the release notes.

aws-lc-rs/src/agreement.rs
```diff
@@ -663,6 +661,136 @@ impl<B: AsRef<[u8]>> UnparsedPublicKey<B> {
...
+                let key = if let Ok(evp_pkey) =
+                    LcPtr::<EVP_PKEY>::parse_rfc5280_public_key(key_bytes, EVP_PKEY_EC)
+                {
+                    validate_ec_evp_key(&evp_pkey.as_const(), nid)?;
+                    format = ParsedPublicKeyFormat::X509;
+                    evp_pkey
+                } else if let Ok(evp_pkey) = parse_sec1_public_point(key_bytes, nid) {
+                    format = match key_bytes[0] {
+                        0x02 | 0x03 => ParsedPublicKeyFormat::Compressed,
+                        _ => ParsedPublicKeyFormat::Uncompressed,
```
samuel40791765: This matching is a bit inaccurate, `Uncompressed` formats typically only match to `0x04`. There's the possibility of `Hybrid` formats which are indicated with `0x06` or `0x07` and the default patch would match this to that.

I don't think we necessarily need to support `Hybrid` formats here if it's too much a hassle as they're rare and relatively unused, but we should definitely be a bit stricter here. Either adding an additional `ParsedPublicKeyFormat::Unknown` or directly failing here on a mismatch would also do?

justsmth: Yeah, I can add an Unknown.

samuel40791765 (approved):

smittals2 (approved):

Issue #849, opened 2025-07-23
Add way to pre-parse signature keys (i.e. turn `signature::UnparsedPublicKey` to `PublicKey`)


LukasKalbertodt:
### Problem:

My application has access to only a public key and needs to verify signatures. I can create a `signature::UnparsedPublicKey` and use that to verify signatures. However, now the key is parsed every time the signature is verified. In `verify_asn1_signature`, `parse_ec_public_key` is called. This has two disadvantages:

- It makes verification slower
- I cannot error early and report an invalid key to the user

### Solution:

There should be a way to create a type that parses the key already and verifies that it's valid. Then that step does not have to be performed during verification and I can detect ahead of time whether the key is even valid. I am not sure if `signature::PublicKey` is the correct type for that or if there should be a new `ParsedPublicKey`.

### Related Issues:

- https://github.com/briansmith/ring/issues/1601
- https://github.com/briansmith/ring/issues/1577

justsmth:
Hello! Thanks for submitting this issue.

I've worked briefly on this issue before, and have a draft for it sitting on a branch somewhere. I'll look for it and try to update it for review. (Possibly late next week at the earliest.)

justsmth:
Ok, I found the branch where I had previously worked on this issue. I even started a PR for it here: https://github.com/aws/aws-lc-rs/pull/519

I'll take a closer look to verify we can do this in a clean way w/o breaking our existing API.

justsmth:
Hello! We just released [aws-lc-rs v1.14.0](https://github.com/aws/aws-lc-rs/releases/tag/v1.14.0), which added a [`ParsedPublicKey` type](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/signature/struct.ParsedPublicKey.html). We believe this release addresses the use-case described above.

This issue will now be closed. Feel free to comment here and/or open another issue if you continue to experience problems after upgrading. Thanks for your help in improving our library! 😀

LukasKalbertodt:
Amazing, thanks for implementing this so quickly :)

I had some benchmarks flying around anyway so I updated to the new version and added benchmarks using the `ParsedKey`:

| Algo | Unparsed | Parsed |
| --- | --- | --- | 
| ECDSA_P256_SHA256_FIXED | 34.6µs | 33.8µs |
| ECDSA_P384_SHA384_FIXED |133.5µs |  132.2µs |
| ED25519 | 24.9µs | 24.8µs |

[...]

PR #863, opened 2025-08-15
ParsedPublicKey for signature

justsmth:
### Issues:
Addresses #849 

### Description of changes: 
Provides a new `signature::ParsedPublicKey` type to be used for signature verification.
* Reduces latency when same public key must be used more than once.
* Informs about the validity of a public key when it's parsed, prior to an attempt to use it.

### Testing:
Added/updated testing to provide coverage.

### DIFF
--- a/aws-lc-rs/src/signature/parsed_public_key_tests.rs
+++ b/aws-lc-rs/src/signature/parsed_public_key_tests.rs
@@ -1,3 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
 use crate::digest::{digest, SHA256};
 use crate::rsa::KeySize;
 use crate::signature::{


aws-lc-rs/src/ec/signature.rs
```diff
@@ -192,43 +193,86 @@ impl VerificationAlgorithm for EcdsaVerificationAlgorithm {
...
 }
 
+impl ParsedVerificationAlgorithm for EcdsaVerificationAlgorithm {
+    fn parsed_verify_sig(
+        &self,
+        public_key: &ParsedPublicKey,
+        msg: &[u8],
+        signature: &[u8],
+    ) -> Result<(), Unspecified> {
+        self.verify_ecdsa(msg, signature, public_key.key())
```
skmcgrail, line 263: Should this and all the other implementations check that the passed in public_key type / algorithm matches? Right now this appears to rely on the AWS-LC APIs to catch the failure for the public key type being incorrect from the looks of it.

skmcgrail: I guess this is unlikely to be problematic given that we don't expose `ParsedVerificationAlgorithm` directly, and manage the detail internally.

aws-lc-rs/src/signature.rs
```diff
@@ -429,6 +572,72 @@ impl<B: AsRef<[u8]>> UnparsedPublicKey<B> {
...
+    bytes: &[u8],
+    algorithm: &'static dyn VerificationAlgorithm,
+) -> Result<ParsedPublicKey, KeyRejected> {
+    let parsed_algorithm: &'static dyn ParsedVerificationAlgorithm;
+
+    let key = if algorithm.type_id() == TypeId::of::<EcdsaVerificationAlgorithm>() {
+        #[allow(clippy::cast_ptr_alignment)]
+        let ec_alg = unsafe {
+            &*(algorithm as *const dyn VerificationAlgorithm).cast::<EcdsaVerificationAlgorithm>()
+        };
```
skmcgrail, line 604: `(algorithm as &dyn Any).downcast_ref::<EcdsaVerificationAlgorithm>().expect("?")` seems like the logical way you are supposed to use with TypeId, and avoids the unsafe aspects.

justsmth: Yeah, that was my first approach and it worked cleanly.  The problem is that Trait upcasting (e.g., `(algorithm as &dyn Any)`) was not supported until Rust 1.86: https://blog.rust-lang.org/2025/04/03/Rust-1.86.0/#trait-upcasting

The approach I took seemed to work best, at least while we maintain 1.70 compatibility.

aws-lc-rs/src/ec/signature.rs
justsmth: Yeah, at this point we've parsed the public key, so we've confirmed that the key matches the indicated signing algorithm.

skmcgrail (approved):

justsmth:
When I rebased there was a merge conflict that I resolved, but it dismissed Sean's approval. 😔

skmcgrail (approved):

samuel40791765 (approved):

PR #869, opened 2025-08-29
Optionally disable padding in StreamingDecryptingKey

tstenner:
### Issues:

Continuation of #847.

Some applications (e.g. [EBICS](https://en.wikipedia.org/wiki/Electronic_Banking_Internet_Communication_Standard)) still require compatibility with padding schemes like ISO 10126 and ANSI X9.23.

### Description of changes: 

This PR adds the method `StreamDecryptingKey::disable_padding()` as a safe wrapper for the existing `EVP_CIPHER_CTX_set_padding()` function. This allows applications to decrypt data padded with one of these padding schemes and validate the padding without adding complexity to aws-lc-rs.

### Testing:

Testing will be added. What kind of coverage is expected?

### DIFF
--- a/aws-lc-rs/src/cipher/streaming.rs
+++ b/aws-lc-rs/src/cipher/streaming.rs
@@ -1,6 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR ISC
 
+use aws_lc::EVP_CIPHER_CTX_set_padding;
+
 use crate::aws_lc::{
     EVP_CIPHER_CTX_new, EVP_CIPHER_iv_length, EVP_CIPHER_key_length, EVP_DecryptFinal_ex,
     EVP_DecryptInit_ex, EVP_DecryptUpdate, EVP_EncryptFinal_ex, EVP_EncryptInit_ex,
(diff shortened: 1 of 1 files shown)


justsmth:
> ... This allows applications to ... validate the padding ...

I'm not familiar with this use-case.  Why would padding need to be validated?

tstenner:
With the padding schemes I need to support, the last byte `n` is the number of padding bytes (so it needs to be <=16 with a valid key). With other schemes, all the other padding bits are supposed to be 0 (some X9.23 implementations) or `n`.
The main validation (and necessary step to remove the padding) would be to check that the last byte is at most the block size.

justsmth:
> ... The main validation (and necessary step to remove the padding) would be to check that the last byte is at most the block size.

I believe that validation is already being done [here](https://github.com/aws/aws-lc-rs/blob/ceb115194a5fc9fbdb66a2ac93a2aa2b32c0edcd/aws-lc-rs/src/cipher/padded.rs#L51-L54) and [here](https://github.com/aws/aws-lc-rs/blob/ceb115194a5fc9fbdb66a2ac93a2aa2b32c0edcd/aws-lc-rs/src/cipher/padded.rs#L63-L66).

justsmth:
> > ... The main validation (and necessary step to remove the padding) would be to check that the last byte is at most the block size.
> 
> I believe that validation is already being done [here](https://github.com/aws/aws-lc-rs/blob/ceb115194a5fc9fbdb66a2ac93a2aa2b32c0edcd/aws-lc-rs/src/cipher/padded.rs#L51-L54) and [here](https://github.com/aws/aws-lc-rs/blob/ceb115194a5fc9fbdb66a2ac93a2aa2b32c0edcd/aws-lc-rs/src/cipher/padded.rs#L63-L66).

Ahh. My links above only apply to `PaddedBlockDecryptingKey`.  However, I believe this validation is occurring already within AWS-LC.  I would need to dig a little deeper to find it.

Do you have reason to believe that this validation is not occurring currently?

tstenner:
The validation happens within `evp_decryptfinal_ex` which is called from `StreamingDecryptingKey::finish()`, but it is hardcoded to PKCS padding (and fails with ciphertexts padded with these legacy padding schemes).

While I'm typing this I thought of working around this problem by omitting the call to `StreamingDecryptingKey::finish()` and pushing a dummy ciphertext block (yielding the last real plaintext block). I'm changing the PR to a draft until I've tested if this works.

tstenner:
Calling

```rust
decryptor.update(&[0u8; 16], &mut self.decrypted);
```

instead of 
```
decryptor.finish(&[0u8; 16], &mut self.decrypted);
```

works for my use case.

In case these legacy padding schemes should be supported a bit better, this PR is a relatively simple fix, but it's not strictly needed.

PR #879, opened 2025-09-12
Common traits for ParsedPublicKey

justsmth:
### Issues:
Addresses #849 

### Description of changes: 
Add several common traits (`Send`, `Sync`, `Clone`, and `AsRef<[u8]>`) for `signature::` and `agreement::` `ParsedPublicKey`.

### Testing:
Tests updated.

### DIFF
--- a/aws-lc-rs/src/signature.rs
+++ b/aws-lc-rs/src/signature.rs
@@ -405,6 +405,13 @@ pub struct ParsedPublicKey {
     bytes: Box<[u8]>,
 }
 
+// See EVP_PKEY documentation here:
+// https://github.com/aws/aws-lc/blob/125af14c57451565b875fbf1282a38a6ecf83782/include/openssl/evp.h#L83-L89
+// An |EVP_PKEY| object represents a public or private key. A given object may
+// be used concurrently on multiple threads by non-mutating functions, provided
+// no other thread is concurrently calling a mutating function. Unless otherwise
+// documented, functions which take a |const| pointer are non-mutating and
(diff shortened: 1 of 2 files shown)


aws-lc-rs/src/signature.rs
```diff
@@ -397,12 +397,17 @@ pub struct UnparsedPublicKey<B: AsRef<[u8]>> {
...
+#[derive(Clone)]
 pub struct ParsedPublicKey {
     algorithm: &'static dyn VerificationAlgorithm,
     parsed_algorithm: &'static dyn ParsedVerificationAlgorithm,
     key: LcPtr<EVP_PKEY>,
+    bytes: Box<[u8]>,
 }
 
+unsafe impl Send for ParsedPublicKey {}
+unsafe impl Sync for ParsedPublicKey {}
```
LukasKalbertodt, line 416: Would be great to have comments explaining why implementing these traits is safe.

justsmth: Yeah, the documentation explains it here: https://github.com/aws/aws-lc/blob/125af14c57451565b875fbf1282a38a6ecf83782/include/openssl/evp.h#L83-L89
```
// An |EVP_PKEY| object represents a public or private key. A given object may
// be used concurrently on multiple threads by non-mutating functions, provided
// no other thread is concurrently calling a mutating function. Unless otherwise
// documented, functions which take a |const| pointer are non-mutating and
// functions which take a non-|const| pointer are mutating.
```

I'll add the link above as a comment.

PR #895, opened 2025-09-25
Adds AES CBC mode no padding

Elvrarin:
### Issues:
See SIM P307101963

### Description of changes: 
Currently, only the PaddedBlock keys and Streaming keys supports CBC mode.
This adds CBC mode for EncryptingKey and DecryptingKey. This allows AES CBC encryption/decryption with no padding.

### Call-outs:
Since no padding is used, input data must be block aligned, length of input must be a multiple of 16 bytes.
I also included the `DANGER` message in the doc string, since I think the use of this is unconventional.

### Testing:
Unit tests added following examples of existing unit tests.

#### KAT
Added known-answer tests from "Appendix F" of SP800-38a:
* https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38a.pdf

### DIFF
--- a/aws-lc-rs/src/cipher.rs
+++ b/aws-lc-rs/src/cipher.rs
@@ -1151,4 +1151,44 @@ mod tests {
         "6bc1bee22e409f96e93d7e117393172aae2d8a571e03ac9c9eb76fac45af8e5130c81c46a35ce411e5fbc1191a0a52eff69f2445df4f9b17ad2b417be66c3710",
         "f3eed1bdb5d2a03c064b5a7e3db181f8591ccb10d410ed26dc5ba74a31362870b6ed21b99ca6f4f9f153e7b1beafed1d23304b7a39f9f3ff067d8d8f9e24ecc7"
     );
+
+    cipher_kat!(
+        test_sp800_38a_cbc_aes128,
+        &AES_128,
+        OperatingMode::CBC,
+        "2b7e151628aed2a6abf7158809cf4f3c",
+        "000102030405060708090a0b0c0d0e0f",
+        "6bc1bee22e409f96e93d7e117393172a",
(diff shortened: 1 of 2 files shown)


aws-lc-rs/src/cipher.rs
```diff
@@ -548,6 +548,23 @@ impl EncryptingKey {
         Self::new(key, OperatingMode::CFB128)
     }
 
+    /// Constructs an `EncryptingKey` operating in cipher block chaining (CBC) mode using the provided key.
+    ///
+    /// # ☠️ ️️️DANGER ☠️
+    /// Offered for computability purposes only. This is an extremely dangerous mode, and
```
justsmth: Here and below: `s/computability/compatibility/`

justsmth:
This is an API change that would cause consumers lots of confusion -- the same code would succeed for some inputs and fail for others at runtime. 

There should be a more explicit runtime check that produces an unambiguous error message indicating when the failure is due to the input size not being aligned w/ the block size. This API might also need to be behind a feature flag so the only users than have access to it are those who have sought it out.

So this likely needs to be part of a slightly larger change to the API to make the functionality more user-friendly.  I'll put some thought into what API changes might be best for this purpose.

Elvrarin:
> the same code would succeed for some inputs and fail for others at runtime.

I agree this is a bit confusing, but the existing ECB mode for EncryptKey and DecryptKey also behaves the same way, that is to require input size to be block-aligned. There is the "skull emoji" DANGER message that will be shown on the docs.rs, I think it already push consumers off of these API, save for those who understands the intricacies of ECB mode.

The larger API/feature flag refactoring is beyond what I can single handedly do. Let me know how I can help here. Thanks.

justsmth (approved):

skmcgrail (approved):

PR #896, opened 2025-09-26
Universal Binding - generated from workflow

justsmth:
### Context
The need for bindings generation has been a pain point for consumers since the library's first release.  To create bindings using "bindgen" requires that the build environment provide libclang and possibly additional environment variables so that the library can be found.  When it's needed, bindings generation also adds significant latency to our build time.  To mitigate this for most common platforms, we provided pre-generated bindings.  However, there are still a significant number of consumers whose builds target platforms for which we don't provide bindings.

### Description of changes: 
This PR updates aws-lc-sys to provide universal bindings that are applicable to all platforms. These bindings cover more narrowly the subset of AWS-LC APIs that are needed by aws-lc-rs. This eliminates the need for bindings generation during the build for the majority of our consumers.

### Call-outs:
 Consumers accessing AWS-LC more directly via aws-lc-sys, might still need to generate bindings if they are targeting a platform that we do not pre-generate for.
* Automated workflow: https://github.com/aws/aws-lc-rs/actions/runs/18530029767
#### Manual commits
* [Refactor build to utilize universal bindings](https://github.com/aws/aws-lc-rs/pull/896/commits/65d8b868bea88a9fb5802b5fc25324c9da69fcdd)
* [Update cross tests to use universal bindings](https://github.com/aws/aws-lc-rs/pull/896/commits/d3f16267d68169ccb3054cbf4474c261f8ff274b)
* [Update aws-lc-sys bindings workflow](https://github.com/aws/aws-lc-rs/pull/896/commits/a96d90369769d9f9db140447eb2ab14cbd193c0e)
* [Fix clippy](https://github.com/aws/aws-lc-rs/pull/896/commits/34fccaf808daffdcaea3a8e90f40875a70a5f8d3)
* [Must generate universal bindings before collecting source](https://github.com/aws/aws-lc-rs/pull/896/commits/a20f0d2617ff1ecf7a13e7c62255150b11982f60)
* [Bump version (no change to alignment)](https://github.com/aws/aws-lc-rs/pull/896/commits/fe2963563e0b0f7d43f810280800b538e4f799b7)

### DIFF
--- a/aws-lc-sys/builder/cc_builder/riscv64gc_unknown_linux_gnu.rs
+++ b/aws-lc-sys/builder/cc_builder/riscv64gc_unknown_linux_gnu.rs
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR ISC
-// Tue Oct 14 13:48:50 UTC 2025
+// Wed Oct 15 13:36:34 UTC 2025
 
 use crate::cc_builder::Library;
 


skmcgrail (approved):

aws-lc-sys/builder/sys_bindgen.rs
```diff
@@ -31,6 +31,48 @@ impl ParseCallbacks for StripPrefixCallback {
...
+
+const BLOCKED_FUNCTIONS: [&str; 5] = [
+    "BN_print_fp",
+    "CBS_parse_generalized_time",
+    "CBS_parse_utc_time",
+    "ERR_print_errors_fp",
+    "RSA_print_fp",
+];
+
+const BLOCKED_TYPES: [&str; 4] = ["FILE", "fpos_t", "tm", "__sFILE"];
```
samuel40791765, line 74: I can put my finger on the reasons why these functions are "BLOCKED", but would it help to document why here to make it easier to understand?

aws-lc-sys/builder/sys_bindgen.rs
```diff
@@ -31,6 +31,48 @@ impl ParseCallbacks for StripPrefixCallback {
     }
 }
 
+const ALLOWED_HEADERS: [&str; 29] = [
```
samuel40791765, line 34: Also curious as to what specific headers would "break" the universal bindings. In other words, which specific header functions cause platform bindings to differ?

aws-lc-sys/builder/sys_bindgen.rs
justsmth: These were the functions from the included headers that accepted a parameter with one of the blocked types. I'll add a comment here in my PR for eliminating the CMake requirement.

aws-lc-sys/builder/sys_bindgen.rs
justsmth: * The "allowed" headers are just the ones necessary to providing the bindings for the functions used by aws-lc-rs. We could include more, but they're not needed.
* What causes the bindings to differ across platforms are the "libc" types like `FILE`, `tm`, etc. These are the types that are "blocked" above.

PR #900, opened 2025-09-30
Failure when CFLAGS provides -O3

justsmth:
### Context
In build environments where CFLAGS provides a flag enabling optimization, this can cause compilation of the CPU jitter entropy logic to fail due to `cc-rs` prioritizing `CFLAGS` over the flags provided by the build script.

### Description of changes: 
This change ensures that the last optimization flag provided by CFLAGS (when compiling the cpu jitter entropy logic) disables optimization: `-O0`.

### Callout
This depends upon the expectation that the environment variables are read and applied during the invocation of `get_compiler`:
* Collection of environment variable: https://github.com/rust-lang/cc-rs/blob/6fb2a8f27d10a991203d5a892fd8320056f7fe41/src/lib.rs#L1977
* Application of flags: https://github.com/rust-lang/cc-rs/blob/6fb2a8f27d10a991203d5a892fd8320056f7fe41/src/lib.rs#L2017

### Testing:
Added CI job to cover this scenario.

### DIFF
--- a/aws-lc-sys/builder/cc_builder.rs
+++ b/aws-lc-sys/builder/cc_builder.rs
@@ -16,8 +16,8 @@ mod x86_64_unknown_linux_musl;
 
 use crate::{
     cargo_env, effective_target, emit_warning, env_var_to_bool, execute_command, get_crate_cflags,
-    is_no_asm, optional_env_optional_crate_target, out_dir, requested_c_std, set_env_for_target,
-    target, target_arch, target_env, target_os, CStdRequested, OutputLibType,
+    is_no_asm, optional_env_optional_crate_target, optional_env_target, out_dir, requested_c_std,
+    set_env_for_target, target, target_arch, target_env, target_os, CStdRequested, OutputLibType,
 };
 use std::cell::Cell;
 use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


aws-lc-sys/builder/cc_builder.rs
```diff
@@ -373,7 +373,19 @@ impl CcBuilder {
...
 
-        let compiler = je_builder.get_compiler();
+        let compiler = if let Some(original_cflags) = optional_env_target("CFLAGS") {
+            let mut new_cflags = original_cflags.clone();
+            new_cflags.push_str(" -O0");
+            set_env_for_target("CFLAGS", &new_cflags);
+            // cc-rs currently prioritizes flags provided by CFLAGS over the flags provided by the build script.
+            // The environment variables used by the compiler are set when `get_compiler` is called.
+            let compiler = je_builder.get_compiler();
+            set_env_for_target("CFLAGS", &original_cflags);
```
Mark-Simulacrum, line 383: Is there any concurrency in the build script? In general `env::set_var` (what `set_env_for_target` uses internally) should not be used in concurrent programs - https://doc.rust-lang.org/nightly/std/env/fn.set_var.html#safety - it's been made `unsafe` in newer editions of Rust for that reason.

justsmth: The logic within the build-script is single-threaded. However, we do enable the "parallel" feature on `cc`.

skmcgrail (approved):

Mark-Simulacrum (approved):

samuel40791765 (approved):

PR #902, opened 2025-10-01
Adds CMAC

Elvrarin:
### Issues:
See SIM P307101963

### Description of changes: 
Currently, aws-lc-rs only supports HMAC
This commit adds CMAC, with support for AES 128, 192, 256 and TDES keys.
https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38b.pdf

### Call-outs:
Some APIs like Key::new(), sign(), Context::update(), Context::sign() may panic. This is done to be consistent with the API of HMAC

### Testing:
Unit tests and integration tests added following examples of existing HMAC tests.


Elvrarin:
I generated the test data in this commit. I will create a new commit that uses CMAC test data from aws-lc

PR #903, opened 2025-10-02
Adds CMAC

Elvrarin:
### Issues:
See SIM P307101963

### Description of changes: 
Currently, aws-lc-rs only supports HMAC
This commit adds CMAC, with support for AES 128, 192, 256 and TDES keys.
https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38b.pdf

### Call-outs:
* Some APIs like Key::new(), sign(), Context::update(), Context::sign() may panic. This is done to be consistent with the API of HMAC
* According to [aws-lc's service indicator](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/service_indicator/service_indicator.c#L147-L154), AES 192 and TDES aren't FIPS approved. (Question: which fips document?)

### Testing:
* Unit tests added.
* Integration Test data is copied from [here in aws-lc](https://github.com/aws/aws-lc/tree/main/crypto/fipsmodule/cmac)


aws-lc-rs/src/cmac.rs
```diff
@@ -0,0 +1,635 @@
...
+    /// The tag length for this CMAC algorithm.
+    #[inline]
+    #[must_use]
+    pub fn tag_len(&self) -> usize {
+        self.tag_len
+    }
+}
+
+/// CMAC using AES-128.
+pub static AES_128: Algorithm = Algorithm {
```
justsmth: Since this is a new API, we should make these all `pub const`.

aws-lc-rs/src/cmac.rs
```diff
@@ -0,0 +1,635 @@
...
+impl AsRef<[u8]> for Tag {
+    #[inline]
+    fn as_ref(&self) -> &[u8] {
+        &self.bytes[..self.len]
+    }
+}
+
+struct LcCmacCtx(*mut CMAC_CTX);
+
+impl LcCmacCtx {
```
justsmth: For consistency with the rest of the code, I'd prefer this use `LcPtr<CMAC_CTX>`. (See our implementation/use of `LcPtr<EVP_PKEY>` for reference.)

aws-lc-rs/src/cmac.rs
```diff
@@ -0,0 +1,608 @@
...
+    // # FIPS
+    // Use this type with one of the following algorithms:
+    // * `AES_128`
+    // * `AES_256`
+    //
+    /// # Errors
+    /// `error::Unspecified` if the `rng` fails or key construction fails.
+    pub fn generate(
+        algorithm: Algorithm,
+        rng: &dyn rand::SecureRandom,
```
justsmth: Remove the `rng` parameter -- a `SecureRandom` is required for `hmac::Key` (an API we inherited), but there's no need for it in aws-lc-rs.  Our library always use the same source of randomness regardless.

aws-lc-rs/src/cmac.rs
```diff
@@ -0,0 +1,608 @@
...
+    ///
+    /// # Panics
+    /// Panics if the CMAC cannot be updated.
+    #[inline]
+    pub fn update(&mut self, data: &[u8]) {
+        Self::try_update(self, data).expect("CMAC_Update failed");
+    }
+
+    #[inline]
+    fn try_update(&mut self, data: &[u8]) -> Result<(), Unspecified> {
```
justsmth: Since this is a new API, I think we should just go ahead and expose the fallibility of these APIs:
```
pub fn update(&mut self, data: &[u8]) -> Result<(), Unspecified> {
...
pub fn sign(mut self) -> Result<Tag, Unspecified> {
...
```

Failure of these API should be extremely rare, but I don't see value in hiding that possibility.

aws-lc-rs/src/cmac.rs
```diff
@@ -0,0 +1,608 @@
...
+/// It is generally not safe to implement CMAC verification by comparing the
+/// return value of `sign` to a tag. Use `verify` for verification instead.
+//
+// # FIPS
+// Use this function with one of the following algorithms:
+// * `AES_128`
+// * `AES_256`
+#[inline]
+#[must_use]
+pub fn sign(key: &Key, data: &[u8]) -> Tag {
```
justsmth: Also expose the possible failure in `sign` and `verify`, here and below.

aws-lc-rs/src/cmac.rs
```diff
@@ -0,0 +1,608 @@
...
+use core::mem::MaybeUninit;
+use core::ptr::null_mut;
+
+/// A CMAC algorithm.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct Algorithm {
+    name: &'static str,
+    key_len: usize,
+    tag_len: usize,
+    cipher_fn: unsafe extern "C" fn() -> *const crate::aws_lc::EVP_CIPHER,
```
justsmth: I'm not crazy about holding a function pointer here. I'd prefer the approach of using a (private) `AlgorithmId` type with a method that provides this value in the form of a `ConstPointer<'_, EVP_CIPHER>`, similar to what is done [here](https://github.com/aws/aws-lc-rs/blob/d4be093dee6572097f239bfabef144dd7f5a7930/aws-lc-rs/src/cipher.rs#L286-L306).

aws-lc-rs/src/cmac.rs
Elvrarin: Ok, I'll remove this parameter and use rand::fill(), similar to what's in iv.rs

aws-lc-rs/src/cmac.rs
Elvrarin: I will expose the fallibility of Key::new(), and update(), and sign(),

aws-lc-rs/src/cmac.rs
Elvrarin: I don't think it's necessary. following the cipher.rs feels like making it unnecessarily complicated. Cipher.rs supports many modes, but we'll only use the CBC mode. I think the hardcoded function pointer in pub const AES_128, AES_192, AES_256 is enough.

justsmth: I disagree. There’s an established pattern in our library of using an enum to represent “AlgorithmId”s, largely for this purpose.

Different example: https://github.com/aws/aws-lc-rs/blob/19f46c2891f97d794f11a3fac346502108cbff59/aws-lc-rs/src/digest.rs#L365

aws-lc-rs/src/cmac.rs
```diff
@@ -0,0 +1,591 @@
...
+
+impl core::fmt::Debug for Algorithm {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> Result<(), core::fmt::Error> {
+        f.debug_struct("Algorithm")
+            .field("id", &self.id)
+            .field("key_len", &self.key_len)
+            .field("tag_len", &self.tag_len)
+            .finish()
+    }
+}
```
justsmth: NP: Since all of the fields are now "Debug", can't we derive "Debug"?

aws-lc-rs/src/cmac.rs
```diff
@@ -0,0 +1,591 @@
...
+    let mut ctx = Context::with_key(key);
+    ctx.update(data)?;
+    ctx.sign()
+}
+
+/// Calculates the CMAC of `data` using the signing key `key`, and verifies
+/// whether the resultant value equals `tag`, in one step.
+///
+/// This is logically equivalent to, but more efficient than, constructing a
+/// `Key` with the same value as `key` and then using `verify`.
```
justsmth: .. but this function accepts a "Key" parameter?

aws-lc-rs/src/cmac.rs
```diff
@@ -0,0 +1,591 @@
...
+                output.as_mut_ptr(),
+                out_len.as_mut_ptr(),
+            )) {
+                return Err(Unspecified);
+            }
+            
+            let actual_len = out_len.assume_init();
+            if actual_len != self.key.algorithm.tag_len() {
+                return Err(Unspecified);
+            }
```
justsmth, line 380: If `actual_len > MAX_CMAC_TAG_LEN`, this seems worthy of a panic. What do we do in other similar situations?

aws-lc-rs/src/cmac.rs
Elvrarin: lol I copied this from hamc.rs's verify doc string. The doc string in hmac.rs is also wrong.  I'll remove this sentence in cmac.rs for now.

aws-lc-rs/tests/cmac_tests.rs
```diff
@@ -0,0 +1,139 @@
...
+        let result = test_case.consume_string("Result");
+        
+        let input = if mlen == 0 { Vec::new() } else { msg };
+        let should_pass = result.starts_with('P');
+        
+        let cmac_key = cmac::Key::new(cmac::AES_128, &key).unwrap();
+        let signature = cmac::sign(&cmac_key, &input).unwrap();
+        
+        // Truncate to tlen
+        let truncated_sig = &signature.as_ref()[..std::cmp::min(signature.as_ref().len(), tlen)];
```
justsmth: The truncation used here and in the other functions is not needed -- it would actually hide a bug were we to produce too long of a signature.

aws-lc-rs/src/cmac.rs
Elvrarin: Discussed privately: Although in aws_lc, actual_len is hardcoded to aes/des block size, but if actual_len > MAX_CMAC_TAG_LEN, this may indicate a memory corruption and deserves a panic.

aws-lc-rs/tests/cmac_tests.rs
Elvrarin: I copied these tests data, and test functions from the aws-lc. The original tests also does a truncation. Without the truncation, the tests start to fail.
https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/cmac/cmac_test.cc#L226-L227

justsmth: Ok, we should just trancate it to `tlen` then -- it should never be longer than the actual tag:
```
let truncated_sig = &signature.as_ref()[..tlen];
```

justsmth (approved):

aws-lc-rs/src/cmac.rs
```diff
@@ -0,0 +1,591 @@
...
+//!
+//! cmac::verify(&key, msg.as_bytes(), tag.as_ref())?;
+//!
+//! # Ok::<(), aws_lc_rs::error::Unspecified>(())
+//! ```
+//!
+//! ## Using the one-shot API:
+//!
+//! ```
+//! use aws_lc_rs::rand::SecureRandom;
```
skmcgrail: Not needed
```suggestion
```

aws-lc-rs/src/cmac.rs
```diff
@@ -0,0 +1,591 @@
...
+//! // integrity of the message.
+//! let v_key = cmac::Key::new(cmac::AES_128, key_value.as_ref())?;
+//! cmac::verify(&v_key, msg.as_bytes(), tag.as_ref())?;
+//!
+//! # Ok::<(), aws_lc_rs::error::Unspecified>(())
+//! ```
+//!
+//! ## Using the multi-part API:
+//! ```
+//! use aws_lc_rs::rand::SecureRandom;
```
skmcgrail: Not needed
```suggestion
```

Issue #828, opened 2025-06-19
Remove cmake dependency for Windows


stefnotch:
### Problem:

I am interested in building and contributing to projects like Deno, which depends on aws-lc-rs.
However, it seems like aws-lc-rs has a dependency on cmake for Windows builds. The nasm dependency can luckily be turned off.

Figuring out the issue, and then figuring out how to install cmake on Windows was not a pleasant experience. It was especially bad, since the project that I was contributing to (Deno) has a very slow build time. This made "installing Cmake and checking if it works" a very slow process.

### Requested solution:

It would be amazing for me if the Windows target also got the [cc treatment](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-sys/builder/cc_builder.rs). For contributing to projects, my main priority is being able to set up the development environment as easily as possible.

#### Relevant details
 * CPU architecture: x86-64 
 * OS: Windows 11

CryZe:
Reqwest just released their first release candidate for 0.13, where they turn on aws-lc-rs by default. This should be considered a high priority issue now. https://github.com/seanmonstar/reqwest/discussions/2906

commonsensesoftware:
What's preventing this from moving forward? It's been 6 months and there's been an open PR for 2 months. Is there something more to do or discuss? It seems folks are ready to help if there's a path forward.

FWIW it's causing pain. I'm using MSVC, which `cmake` supports and detects, but it's still not being picked up and `aws-lc` causes the build to fail. I can't explain why. You can see the check here:

https://github.com/rust-lang/cmake-rs/blob/511160ffce8d27632d20770b7090b9f73dda26c6/src/lib.rs#L645

`cargo tree` confirmed this is the version I'm using.

`rustup show` outputs:

```text
Default host: x86_64-pc-windows-msvc

installed toolchains
--------------------
stable-x86_64-pc-windows-msvc (active, default)
nightly-x86_64-pc-windows-msvc

active toolchain
----------------
name: stable-x86_64-pc-windows-msvc
active because: it's the default toolchain
installed targets:
  x86_64-pc-windows-msvc
```

justsmth:
> What's preventing this from moving forward? 

There's been some internal discussion about alternative/better ways to implement this. I hope to have resolution on this soon.

> FWIW it's causing pain. I'm using MSVC, which `cmake` supports and detects, but it's still not being picked up and `aws-lc` causes the build to fail.

[...]

Issue #766, opened 2025-04-03
Support `aws-lc-sys` on `x86_64-unknown-freebsd` without cmake


LHolten:
At work we would like to build `aws-lc-sys` without cmake for the `x86_64-unknown-freebsd` target.
From what I understand this is supported on a few platforms already with pre-generated bingen output.

I can *probably* make a PR if that is desired, depending on how difficult this is.

### Problem:

Reproduce the problem with `cargo build --target x86_64-unknown-freebsd` on a system without `cmake`.
Or run `AWS_LC_SYS_CMAKE_BUILDER=0 cargo build --target x86_64-unknown-freebsd`.

```text
   Compiling aws-lc-sys v0.28.0 (/home/lucasholten/opensource/aws-lc-rs/aws-lc-sys)
error: failed to run custom build command for `aws-lc-sys v0.28.0 (/home/lucasholten/opensource/aws-lc-rs/aws-lc-sys)`

Caused by:
  process didn't exit successfully: `/home/lucasholten/opensource/aws-lc-rs/target/debug/build/aws-lc-sys-b310cb966a2d0e6f/build-script-main` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=AWS_LC_SYS_NO_PREFIX
  cargo:rerun-if-env-changed=AWS_LC_SYS_PREGENERATING_BINDINGS
  cargo:rerun-if-env-changed=AWS_LC_SYS_EXTERNAL_BINDGEN
  cargo:rerun-if-env-changed=AWS_LC_SYS_NO_ASM
  cargo:rerun-if-env-changed=AWS_LC_SYS_CFLAGS
  cargo:rerun-if-env-changed=AWS_LC_SYS_PREBUILT_NASM
  cargo:rerun-if-env-changed=AWS_LC_SYS_C_STD
  cargo:rerun-if-env-changed=AWS_LC_SYS_CMAKE_BUILDER
  cargo:rerun-if-env-changed=AWS_LC_SYS_NO_PREGENERATED_SRC
  cargo:rerun-if-env-changed=AWS_LC_SYS_STATIC
  cargo:rerun-if-env-changed=CMAKE

  --- stderr
  Missing dependency: cmake

  thread 'main' panicked at aws-lc-sys/builder/main.rs:382:40:
  called `Result::unwrap()` on an `Err` value: "Required build dependency is missing. Halting build."
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

[...]

PR #911, opened 2025-10-08
Eliminate CMake Requirement

justsmth:
### Issues:
Addresses:
* #828
* #766

### Motivation
CMake creates friction for users in constrained build environments, minimal containers, or when cross-compiling. This change simplifies the build to require only a C compiler and the Rust toolchain.

### Description of Changes
This PR removes the CMake build dependency by enhancing the `cc`-based build system to directly compile AWS-LC with platform-specific source file lists.
* Replaced per-triple configs (e.g., `aarch64_unknown_linux_gnu.rs`) with a universal source list plus OS+arch-specific modules (`linux_aarch64.rs`, `win_x86_64.rs`, etc.)
* Added `nasm_builder.rs` for direct Windows assembly compilation, with prebuilt object fallback
* Added shell scripts in `aws-lc-sys/scripts/cc_builder/` to generate and maintain platform source lists
* Added MSYS2 no-CMake test coverage for Windows MinGW builds

### Call-out

The shell scripts are fragile, and likely to break as we add more assembly. Any errors related to this would occur during the linking phase of the build due to missing symbols.
* I would like to refactor this to instead leverage CMake's ability to produce "compilation databases" listing all of the source files relevant for a particular platform. But I think this refactoring can be done subsequently.

### Testing
* All existing (relevant) tests should continue to succeed.
* Added `msys2-no-cmake` test job to verify Windows MinGW builds without CMake.

### DIFF
--- a/.github/workflows/sys-bindings-generator.yml
+++ b/.github/workflows/sys-bindings-generator.yml
@@ -285,11 +285,7 @@ jobs:
       - generate-bindings-and-commit
       - generate-universal-bindings-and-commit
     if: github.repository == 'aws/aws-lc-rs'
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ ubuntu-latest, macos-15-intel, macos-latest ]
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
         with:
--- a/aws-lc-sys/builder/cc_builder.rs
+++ b/aws-lc-sys/builder/cc_builder.rs
@@ -419,7 +419,7 @@ impl CcBuilder {
             set_env_for_target("CFLAGS", &new_cflags);
             // cc-rs currently prioritizes flags provided by CFLAGS over the flags provided by the build script.
             // The environment variables used by the compiler are set when `get_compiler` is called.
-            let _compiler = je_builder.get_compiler();
+            je_builder.get_compiler();
             set_env_for_target("CFLAGS", &original_cflags);
         }
 


WillChilds-Klein:
> I would like to refactor this to instead leverage CMake's ability to produce "compilation databases" listing all of the source files relevant for a particular platform. But I think this refactoring can be done subsequently.

Does this imply that you envision using CMake as a "developer dependency" rather than a "build dependency"?

.github/workflows/compilers.yml
```diff
@@ -146,8 +146,10 @@ jobs:
       - name: Checkout
         run: |
           set -x
+          mkdir -p /tmp/aws-lc-rs
```
WillChilds-Klein, line 149: nitpick: could we isolate the well-known dir name to a higher-level variable? don't expect collisions in the future, but central definition seems preferable.

.github/workflows/sys-bindings-generator.yml
```diff
@@ -298,37 +298,15 @@ jobs:
...
-          submodules: 'recursive'
-          ref: ${{ github.ref_name }}
-      - uses: dtolnay/rust-toolchain@stable
-        with:
-          components: 'rustfmt'
-      - name: Install cross
-        run: cargo install cross --locked --git https://github.com/cross-rs/cross
-      - name: Collect source for ${{ matrix.target }}
-        run: ./scripts/build/collect_build_src.sh -t ${{ matrix.target }}
+      - name: Install ripgrep
```
WillChilds-Klein, line 297: nitpick: is `rg` really _that_ much faster for our use-case than vanilla grep to justify another build dependency? I don't have an opinion, just curious what motivated this addition.

aws-lc-sys/builder/cc_builder.rs
```diff
@@ -283,19 +284,19 @@ impl CcBuilder {
...
-                || target_os().ends_with("bsd")
-                || target_env() == "gnu"
-                || target_env() == "musl"
-            {
+            build_options.push(BuildOption::flag("-pthread"));
+            if target_os() == "linux" {
                 build_options.push(BuildOption::define("_XOPEN_SOURCE", "700"));
-                build_options.push(BuildOption::flag("-pthread"));
+            } else if target_vendor() != "apple" {
+                // Needed by illumos
```
WillChilds-Klein, line 291: is there a specific `target_vendor()` value we can match on for ilumos?

aws-lc-sys/builder/cc_builder.rs
```diff
@@ -383,38 +384,53 @@ impl CcBuilder {
...
+            if is_like_msvc {
+                new_cflags.push_str(" -Od");
+            } else {
+                new_cflags.push_str(" -O0 -Wp,-U_FORTIFY_SOURCE");
+            }
             set_env_for_target("CFLAGS", &new_cflags);
             // cc-rs currently prioritizes flags provided by CFLAGS over the flags provided by the build script.
             // The environment variables used by the compiler are set when `get_compiler` is called.
-            let compiler = je_builder.get_compiler();
+            let _compiler = je_builder.get_compiler();
```
WillChilds-Klein: clippy seems fine w/ this, so purely a style nit but `_` might be more conventional for "this is unused and preventing optimizing-away the RHS function call".

aws-lc-sys/builder/cc_builder.rs
```diff
@@ -709,8 +737,8 @@ impl crate::Builder for CcBuilder {
             return Err("CcBuilder only supports static builds".to_string());
         }
 
-        if PlatformConfig::default_for(&effective_target()).is_none() {
-            return Err(format!("Platform not supported: {}", effective_target()));
+        if target_env() == "ohos" {
```
WillChilds-Klein, line 740: presumably `ohos` is now the only CMake-only supported  target env? what happens if other CMake-only supported targets are added in the future? Similarly, what's stopping us from supporting `ohos` w/ `cc_builder.rs`?

aws-lc-sys/builder/nasm_builder.rs
```diff
@@ -0,0 +1,143 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+use std::ffi::OsString;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+use crate::{execute_command, target_arch, test_nasm_command, use_prebuilt_nasm};
+
+#[derive(Debug)]
+pub(crate) struct NasmBuilder {
```
WillChilds-Klein, line 11: my understanding is that `pub(crate)` [does not expose](https://doc.rust-lang.org/reference/visibility-and-privacy.html#r-vis.scoped.crate) this interface to dependents of aws-lc-sys. is that correct?

aws-lc-sys/builder/cc_builder.rs
```diff
@@ -709,8 +737,8 @@ impl crate::Builder for CcBuilder {
             return Err("CcBuilder only supports static builds".to_string());
         }
 
-        if PlatformConfig::default_for(&effective_target()).is_none() {
```
WillChilds-Klein, line 712: should we still have a catch-all Err case for `PlatformConfig::default_for(&effective_target()).is_none()`?

.github/workflows/sys-bindings-generator.yml
justsmth: I needed ripgrep b/c of inconsistent behavior of `sed` between Linux and Mac.

aws-lc-sys/builder/cc_builder.rs
justsmth: I'm not sure about `target_vendor()` on illumos.  However, I think this is also needed on Solaris (and BSDs?). The reason is that the header files on these platforms have strict API standards conformance, due to `-std=c99`, unless this or similar macro is defined. AWS-LC defines the `_GNU_SOURCE` macro a few places in our source code for a similar purpose for Linux.

aws-lc-sys/builder/cc_builder.rs
justsmth: I'll just remove this variable. I thought the return value was a `must_use`, requiring a variable to be declared for it, but apparently it's not.

aws-lc-sys/builder/cc_builder.rs
justsmth: > ... Similarly, what's stopping us from supporting ohos w/ cc_builder.rs?

When I was testing this, I noticed that [the ohos ("open harmony") CI job](https://github.com/aws/aws-lc-rs/blob/main/.github/workflows/cross.yml#L452-L467) was failing using "cc_builder". I didn't deep dive into it, but I'm pretty sure it'd be possible to support this target w/ cc_builder, but it's just not a priority ATM.

aws-lc-sys/builder/nasm_builder.rs
justsmth: This is only part of our build script, so it wouldn't be exposed to consumers even as just `pub`.

aws-lc-sys/builder/cc_builder.rs
justsmth: `PlatformConfig` is completely gone now. But to your point, I don't believe there are any other cases that we should return an `Err` here other than the ones currently listed.

WillChilds-Klein (approved):

.github/workflows/sys-bindings-generator.yml
```diff
@@ -285,50 +285,24 @@ jobs:
...
     runs-on: ubuntu-latest
-    env:
-      AWS_LC_RS_DISABLE_SLOW_TESTS: 1
-    strategy:
-      fail-fast: false
-      matrix:
-        target: [ aarch64-unknown-linux-gnu, i686-unknown-linux-gnu, aarch64-unknown-linux-musl, riscv64gc-unknown-linux-gnu, x86_64-unknown-linux-musl ]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v3
```
samuel40791765, line 290: Nit: Just curious, why the downgrade?

aws-lc-sys/scripts/cc_builder/.common.sh
```diff
@@ -0,0 +1,66 @@
...
+
+pub(super) const CRYPTO_LIBRARY: &[&str] = &[
+EOF
+    for FILE in "${@}"; do
+        echo "    \"${FILE}\","
+    done
+    cat << EOF
+];
+EOF
+}
```
samuel40791765, line 35: Nit: should we add a tag here that mentions auto-generated by `aws-lc-sys/scripts/cc_builder/.common.sh` or the corresponding parent script?

samuel40791765 (approved):

Issue #912, opened 2025-10-08
Build broken on macos darwin


TroyKomodo:
### Problem:

Seems to have broken between `aws-lc-sys`: `0.31.0` -> `0.32.2`

#### Relevant details
AWS-LC for Rust versions or commit: (6b1bce0...)

Root cause:
```
cargo:warning=In file included from /Users/runner/.bazel/sandbox/darwin-sandbox/4272/execroot/_main/bazel-out/darwin_arm64-opt/bin/external/+cargo_vendor+cargo_vendor__aws-lc-sys-0.32.2/_bs.cargo_runfiles/+cargo_vendor+cargo_vendor__aws-lc-sys-0.32.2/aws-lc/third_party/jitterentropy/jitterentropy-library/src/jitterentropy-base.c:33:
cargo:warning=In file included from /Users/runner/.bazel/sandbox/darwin-sandbox/4272/execroot/_main/bazel-out/darwin_arm64-opt/bin/external/+cargo_vendor+cargo_vendor__aws-lc-sys-0.32.2/_bs.cargo_runfiles/+cargo_vendor+cargo_vendor__aws-lc-sys-0.32.2/aws-lc/third_party/jitterentropy/jitterentropy-library/src/jitterentropy-gcd.h:23:
cargo:warning=In file included from /Users/runner/.bazel/sandbox/darwin-sandbox/4272/execroot/_main/bazel-out/darwin_arm64-opt/bin/external/+cargo_vendor+cargo_vendor__aws-lc-sys-0.32.2/_bs.cargo_runfiles/+cargo_vendor+cargo_vendor__aws-lc-sys-0.32.2/aws-lc/third_party/jitterentropy/jitterentropy-library/src/jitterentropy-internal.h:45:
cargo:warning=In file included from /Users/runner/.bazel/sandbox/darwin-sandbox/4272/execroot/_main/bazel-out/darwin_arm64-opt/bin/external/+cargo_vendor+cargo_vendor__aws-lc-sys-0.32.2/_bs.cargo_runfiles/+cargo_vendor+cargo_vendor__aws-lc-sys-0.32.2/aws-lc/third_party/jitterentropy/jitterentropy-library/jitterentropy.h:168:
[...]

Issue #899, opened 2025-09-29
Large performance regression 1.14.0 -> 1.14.1


ctz:
Downstream in rustls we have some simple performance tests that show a 26000% regression in full handshake performance when taking the aws-lc-rs 1.14.0 -> 1.14.1 upgrade.

This is measured in valgrind instruction-equivalents, and those go from approx 1,500,000 insns (1.14.0) to 423,000,000 insns (1.14.1).

I haven't looked into this in detail yet, but a few observations so far:

- this is a non-FIPS use, so is jitterent the right thing to be using? I thought it was a legalistic work-around for FIPS requirements about having a standalone entropy source.
- SHA3 has generally poor CPU performance compared to SHA2 (especially when SHA2 has wide CPU support, but SHA3 does not yet), so is that a good choice for a default entropy source?
- It seems the jitter entropy code has its own SHA3 code? What's up with that?
- It seems the jitter entropy code is compiled with `-O0`. Intentional?

At this point it's unclear to me whether this is "real" or specific to how valgrind measures things.

justsmth:
Starting with [AWS-LC v1.60](https://github.com/aws/aws-lc/releases/tag/v1.60.0) our entropy source for both FIPS and non-FIPS builds switched to CPU jitter. We believe this is more secure, but it can add a few milliseconds of latency when (and only when) a new process is forked.

I checked our versioning to verify that it aligns with your observation -- aws-lc-rs v1.14.0 (i.e., aws-lc-sys v0.31.0) [aligned with AWS-LC v1.59.0](https://github.com/aws/aws-lc-rs/pull/867), while aws-lc-rs v1.14.1 (i.e., aws-lc-sys v0.32.x) aligns with AWS-LC v1.61.x.)

I apologize for not calling this out clearly in our release notes. In most real-world scenarios the effect should be minimal, but it could definitely affect benchmarking.  To avoid the issue, something would need to trigger entropy collection prior to the benchmark beginning data collection.

Let us know your thoughts. We can discuss mitigations/alternatives for this if needed.

justsmth:
> It seems the jitter entropy code is compiled with -O0. Intentional?

Yeah, this is intentional. Unfortunately, it's [required by the logic](https://github.com/aws/aws-lc/blob/main/third_party/jitterentropy/jitterentropy-library/src/jitterentropy-base.c#L47): 
```
#ifdef __OPTIMIZE__
[...]

PR #922, opened 2025-10-14
Prepare aws-lc-sys v0.32.3

justsmth:
### Issues:
Addresses:
* https://github.com/aws/aws-lc-rs/issues/912
* https://github.com/aws/aws-lc-rs/issues/899

### Description of changes: 
This bumps the version of aws-lc-sys to v0.32.3 and aligns it with [AWS-LC v1.62.0](https://github.com/aws/aws-lc/releases/tag/v1.62.0).
* This release address a couple of concerns related to the introduction of CPU Jitter Entropy. 
  * CPU Jitter Entropy introduces latency of (typically) 10-50 milliseconds the first time a process requires "randomness". This latency is generally only observed once for the lifetime of a process, but some consumers may be sensitive to such latencies.
  * Some build environment might be unable to build jitter-entropy code. The code is intended to support almost all platforms, but there are some build environments that might lack the required headers or libraries.
* Adds support for an `AWS_LC_SYS_NO_JITTER_ENTROPY` environment variable. When the variable is set to a positive indicator (e.g., `1` or `y`, etc.) then CPU Jitter Entropy is not included in the build and not used at runtime.

### Call-outs:
* Manual commit:
  * https://github.com/aws/aws-lc-rs/pull/922/commits/aa4fa63f2422ff2e764f1247c048e36e3ab20c2b
  * https://github.com/aws/aws-lc-rs/pull/922/commits/912ca3b6c8e227388d69cac22ed59ca995a37492
  * https://github.com/aws/aws-lc-rs/pull/922/commits/180a7c837385562e02e9c36d71389eb52dfe1125
  * https://github.com/aws/aws-lc-rs/pull/922/commits/f6eb7f94d1068ecd0aa36128da35f447d154c290
  * https://github.com/aws/aws-lc-rs/pull/922/commits/c3d18eab275ae7062639a3d4704ab6dfba576f32
  * https://github.com/aws/aws-lc-rs/pull/922/commits/9f2897c3e55e1dd5d3b2419f091e2bc3079e9aeb
* Automated workflow: https://github.com/aws/aws-lc-rs/actions/runs/18497985201

### Testing:
Added tests to verify support for new build configurations.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -316,18 +316,10 @@ jobs:
             echo jitter-entropy not enabled, but should be.
             exit 1
           fi
-      - name: Verify latency is present
+      - name: Log the latency
         run: |
           LATENCY=$(./scripts/tests/first_random_byte_latency.rs)
           echo LATENCY: ${LATENCY}
-          if [ "$LATENCY" -lt 5 ]; then
-            echo LATENCY too low. Was jitter-entropy actually used?
-            exit 1
-          elif [ "$LATENCY" -gt 70 ]; then
-            echo LATENCY too high. Bug in jitter-entropy?
-            exit 1
-          fi
-          exit 0;
   no-jitter-entropy:
     if: github.repository_owner == 'aws'
     name: Verify jitter-entropy is not enabled


.github/workflows/analysis.yml
```diff
@@ -292,3 +292,75 @@ jobs:
...
+      - name: Verify latency is present
+        run: |
+          LATENCY=$(./scripts/tests/first_random_byte_latency.rs)
+          echo LATENCY: ${LATENCY}
+          if [ "$LATENCY" -lt 5 ]; then
+            echo LATENCY too low. Was jitter-entropy actually used?
+            exit 1
+          elif [ "$LATENCY" -gt 70 ]; then
+            echo LATENCY too high. Bug in jitter-entropy?
+            exit 1
```
skmcgrail:
```suggestion
          if [ "$LATENCY" -lt 5 ]; then
            echo $LATENCY too low. Was jitter-entropy actually used?
            exit 1
          elif [ "$LATENCY" -gt 70 ]; then
            echo $LATENCY too high. Bug in jitter-entropy?
            exit 1
```

Wasn't there an API that could be called at runtime to check this in the sys crate? This seems like less then ideal.

skmcgrail: Ah you call it in a different test, can we just remove this?

justsmth: The latency is fairly easy for consumers to observe, and the latency seems fairly consistent to be in the 10-50ms range -- we can expand this range further if we're concerned about false-failures. (I like having a "sanity check" in addition to just verifying that AWS-LC says that it's enabled.)

PR #934, opened 2025-11-05
Fix windows build when "+crt-static"

honsunrise:
### Description of changes: 

Fix the issue where Windows build incorrectly uses -MD for compilation when +crt-static is specified, mainly with the following modifications:

1. Enable CMP0091
2. Set CMAKE_MSVC_RUNTIME_LIBRARY to MultiThreaded
3. When CMAKE_TOOLCHAIN_FILE exists, also call configure_windows for it

### DIFF
--- a/aws-lc-sys/builder/cmake_builder.rs
+++ b/aws-lc-sys/builder/cmake_builder.rs
@@ -95,6 +95,22 @@ impl CmakeBuilder {
             set_env("CMAKE_GENERATOR", generator);
         }
 
+        let opt_level = cargo_env("OPT_LEVEL");
+        match opt_level.as_str() {
+            "0" => {
+                cmake_cfg.define("CMAKE_BUILD_TYPE", "debug");
+            }
+            "1" | "2" => {
+                cmake_cfg.define("CMAKE_BUILD_TYPE", "relwithdebinfo");
+            }
+            "s" | "z" => {
+                cmake_cfg.define("CMAKE_BUILD_TYPE", "minsizerel");
+            }
+            _ => {
+                cmake_cfg.define("CMAKE_BUILD_TYPE", "release");
+            }
+        }
+
         if OutputLibType::default() == OutputLibType::Dynamic {
             cmake_cfg.define("BUILD_SHARED_LIBS", "1");
         } else {
@@ -167,9 +183,11 @@ impl CmakeBuilder {
         // Allow environment to specify CMake toolchain.
         if let Some(toolchain) = optional_env_optional_crate_target("CMAKE_TOOLCHAIN_FILE") {
             set_env_for_target("CMAKE_TOOLCHAIN_FILE", toolchain);
-            if use_prebuilt_nasm() {
-                self.configure_prebuilt_nasm(&mut cmake_cfg);
+
+            if target_os() == "windows" {
+                self.configure_windows(&mut cmake_cfg);
             }
+
             return cmake_cfg;
         }
         // We only consider compiler CFLAGS when no cmake toolchain is set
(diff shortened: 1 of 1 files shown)


aws-lc-sys/builder/cmake_builder.rs
```diff
@@ -167,9 +167,13 @@ impl CmakeBuilder {
         // Allow environment to specify CMake toolchain.
         if let Some(toolchain) = optional_env_optional_crate_target("CMAKE_TOOLCHAIN_FILE") {
             set_env_for_target("CMAKE_TOOLCHAIN_FILE", toolchain);
-            if use_prebuilt_nasm() {
+
+            if target_os() == "windows" {
+                self.configure_windows(&mut cmake_cfg);
+            } else if use_prebuilt_nasm() {
                 self.configure_prebuilt_nasm(&mut cmake_cfg);
             }
```
justsmth, line 189: * There's no reason to call `configure_prebuilt_nasm` if the OS is something other than "windows".
* I believe the options that we set in `configure_windows` would take precedent over what's provided by "CMAKE_TOOLCHAIN_FILE" -- which I'm not sure we want to override.  My thinking is that anyone that sets "CMAKE_TOOLCHAIN_FILE" for a Windows build (fairly rare situation) would be doing so to force alternative  configuration values to be used for our build.

honsunrise: For instance, xwin uses CMAKE_TOOLCHAIN_FILE to override certain behaviors rather than overriding the entire logic. In practice, if one insists on fully overriding the logic, at least `cmake_cfg.static_crt(is_crt_static())` is required to completely remove the incorrect -MD.

justsmth: > For instance, xwin uses CMAKE_TOOLCHAIN_FILE to override certain behaviors rather than overriding the entire logic.

Ahh. Ok, I didn't see the connection before. Is the need to have +crt-static enforced when building using "cargo xwin"?  I'll take a closer look.

honsunrise: cargo xwin does not enforce the `+crt-static` requirement, but `--release` seems to be mandatory.

As for the error in CI, I haven't found the source of the behavior where the debug build uses `/defaultlib:msvcrt` instead of `/defaultlib:msvcrtd`. Do you have any idea?

justsmth: Hello! I created a separate PR for this and updated the cargo-xwin tests to also include +crt-static: #938

Can you confirm that PR addresses the concerns that you had? Thank!

honsunrise: Wow, the PR #938 test case **LGTM**.

However, I think `CMAKE_MSVC_RUNTIME_LIBRARY` should be changed to:
```rust
let runtime_library = match (is_crt_static(), cargo_env("DEBUG").as_str()) {
    (true, "true") => "MultiThreadedDebug",
    (true, "false") => "MultiThreaded",
    (false, "true") => "MultiThreadedDebugDLL",
    (false, "false") => "MultiThreadedDLL",
    (true, _) => {
        emit_warning(format!("Unknown DEBUG value: {}", cargo_env("DEBUG")));
        "MultiThreadedDebug"
    }
    (false, _) => {
        emit_warning(format!("Unknown DEBUG value: {}", cargo_env("DEBUG")));
        "MultiThreadedDebugDLL"
    }
};
```

But during testing, I encountered the issue I mentioned above. 

> As for the error in CI, I haven't found the source of the behavior where the debug 
> build uses `/defaultlib:msvcrt` instead of `/defaultlib:msvcrtd`. 

So in this PR, the condition for determining whether it's a Debug build has been removed. 

Maybe we can fix this issue in the future (but in the vast majority of cases, we probably don't need to link the Debug version of the CRT libraries anyway).

justsmth: I tried linking to the Debug libraries when the DEBUG environment variable was set to a positive value, but it was giving me linker errors. I can give it another shot, but I got the impression that Rust always wants to link to the release variants of the CRT.(?)

justsmth: https://github.com/rust-lang/rust/issues/39016

honsunrise: > [rust-lang/rust#39016](https://github.com/rust-lang/rust/issues/39016)

It seems there is no effective way to solve this problem at this stage.

justsmth:
Will resolve the concern using PR #938

PR #940, opened 2025-11-08
Fix: Add -Wa,--noexecstack to cc_builder for assembly files (resolves #939)

### LINKED ISSUE #939: ML-KEM Assembly Files Missing .note.GNU-stack Section
## Summary

ML-KEM native x86_64 assembly files compiled by `cc_builder` are missing `.note.GNU-stack` sections, resulting in executables with executable stacks on older Linux systems. This bypasses DEP/NX (Data Execution Prevention/No-eXecute) protections and represents a potential security flaw (not on its own, but in combination with other flaws).

## Severity

While modern Linux kernels (>= 2.6.39) default to non-executable stacks, older systems and certain embedded Linux distributions still rely on the `.note.GNU-stack` section to determine stack executability. Executables with executable stacks are vulnerable to return-oriented programming (ROP) and stack-based code injection attacks.

## Affected Versions

- aws-lc-rs >= v0.32.3 (when ML-KEM native assembly was introduced in October 2025)
- Specifically affects the `cc_builder` backend when building on x86_64 Linux

## Root Cause

**This is NOT an upstream bug.** Both mlkem-native and AWS-LC handle this correctly:

1. **mlkem-native** uses `-z noexecstack` linker flag
2. **AWS-LC's CMake build** uses `-Wa,--noexecstack` assembler flag (see [crypto/CMakeLists.txt:77](https://github.com/aws/aws-lc/blob/main/crypto/CMakeLists.txt#L77))
3. **aws-lc-rs `cmake_builder`** passes flags through to CMake → Works correctly ✓
4. **aws-lc-rs `cc_builder`** does NOT pass `-Wa,--noexecstack` → **Bug** ✗

The ML-KEM assembly files rely on the **build system** to add `.note.GNU-stack` sections via assembler flags, rather than having them hardcoded in the source files (unlike s2n-bignum files which include them explicitly).

## Affected Files

All 12 ML-KEM x86_64 assembly files in `crypto/fipsmodule/ml_kem/mlkem/native/x86_64/src/`:
- `intt.S`
- `mulcache_compute.S`
- `ntt.S`
- `nttfrombytes.S`
- `ntttobytes.S`
- `nttunpack.S`
- `polyvec_basemul_acc_montgomery_cached_asm_k2.S`
- `polyvec_basemul_acc_montgomery_cached_asm_k3.S`
- `polyvec_basemul_acc_montgomery_cached_asm_k4.S`
- `reduce.S`
- `rej_uniform_asm.S`
- `tomont.S`

### On x86_64 Linux:

```bash
# Build aws-lc-rs
cargo build --release -p aws-lc-sys

# Check object file for .note.GNU-stack section
readelf -S $(find target/release/build/aws-lc-sys-*/out -name "*nttfrombytes.o" | head -1) | grep "\.note\.GNU-stack"

# If section is missing, the stack will be executable
```

### Automated verification:

A complete verification script is available that:
1. Builds BEFORE the fix and confirms the bug
2. Applies the fix
3. Rebuilds and confirms the fix works

See: [`verify-execstack-fix.sh`](https://gist.github.com/psifertex/d8caca2f04bdbdfc3d92af33d5bc02a1)

## Solution

Add the `-Wa,--noexecstack` assembler flag in `aws-lc-sys/builder/cc_builder.rs` to match AWS-LC's CMake build behavior.

### Fix Implementation

In `aws-lc-sys/builder/cc_builder.rs`, add to the `prepare_builder()` function:

```rust
pub fn prepare_builder(&self) -> cc::Build {
    let mut cc_build = self.create_builder();
    let (_, build_options) = self.collect_universal_build_options(&cc_build);
    for option in build_options {
        option.apply_cc(&mut cc_build);
    }
    let cflags = get_crate_cflags();
    if !cflags.is_empty() {
        set_env_for_target("CFLAGS", cflags);
    }

    // Add --noexecstack flag for assembly files to prevent executable stacks
    // This matches the behavior of AWS-LC's CMake build which uses -Wa,--noexecstack
    // See: https://github.com/aws/aws-lc/blob/main/crypto/CMakeLists.txt#L77
    if target_os() == "linux" || target_os().ends_with("bsd") {
        cc_build.asm_flag("-Wa,--noexecstack");
    }

    cc_build
}
```

The `asm_flag()` method is available in `cc` crate >= 1.0.77 (added November 2022), and aws-lc-rs currently uses `cc` 1.2.26.

## Impact

**Who is affected:**
- Projects using aws-lc-rs v0.32.3+ on x86_64 Linux
- Particularly systems with older kernels or embedded Linux distributions
- Only affects builds using `cc_builder` (not `cmake_builder`)

**Security impact:**
- Executable stacks enable exploitation techniques that bypass modern security protections
- ROP attacks become easier to execute
- Stack-based code injection is possible

**Modern systems:**
- Recent Linux kernels ignore this and use non-executable stacks by default
- However, relying on OS defaults rather than explicit markers is poor security practice

## References

- [AWS-LC CMakeLists.txt](https://github.com/aws/aws-lc/blob/main/crypto/CMakeLists.txt#L77) - Shows correct usage of `-Wa,--noexecstack`
- [mlkem-native Issue #590](https://github.com/pq-code-package/mlkem-native/issues/590) - Similar issue, resolved with linker flags
- [mlkem-native PR #592](https://github.com/pq-code-package/mlkem-native/pull/592) - Fix implementation
- [Hardened GNU stack quickstart](https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart)
- [cc-rs asm_flag method](https://docs.rs/cc/latest/cc/struct.Build.html#method.asm_flag)
- [cc-rs PR #752](https://github.com/rust-lang/cc-rs/pull/752) - Added `asm_flag()` support

## Recommended Action

Apply the fix in the next patch release and consider backporting to v0.32.x releases since this is a security issue.

psifertex:
## Summary

Fixes missing `.note.GNU-stack` sections in ML-KEM assembly object files when using `cc_builder`, which causes executable stacks on older Linux systems.

## Problem

ML-KEM native x86_64 assembly files compiled by `cc_builder` are missing `.note.GNU-stack` sections. On older Linux systems, this results in executable stacks, bypassing DEP/NX protections.

## Root Cause

- AWS-LC's CMake build uses `-Wa,--noexecstack` ([crypto/CMakeLists.txt:77](https://github.com/aws/aws-lc/blob/main/crypto/CMakeLists.txt#L77))
- `cmake_builder` passes this through correctly ✓
- `cc_builder` was missing this flag ✗

ML-KEM assembly files rely on the build system to add `.note.GNU-stack` sections via assembler flags (unlike s2n-bignum which has them in source).

## Solution

Add `-Wa,--noexecstack` using `cc-rs`'s `asm_flag()` method in `prepare_builder()`.

## Verification

Tested on x86_64 Linux:

**Before fix:**
```
readelf -S nttfrombytes.o | grep "\.note\.GNU-stack"
(no output - section missing)
```

**After fix:**
```
readelf -S nttfrombytes.o | grep "\.note\.GNU-stack"
[17] .note.GNU-stack   PROGBITS        0000000000000000  00000140
```

Full verification script available: [verify-execstack-fix.sh](https://gist.github.com/psifertex/d8caca2f04bdbdfc3d92af33d5bc02a1)

## Impact

- **Affected:** aws-lc-rs >= v0.32.3 on x86_64 Linux using `cc_builder`
- **Severity:** Medium - Older Linux systems get executable stacks
- **Modern systems:** Unaffected (kernel defaults to non-executable stacks)

## References

- mlkem-native handled this with linker flags: [Issue #590](https://github.com/pq-code-package/mlkem-native/issues/590), [PR #592](https://github.com/pq-code-package/mlkem-native/pull/592)
- [Hardened GNU stack quickstart](https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart)

### DIFF
--- a/aws-lc-sys/builder/cc_builder.rs
+++ b/aws-lc-sys/builder/cc_builder.rs
@@ -354,6 +354,14 @@ impl CcBuilder {
         if !cflags.is_empty() {
             set_env_for_target("CFLAGS", cflags);
         }
+
+        // Add --noexecstack flag for assembly files to prevent executable stacks
+        // This matches the behavior of AWS-LC's CMake build which uses -Wa,--noexecstack
+        // See: https://github.com/aws/aws-lc/blob/main/crypto/CMakeLists.txt#L77
+        if target_os() == "linux" || target_os().ends_with("bsd") {
+            cc_build.asm_flag("-Wa,--noexecstack");
+        }
+
         cc_build
     }
 


justsmth:
Thanks for the PR!  We require that contributions be licensed the same as our library, and this is enforced by our CI.  If you agree, would you add the following statement to the PR description, and I'll re-run the CI job:

justsmth:
I verified that with this change, all object files contain the required section:
```
❯ for obj_file in $(find target -name "*.o"); do if readelf -S $obj_file | grep "\.note\.GNU-stack" > /dev/null; then x=1; else echo Not found in $obj_file; fi; done

❯ 
```

psifertex:

aws-lc-sys/builder/cc_builder.rs
```diff
@@ -354,6 +354,14 @@ impl CcBuilder {
         if !cflags.is_empty() {
             set_env_for_target("CFLAGS", cflags);
         }
+
+        // Add --noexecstack flag for assembly files to prevent executable stacks
+        // This matches the behavior of AWS-LC's CMake build which uses -Wa,--noexecstack
+        // See: https://github.com/aws/aws-lc/blob/main/crypto/CMakeLists.txt#L77
+        if target_os() == "linux" || target_os().ends_with("bsd") {
+            cc_build.asm_flag("-Wa,--noexecstack");
```
justsmth, line 362: I suspect this option would be appropriate (although possibly unnecessary) for any GNU or Clang compiler, regardless of the target OS.

justsmth: Ok, it appears that this option is primarily for platforms using ELF.

PR #942, opened 2025-11-10
CI to verify GNU-stack section is present

justsmth:
### Issues:
Addresses #939

### Description of changes: 
Verifies that all ELF object files contain a `.note.GNU-stack` section.

### Call-outs:
The `*/bcm_c_generated_asm.dir/bcm.c.o` is actually an archive of assembly files that is passed to the delocator for parsing. A failure on this file is ignored.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -362,8 +362,14 @@ jobs:
 
   gnu-stack-check:
     if: github.repository_owner == 'aws'
-    name: Verify GNU-stack section in object files
-    runs-on: ubuntu-latest
+    name: Verify (${{ matrix.host }}) GNU-stack section
+    runs-on: ${{ matrix.host }}
+    strategy:
+      fail-fast: false
+      matrix:
+        host:
+          - ubuntu-latest
+          - ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v3
         with:


.github/workflows/analysis.yml
```diff
@@ -359,3 +359,48 @@ jobs:
...
+      - name: Build project
+        run: cargo build --workspace --all-targets --release
+      - name: Check for GNU-stack section in object files
+        run: |
+          echo "Checking for .note.GNU-stack section in all object files..."
+          MISSING_COUNT=0
+          TOTAL_COUNT=0
+          for obj_file in $(find target -name "*.o"); do
+            # Skip known non-ELF object files
+            if [[ "$obj_file" == *"bcm_c_generated_asm.dir/bcm.c.o" ]]; then
```
WillChilds-Klein, line 389: you could make this more general by testing with `file $obj_file | grep ELF`

```
$ find . -name bcm.c.o -exec file {} \;
./cmake-build-debug-coverage/crypto/fipsmodule/CMakeFiles/fipsmodule.dir/bcm.c.o: ELF 64-bit LSB relocatable, x86-64, version 1 (SYSV), with debug_info, not stripped
./scratch/amazon-corretto-crypto-provider/build/awslc/build/crypto/fipsmodule/CMakeFiles/bcm_c_generated_asm.dir/bcm.c.o: assembler source, ASCII text
./cmake-build-debug/crypto/fipsmodule/CMakeFiles/fipsmodule.dir/bcm.c.o: ELF 64-bit LSB relocatable, x86-64, version 1 (SYSV), with debug_info, not stripped
./build/crypto/fipsmodule/CMakeFiles/fipsmodule.dir/bcm.c.o: ELF 64-bit LSB relocatable, x86-64, version 1 (SYSV), with debug_info, not stripped
./PYTHON_BUILD_ROOT/aws-lc-build/crypto/fipsmodule/CMakeFiles/fipsmodule.dir/bcm.c.o: ELF 64-bit LSB relocatable, x86-64, version 1 (SYSV), not stripped
```

.github/workflows/analysis.yml
```diff
@@ -359,3 +359,48 @@ jobs:
           else
             exit 0
           fi
+
+  gnu-stack-check:
+    if: github.repository_owner == 'aws'
+    name: Verify GNU-stack section in object files
+    runs-on: ubuntu-latest
```
WillChilds-Klein: should we pin this to a specific LTS? are there other platforms we want to test here?

.github/workflows/analysis.yml
justsmth: Having a "*.o" that's not ELF should never happen on Linux. The only reason it happens in our FIPS build is that we "bend over backwards" to get an assembly file embedded into a "*.a".

.github/workflows/analysis.yml
justsmth: The concern is primarily platforms using ELF for which we provide assembly code. I'll add an arm64 platform for testing as well.

.github/workflows/analysis.yml
```diff
@@ -359,3 +359,48 @@ jobs:
...
+        run: cargo build --workspace --all-targets --release
+      - name: Check for GNU-stack section in object files
+        run: |
+          echo "Checking for .note.GNU-stack section in all object files..."
+          MISSING_COUNT=0
+          TOTAL_COUNT=0
+          for obj_file in $(find target -name "*.o"); do
+            # Skip known non-ELF object files
+            if [[ "$obj_file" == *"bcm_c_generated_asm.dir/bcm.c.o" ]]; then
+              echo "⊘ Skipping (non-ELF): $obj_file"
```
torben-hansen, line 390: emojification?

justsmth: Yeah, my friend Claude likes emojis. 🤣

PR #947, opened 2025-11-12
Make aws-lc-fips-sys's config CMAKE_BUILD_TYPE logic same with aws-lc-sys

honsunrise:
### Description of changes: 

Define `CMAKE_BUILD_TYPE` so that CMake can use the correct CMAKE_XXX_FLAGS_XXX.

### DIFF
--- a/aws-lc-fips-sys/builder/cmake_builder.rs
+++ b/aws-lc-fips-sys/builder/cmake_builder.rs
@@ -121,46 +121,37 @@ impl CmakeBuilder {
         }
 
         let cc_build = cc::Build::new();
-        let opt_level = cargo_env("OPT_LEVEL");
-        if opt_level.ne("0") {
-            if opt_level.eq("1") || opt_level.eq("2") {
-                cmake_cfg.define("CMAKE_BUILD_TYPE", "relwithdebinfo");
-            } else {
-                if opt_level.eq("s") || opt_level.eq("z") {
-                    cmake_cfg.define("CMAKE_BUILD_TYPE", "minsizerel");
-                } else {
-                    cmake_cfg.define("CMAKE_BUILD_TYPE", "release");
-                }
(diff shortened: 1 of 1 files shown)


justsmth:
Hello!  Thanks for the PR. For `CMAKE_BUILD_TYPE`, we defer to [the logic provided by cmake-rs here](https://github.com/rust-lang/cmake-rs/blob/b59be3b5cce397de4018aacd51501f925ce22ecd/src/lib.rs#L122-L178).

Is there a particular problem that you're trying to address with this change?

honsunrise:
> Hello! Thanks for the PR. For `CMAKE_BUILD_TYPE`, we defer to [the logic provided by cmake-rs here](https://github.com/rust-lang/cmake-rs/blob/b59be3b5cce397de4018aacd51501f925ce22ecd/src/lib.rs#L122-L178).
> 
> Is there a particular problem that you're trying to address with this change?

Maintain consistency with the logic in `aws-lc-fips-sys`, so that the CMAKE_BUILD_TYPE has a consistent definition under the same build profile (it seems unnecessary on Windows to require `aws-lc-sys` to follow "relwithdebinfo"). Of course, if this is not mandatory, following the logic of cmake-rs would seem more elegant.

However, another approach would be for aws-lc-fips-sys to migrate to the cmake-rs logic and handle the Windows case separately, which also seems better. If this is acceptable, I can modify this MR.

justsmth:
> However, another approach would be for aws-lc-fips-sys to migrate to the cmake-rs logic and handle the Windows case separately, which also seems better. If this is acceptable, I can modify this MR.

Yeah, that logic in aws-lc-fips-sys is really only needed on Windows due to our FIPS build on Windows not supporting debug.

justsmth:
We are working in this PR #995 to unify the build scripts for the two -sys crates, so this won't be needed.

PR #951, opened 2025-11-15
Uncap HKDF salt length restriction

skmcgrail:
### Description of changes: 
Remove the salt length restriction for HKDF operations by moving the salt bytes field to be a Box'd value.

### DIFF
--- a/aws-lc-rs/src/hkdf.rs
+++ b/aws-lc-rs/src/hkdf.rs
@@ -70,10 +70,6 @@ pub const HKDF_SHA384: Algorithm = Algorithm(hmac::HMAC_SHA384);
 /// HKDF using HMAC-SHA-512.
 pub const HKDF_SHA512: Algorithm = Algorithm(hmac::HMAC_SHA512);
 
-/// General Salt length's for HKDF don't normally exceed 256 bits.
-/// We set the limit to something tolerable, so that the Salt structure can be stack allocatable.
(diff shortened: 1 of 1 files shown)


torben-hansen (approved):

aws-lc-rs/src/hkdf.rs
```diff
@@ -128,24 +124,16 @@ impl Salt {
...
-        }
-        let mut salt_bytes = [0u8; MAX_HKDF_SALT_LEN];
-        salt_bytes[0..salt_len].copy_from_slice(value);
-        Ok(Self {
-            algorithm,
-            bytes: salt_bytes,
-            len: salt_len,
-        })
+        let bytes = value.to_vec().into_boxed_slice();
+        Ok(Self { algorithm, bytes })
```
justsmth: I don't think `try_new` is needed now that we have no effective limit, we can just move this implementation into `new`.

aws-lc-rs/src/hkdf.rs
```diff
@@ -107,6 +102,7 @@ impl fmt::Debug for Salt {
 
 impl Drop for Salt {
     fn drop(&mut self) {
+        // Box<[u8]> implements Zeroize, so we can call it directly
         self.bytes.zeroize();
```
justsmth, line 106: NP: I think the Salt is a public value, so I'm not sure this is necessary.

torben-hansen (approved):

PR #954, opened 2025-11-19
Explicit smart pointer conversion

justsmth:
### Description of changes: 
We previously have been using the `Deref` crate as 'syntactic sugar" for converting smart pointers into raw pointers. This unnecessarily obfuscates what's happening. This change adds functions so that smart pointer conversions are explicit, which improves code readability.

### Call-out
* Moved `as_mut_unsafe_ptr` so that it's only available for `EVP_PKEY`.

### Testing:
Added a valgrind job to our CI tests.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -381,6 +381,10 @@ jobs:
         run: |
           CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="valgrind --error-exitcode=1 --leak-check=full --show-leak-kinds=all --suppressions=$(pwd)/.valgrind/rust-test.supp" \
             cargo test --release -p aws-lc-rs --features unstable --all-targets -- --test-threads=1
+      - name: Run FIPS release tests under Valgrind
+        run: |
+          CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="valgrind --error-exitcode=1 --leak-check=full --show-leak-kinds=all --suppressions=$(pwd)/.valgrind/rust-test.supp" \
+            cargo test --release -p aws-lc-rs --features fips --all-targets -- --test-threads=1
 
   gnu-stack-check:
     if: github.repository_owner == 'aws'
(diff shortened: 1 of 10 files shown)


.github/workflows/analysis.yml
```diff
@@ -360,6 +360,32 @@ jobs:
             exit 0
           fi
 
+  valgrind:
+    if: github.repository_owner == 'aws'
+    name: Valgrind Memory Check
+    runs-on: ubuntu-latest
+    env:
+      AWS_LC_RS_DISABLE_SLOW_TESTS: 1
```
samuel40791765, line 368: Is this redundant if we're already defining it in the bash script?

.github/workflows/analysis.yml
```diff
@@ -360,6 +360,32 @@ jobs:
...
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Install Valgrind
+        run: sudo apt-get update && sudo apt-get install -y valgrind
+      - name: Run debug tests under Valgrind
+        run: |
+          CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="valgrind --error-exitcode=1 --leak-check=full --show-leak-kinds=all --suppressions=$(pwd)/.valgrind/rust-test.supp" \
+            cargo test -p aws-lc-rs --features unstable --all-targets -- --test-threads=1
```
samuel40791765, line 379: Can we run this on multiple threads? Not sure how `TEST_THREADS` in the bash script is actually configured or used

.github/workflows/analysis.yml
justsmth: The script is for use in a development environment. It's not used by the CI job.

.github/workflows/analysis.yml
justsmth: Using multiple threads requires more Valgrind suppression rules, which I'm trying to minimize. The types of memory problems I'm hoping to surface could be observable using a single thread.

PR #957, opened 2025-11-20
Add zeroization for cmac tag

Elvrarin:
### Description of changes: 
Adding a drop implementation for cmac tag that zeroize it.

### Testing:
cargo test, make clippy, make format, all passes

### DIFF
--- a/aws-lc-rs/src/cmac.rs
+++ b/aws-lc-rs/src/cmac.rs
@@ -101,6 +101,7 @@ use crate::ptr::{ConstPointer, LcPtr};
 use crate::{constant_time, rand};
 use core::mem::MaybeUninit;
 use core::ptr::null_mut;
+use zeroize::Zeroize;
 
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 enum AlgorithmId {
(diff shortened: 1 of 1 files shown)


aws-lc-rs/src/cmac.rs
```diff
@@ -182,7 +183,7 @@ const MAX_CMAC_TAG_LEN: usize = 16;
 /// A CMAC tag.
 ///
 /// For a given tag `t`, use `t.as_ref()` to get the tag value as a byte slice.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Debug)]
```
justsmth, line 186: Removing `Copy` breaks sem-ver.

Elvrarin: In TR31 usage. CMAC is used to derive a encryption key. Therefore Tag has to be treated as a secret value here

ctz: As a general point, a MAC's tag is never secret. I would say you want a different, additional interface to this one -- some sort of CMAC-based KDF?

fabrice102:
For completeness, [`sign_to_buffer`](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/cmac/fn.sign_to_buffer.html) can now be used to allow zeroizing the CMAC tag (by using `sign_to_buffer` to write the tag to a zeroizable buffer, instead of directly using [`sign`](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/cmac/fn.sign.html)).

Issue #858, opened 2025-08-13
Build script fails when using bazel


TroyKomodo:
### Problem:

I am using bazel to build aws-lc-rs however we run into issues with the linker not being found, since I am using a entirely freestanding toolchain with no host system dependencies we have the fo

#### Relevant details

latest version

`Linux 6.12.35 #1-NixOS SMP PREEMPT_DYNAMIC Fri Jun 27 10:11:46 UTC 2025 x86_64 unknown`

Build log:
 * The log tells compiler and version.
 
```text
Build script process failed with exit code 101
--stdout:
cargo:rerun-if-env-changed=AWS_LC_SYS_NO_PREFIX_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=AWS_LC_SYS_NO_PREFIX
cargo:rerun-if-env-changed=AWS_LC_SYS_PREGENERATING_BINDINGS_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=AWS_LC_SYS_PREGENERATING_BINDINGS
cargo:rerun-if-env-changed=AWS_LC_SYS_EXTERNAL_BINDGEN_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=AWS_LC_SYS_EXTERNAL_BINDGEN
cargo:rerun-if-env-changed=AWS_LC_SYS_NO_ASM_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=AWS_LC_SYS_NO_ASM
cargo:rerun-if-env-changed=AWS_LC_SYS_CFLAGS_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=AWS_LC_SYS_CFLAGS
cargo:rerun-if-env-changed=CFLAGS_x86_64_unknown_linux_gnu
cargo:rerun-if-env-changed=CFLAGS
[...]

PR #958, opened 2025-11-21
Respect LDFLAGS in memcmp_check

justsmth:
** The author of this is @aaronmondal

This check invokes the compiled executable and hence needs to link it. CMake handles this via LDFLAGS but `cc` doesn't. In setups with custom linker setups this could lead to a mismatch between the expected and the actually used linker. Explicitly respecting LDFLAGS here brings us back to parity with CMake.

Fixes https://github.com/aws/aws-lc-rs/issues/858 whose resolution is not applicable if ignoring compiler warnings is not an option.

### DIFF
--- a/aws-lc-sys/builder/cc_builder.rs
+++ b/aws-lc-sys/builder/cc_builder.rs
@@ -582,6 +582,18 @@ impl CcBuilder {
             return;
         }
         let mut memcmp_compile_args = Vec::from(memcmp_compiler.args());
+
+        // This check invokes the compiled executable and hence needs to link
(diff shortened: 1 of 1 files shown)


justsmth:
-- I'd like to call out that this only respects LDFLAGS for the compilation of `memcmp_invalid_stripped_check` which is an executable that (except when cross-compiling) our build runs to verify sanity of the compiler w/ `memcmp`.  

Thanks to @cerisier for pointing me to @aaronmondal fix. Also thanks to @TroyKomodo for raising this issue -- my apologies for not catching on sooner. 😀

WillChilds-Klein (approved):

skmcgrail (approved):

PR #968, opened 2025-12-08
CMAC: `sign_to_buffer` and `Context::verify`

justsmth:
### Description of changes: 
Adds functions `cmac::sign_to_buffer` and `cmac::Context::verify`:
* `cmac::sign_to_buffer` allows the caller to provide their own buffer into which the tag will be written.
* `cmac::Context::verify` allows for verification of multi-part data.

### Testing:
Updated/expanded testing around cmac.

### DIFF
--- a/aws-lc-rs/src/cmac.rs
+++ b/aws-lc-rs/src/cmac.rs
@@ -501,7 +501,7 @@ pub fn verify(key: &Key, data: &[u8], tag: &[u8]) -> Result<(), Unspecified> {
         result.len()
     };
 
-    constant_time::verify_slices_are_equal(&output[0..output_len], &tag)
+    constant_time::verify_slices_are_equal(&output[0..output_len], tag)
 }
 
 #[cfg(test)]


aws-lc-rs/src/cmac.rs
```diff
@@ -358,33 +358,71 @@ impl Context {
...
-            let actual_len = out_len.assume_init();
-            // This indicates a memory corruption.
-            assert!(
-                actual_len <= MAX_CMAC_TAG_LEN,
-                "CMAC tag length {actual_len} exceeds maximum {MAX_CMAC_TAG_LEN}"
-            );
-            if actual_len != self.key.algorithm.tag_len() {
-                return Err(Unspecified);
-            }
+        constant_time::verify_slices_are_equal(&output[0..output_len], tag)
```
WillChilds-Klein, line 395:
```suggestion
        constant_time::verify_slices_are_equal(&output[0..output_len], &tag[0..output_len])
```

should we slice the tag to `output_len` too?

aws-lc-rs/src/cmac.rs
```diff
@@ -408,22 +446,62 @@ pub fn sign(key: &Key, data: &[u8]) -> Result<Tag, Unspecified> {
...
 pub fn verify(key: &Key, data: &[u8], tag: &[u8]) -> Result<(), Unspecified> {
-    let computed_tag = sign(key, data)?;
-    constant_time::verify_slices_are_equal(computed_tag.as_ref(), tag)
+    let mut output = [0u8; MAX_CMAC_TAG_LEN];
+    let output_len = {
+        let result = sign_to_buffer(key, data, &mut output)?;
+        result.len()
+    };
+
+    constant_time::verify_slices_are_equal(&output[0..output_len], tag)
```
WillChilds-Klein, line 504: same question about slicing tag

aws-lc-rs/src/cmac.rs
```diff
@@ -586,4 +664,95 @@ mod tests {
         assert_eq!(tag.as_ref().len(), 8); // 3DES block size
         assert!(verify(&key, data, tag.as_ref()).is_ok());
     }
+
+    #[test]
+    fn cmac_sign_to_buffer_test() {
```
WillChilds-Klein, line 669: why put these tests here instead of cmac_tests.rs? doesn't look like they directly call any crate-internal functions

aws-lc-rs/src/cmac.rs
justsmth: I think that would mask a problem. If the tag the caller provides is the wrong length, we would want this operation to fail.

aws-lc-rs/src/cmac.rs
justsmth: I like having a few tests living in the same file as the implementation, mostly for the convenience. And generally, I prefer it when we have decent coverage from both unit and integration tests.

PR #971, opened 2025-12-13
update cmake-rs version to support vs2026

nimeng1299:
### Issues:
cmake-rs supports vs18 (2026) from 0.1.55 [cmake-rs Add Visual Studio 2026 support](https://github.com/rust-lang/cmake-rs/pull/255)

### Description of changes: 
update cmake version 0.1.54 -> 0.1.55

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ resolver = "2"
 bindgen = "0.72.0"
 cc = "1.2.26"
 clap = "=4.4.18"# v4.5 is MSRV 1.74
-cmake = "0.1.54"
+cmake = "0.1.55"
 criterion = "0.5.1" # v0.6.0 depends on clap v4.5
 dunce = "1.0.5"
 fs_extra = "1.3.0"


Cargo.toml
```diff
@@ -19,7 +19,7 @@ resolver = "2"
 bindgen = "0.72.0"
 cc = "1.2.26"
 clap = "=4.4.18"# v4.5 is MSRV 1.74
-cmake = "0.1.54"
+cmake = "0.1.55"
```
justsmth, line 22: cmake v0.1.55 broke support for CMake v3.12 and earlier.

With our current configuration (`cmake = "0.1.54"`), consumers are still free to use v0.1.55 (or later). Making this change would only prevent consumers from continuing to use v0.1.54, which is not necessary.

justsmth:
I'd be ok with bumping [cmake to v0.1.57](https://github.com/rust-lang/cmake-rs/releases). There appear to be some important bug fixes in the last couple of releases.

PR #973, opened 2025-12-15
More robust rustls integ test

justsmth:
### Description of changes: 
Use a `[patch.crates-io]` to override dependency on aws-lc-rs to use local workspace.


justsmth:
The following failure in the rustls integ tests are unrelated to aws-lc-rs:
```
    Updating crates.io index
error: failed to select a version for the requirement `libcrux-sha3 = "^0.0.3"`
  version 0.0.3 is yanked
location searched: crates.io index
required by package `hpke-rs v0.3.0`
    ... which satisfies dependency `hpke-rs = "^0.3"` of package `rustls-provider-example v0.0.1 (/private/var/folders/cm/7x0txmbs1bd_fn4jvfh2gqzr0000gn/T/tmp.yBgThfzWu5/provider-example)`
```

aws-lc-rs/scripts/run-rustls-integration.sh
```diff
@@ -31,157 +35,145 @@ function cleanup() {
...
-    ][0].num
-  ')"
+# Clone a repository and optionally checkout a specific commit
+# Usage: clone_repo <url> <destination> [commit]
+function clone_repo() {
+  local url="$1"
+  local dest="$2"
+  local commit="${3:-}"
+
+  git clone "$url" "$dest"
```
WillChilds-Klein, line 102: presumably there are no submodules to init?

aws-lc-rs/scripts/run-rustls-integration.sh
```diff
@@ -31,157 +35,145 @@ function cleanup() {
...
+  if [ -n "$commit" ]; then
```
WillChilds-Klein, line 103: what about the else case? should we fail with an indicative message?

aws-lc-rs/scripts/run-rustls-integration.sh
```diff
@@ -31,157 +35,145 @@ function cleanup() {
...
-
-# Print the dependency tree for debug purposes, if we did everything right there
-# should only be one aws-lc-rs version. Otherwise this will fail sine there are multiple versions
 cargo tree -i aws-lc-rs --features aws-lc-rs
-# Run the rustls tests with the aws-lc-rs feature enabled
 cargo test --features aws-lc-rs
-popd &>/dev/null # ./rustls
-popd &>/dev/null # ${RUSTLS_DIR}
+popd > /dev/null
+popd > /dev/null
```
WillChilds-Klein, line 177: nit: retain comments showing which dir you're popping out of

aws-lc-rs/scripts/run-rustls-integration.sh
justsmth: I don't think there are any, but it might be good to add `--recurse-submodules` to make it more robust were this to change.

aws-lc-rs/scripts/run-rustls-integration.sh
justsmth: If these third parameter ("commit") is not provided, then we don't need to align onto a specific commit -- we instead just take the latest commit on the default branch.

justsmth:
Feedback addressed in #979

PR #987, opened 2026-01-02
[DRAFT] Panic on any FIPS-relevant cryptographic operation

justsmth:
### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 
Describe current behavior and how your code changes that behavior. If there are no issues this pr is resolving, explain why this change is necessary.

### Call-outs:
Point out areas that need special attention or support during the review process. Discuss architecture or design changes.

### Testing:
How is this change tested (unit tests, fuzz tests, etc.)? Are there any testing steps to be verified by the reviewer?

### DIFF
--- a/aws-lc-rs/Cargo.toml
+++ b/aws-lc-rs/Cargo.toml
@@ -38,6 +38,7 @@ asan = ["aws-lc-sys?/asan", "aws-lc-fips-sys?/asan"]
 test_logging = []
 unstable = []
 prebuilt-nasm = ["aws-lc-sys?/prebuilt-nasm"]
+panic_on_fips = []
 
 # require non-FIPS
 non-fips = ["aws-lc-sys"]
--- a/aws-lc-rs/src/ec.rs
+++ b/aws-lc-rs/src/ec.rs
@@ -71,12 +71,12 @@ pub(crate) fn validate_ec_evp_key(
     })?;
     verify_ec_key_nid(&ec_key, expected_curve_nid)?;
 
-    #[cfg(not(feature = "fips"))]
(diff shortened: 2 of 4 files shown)


aws-lc-rs/Cargo.toml
```diff
@@ -38,6 +38,7 @@ asan = ["aws-lc-sys?/asan", "aws-lc-fips-sys?/asan"]
 test_logging = []
 unstable = []
 prebuilt-nasm = ["aws-lc-sys?/prebuilt-nasm"]
+panic_on_fips = []
```
jlizen, line 41: This should almost certainly be a cfg flag rather than a feature, since features are transitive? I don't think we want to let dependencies enable the "panic on fips" mode in binaries?

justsmth: So this would be needed when there are two (or more) versions of aws-lc-rs resolved to -- one is desired, the others are not. So we'd need to ensure that the one desired version remains unaffected.

One option might be to check some environment variable in the aws-lc-rs build script, like
`AWS_LC_RS_PANIC_ON_FIPS_UNLESS_VERSION=1.11.1`, and emit a `cargo:rustc-cfg=panic_on_fips` only if `CARGO_PKG_VERSION != AWS_LC_RS_PANIC_ON_FIPS_UNLESS_VERSION`.

jlizen: I was assuming it would be something like, `--cfg aws-lc-panic-unless-version="1.11"` (and presumably patch version bumps are still ok, idk).

But ENV works too, anyway main point is, seems sketchy as a (transitive from dependencies) cargo feature flag.

Issue #990, opened 2026-01-06
StreamingDecryptingKey and StreamingEncryptingKey need useless extra block in output


bjorn3:
### Problem:

`StreamingDecryptingKey` and `StreamingEncryptingKey` require the output buffer size to be input buffer size + block length. Even when the caller ensures that the input buffer is always a multiple of the block length and thus no extra space in the output buffer would be necessary for storing a partial block. This requires needlessly zeroing extra space in the output buffer using `decrypted_buf.resize(buf.len() + block_len, 0);` and then using it like `decrypting_key.update(&self.buf[block_len..packet_length], &mut self.decrypted_buf[block_len..packet_length + block_len])`.

### Solution:

Three options I can think of are:
* Remove the requirement for the extra block in the output buffer in `update` when all previous `update` calls accepted input buffers that are a multiple of the block length.
* Add a dedicated method which requires the input buffer to be a multiple of the block length and doesn't require an extra block in the output buffer. And then don't allow mixing this method and `update.
* Same as option 2, but introduce an entirely separate type to enforce the requirements at compile time.

* **Does this change any public APIs?** Yes, but not a breaking change.
* **Which algorithm(s) will this impact?** All ciphers when used with `Streaming*Key`.

### Requirements / Acceptance Criteria:

What must a solution address in order to solve the problem? How do we know the solution is complete?

* **RFC links:** Links to relevant RFC(s)
* **Related Issues:** Link any relevant issues
* **Will the Usage Guide or other documentation need to be updated?**
* **Testing:** How will this change be tested? Call out new integration tests, functional tests, or particularly
  interesting/important unit tests.

### Out of scope:

Is there anything the solution will intentionally NOT address?

justsmth:
> Even when the caller ensures that the input buffer is always a multiple of the block length ...

This is mostly due to the [underlying function](https://github.com/aws/aws-lc/blob/56701c6294879968bcfb9bc3d429e7a3ac164891/include/openssl/cipher.h#L186-L197) we [use for the operation](https://github.com/aws/aws-lc-rs/blob/1cec9de4bfbf426e25ca223b20739ada5f72cd11/aws-lc-rs/src/cipher/streaming.rs#L185):
```
// EVP_EncryptUpdate encrypts |in_len| bytes from |in| to |out|. The number
// of output bytes may be up to |in_len| plus the block length minus one and
// |out| must have sufficient space. The number of bytes actually output is
[...]

PR #991, opened 2026-01-07
`less_safe_update` for Streaming ciphers

justsmth:
### Issues:
Addresses #990

### Description of changes:

This PR adds `less_safe_update` methods to `StreamingEncryptingKey` and `StreamingDecryptingKey` that have looser output buffer size requirements compared to the existing `update` methods.

**New `less_safe_update` API:**
- Calculates minimum output buffer size based on the total bytes of output already generated and the cipher's block length
- The minimum size is: `input.len() + ((block_len - (next_total % block_len)) % block_len)`
- This allows callers to use tighter buffer allocations when processing data in chunks, particularly useful when input is provided one block at a time (where no additional buffer space is needed)

**Improvements to existing `update` methods:**
- Loosened the output buffer requirement for `StreamingDecryptingKey::update` from `input.len() + block_len` to `input.len() + block_len - 1`, consistent with `EVP_DecryptUpdate` documentation and matching the existing `StreamingEncryptingKey::update` behavior
- Fixed documentation for `StreamingEncryptingKey::update` which incorrectly referred to "ciphertext" input instead of "plaintext"

**Code refactoring:**
- Extracted common cipher update logic into private `update_internal` helper methods to reduce code duplication between `update` and `less_safe_update`
- Refactored test helper functions (`step_encrypt`, `step_decrypt`) to use a closure-based pattern, eliminating duplication between safe and less_safe test variants

### Call-outs:
- The `less_safe_update` methods include a runtime assertion that panics if the cipher writes more bytes than the output buffer length (memory corruption safeguard)
- The buffer size calculation in `less_safe_update` is based on observed behavior of the underlying `EVP_EncryptUpdate`/`EVP_DecryptUpdate` functions

### Testing:
- Existing streaming cipher tests (`test_step_cbc`, `test_step_ctr`, `test_step_cfb128`, `test_step_ecb_pkcs7`) continue to pass
- Tests verify that the tighter buffer calculations work correctly across various input sizes and step sizes
- Tests include assertions that when input is provided one block at a time, no additional buffer space is required

### DIFF
--- a/aws-lc-rs/src/cipher/streaming.rs
+++ b/aws-lc-rs/src/cipher/streaming.rs
@@ -22,7 +22,6 @@ pub struct StreamingEncryptingKey {
     mode: OperatingMode,
     cipher_ctx: LcPtr<EVP_CIPHER_CTX>,
     context: EncryptionContext,
-    #[cfg(feature = "unstable")]
     output_generated: usize,
 }
 
@@ -154,7 +153,6 @@ impl StreamingEncryptingKey {
             mode,
             cipher_ctx,
             context,
-            #[cfg(feature = "unstable")]
             output_generated: 0,
         })
     }
@@ -184,11 +182,8 @@ impl StreamingEncryptingKey {
         }
         let outlen: usize = outlen.try_into()?;
         debug_assert!(outlen <= min_outsize);
-        #[cfg(feature = "unstable")]
-        {
-            self.output_generated += outlen;
-            assert!(outlen <= output.len());
-        }
+        self.output_generated += outlen;
+        assert!(outlen <= output.len());
 
         Ok(BufferUpdate::new(output, outlen))
     }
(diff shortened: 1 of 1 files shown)


aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -194,9 +184,83 @@ impl StreamingEncryptingKey {
...
+    pub fn update<'a>(
+        &mut self,
+        input: &[u8],
+        output: &'a mut [u8],
+    ) -> Result<BufferUpdate<'a>, Unspecified> {
+        let min_outsize = input
+            .len()
+            .checked_add(self.algorithm().block_len())
+            .ok_or(Unspecified)?
+            - 1;
```
justsmth: We could use a `checked_sub(1)` here.  We only use a block length of 16, so the compiler should be able to optimize away the check.

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -874,6 +1119,227 @@ mod tests {
...
+                1,
+            );
+        }
+    }
+
+    #[test]
+    fn test_step_ctr_less_safe() {
+        let random = SystemRandom::new();
+        let mut key = [0u8; AES_256_KEY_LEN];
+        random.fill(&mut key).unwrap();
```
WillChilds-Klein, line 1182: np: here and elsewhere, should we seed the DRBG and print the seed to allow reproduction?

justsmth: Unfortunately for testing, AWS-LC doesn't have a mechanism for explicitly seeding its DRBG.

skmcgrail (approved):

PR #995, opened 2026-01-08
Unify sys-crate build scripts

justsmth:
## Description

This PR unifies the build scripts for `aws-lc-sys` and `aws-lc-fips-sys` into a single shared `./builder` directory. Both sys-crate `builder` directories are now symlinks to `../builder`, eliminating significant code duplication between the two crates.

## Call-outs

- **_fips_ feature for `aws-lc-sys`**: This will not used by `aws-lc-rs`, but `aws-lc-sys` now supports a "fips" feature that triggers a FIPS-module build of its source.
- **Simplified `bindgen` dependency**: `bindgen` is now a standard build dependency for `aws-lc-fips-sys`. (Like with `aws-lc-sys`, it's no longer conditionally included based on target platform.)
- **Removed `rust_wrapper` from FIPS**: The FIPS build no longer produces a separate `rust_wrapper` library; helper functions are now implemented in pure Rust
- **Added `EnvGuard`**: RAII helper for temporary environment variable modifications during builds.

### DIFF
--- a/builder/cmake_builder.rs
+++ b/builder/cmake_builder.rs
@@ -158,7 +158,6 @@ impl CmakeBuilder {
 
             if target_env() == "ohos" {
                 Self::configure_open_harmony(&mut cmake_cfg);
-                return cmake_cfg;
             }
         }
 
@@ -297,16 +296,23 @@ impl CmakeBuilder {
             );
             bat_script
         } else {
-            // Fallback to current logic if neither can execute
+            // Neither script could be verified as executable. This can happen in sandboxed or
+            // restricted build environments where trial execution is blocked. We fall back to
+            // selecting a script based on the host OS, which matches the behavior prior to
+            // execution-based detection. If this fallback is incorrect, the build will fail
+            // when CMake attempts to invoke the selected script.
             let fallback_script = if cfg!(target_os = "windows") {
                 bat_script
             } else {
                 sh_script
             };
-            emit_warning(
-                format!(
-                    "Neither script could be tested for execution, falling back to target-based selection: {}",
-                    fallback_script.file_name().unwrap().to_str().unwrap()));
+            emit_warning(format!(
+                "WARNING: Neither prebuilt-nasm.sh nor prebuilt-nasm.bat could be verified as \
+                     executable. Falling back to target-based selection: {}. If the build fails \
+                     during assembly, verify that the selected script is appropriate for your \
+                     build environment.",
+                fallback_script.file_name().unwrap().to_str().unwrap()
+            ));
             fallback_script
         }
     }
--- a/builder/main.rs
+++ b/builder/main.rs
@@ -945,10 +945,11 @@ fn setup_include_paths(out_dir: &Path, manifest_dir: &Path) -> PathBuf {
     for path in include_paths {
         for child in std::fs::read_dir(path).into_iter().flatten().flatten() {
             if child.file_type().map_or(false, |t| t.is_file()) {
-                let _ = std::fs::copy(
+                std::fs::copy(
                     child.path(),
                     include_dir.join(child.path().file_name().unwrap()),
-                );
+                )
+                .expect("Failed to copy include file during build setup");
                 continue;
             }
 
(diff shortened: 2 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -405,6 +405,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       AWS_LC_SYS_NO_ASM: 1
+      AWS_LC_RS_DISABLE_SLOW_TESTS: 1
```
WillChilds-Klein: don't think this mentioned in the description -- why do we need to disable slow tests for these platform combos now?

aws-lc-fips-sys/src/lib.rs
```diff
@@ -82,25 +57,31 @@ platform_binding!(
...
-    unsafe { ERR_GET_LIB_RUST(packed_error) }
+    ((packed_error >> 24) & 0xFF) as i32
 }
 
-#[allow(non_snake_case)]
+#[allow(non_snake_case, clippy::cast_possible_wrap)]
 #[must_use]
 pub fn ERR_GET_REASON(packed_error: u32) -> i32 {
-    unsafe { ERR_GET_REASON_RUST(packed_error) }
+    (packed_error & 0xFFF) as i32
```
WillChilds-Klein: lowest 12 bits? should this be `0xFFFF`? stylistically, suggest `0x0FFF` if not

justsmth: Yeah, the bottom 12 bits. It's essentially copied [from here](https://github.com/aws/aws-lc/blob/2a73d20f124106ebe09a910a74f1393c6110958a/include/openssl/err.h#L179).
```
// ERR_GET_REASON returns the reason code for the error. This is one of
// library-specific |LIB_R_*| values where |LIB| is the library (see
// |ERR_GET_LIB|). Note that reason codes are specific to the library.
OPENSSL_INLINE int ERR_GET_REASON(uint32_t packed_error) {
  return (int)(packed_error & 0xfff);
}
```

.github/workflows/tests.yml
justsmth: Not needed.  These are the tests that I'm always waiting ~30 minutes for them to complete, and thought I'd add the environment variable to avoid running the long RSA 8192 key generation tests.

WillChilds-Klein (approved):

builder/cmake_builder.rs
```diff
@@ -118,17 +140,26 @@ impl CmakeBuilder {
...
+                cmake_cfg.define("DISABLE_GO", "ON");
+            }
+            if Some(true) == disable_jitter_entropy() {
+                cmake_cfg.define("DISABLE_CPU_JITTER_ENTROPY", "ON");
+            }
+
+            if target_env() == "ohos" {
+                Self::configure_open_harmony(&mut cmake_cfg);
+                return cmake_cfg;
+            }
```
skmcgrail, line 161: Why do we shortcut out early for open harmony but don't do so for all the other target environments?

builder/main.rs
skmcgrail: Should these copies be an unwrap in the event they fail? Otherwise could result in some unclear error messages?

builder/cmake_builder.rs
skmcgrail: In what scenario would this still work? i.e. Neither script was executable by our testing, but then we fallback and just pick one of them based off the target_os? Wouldn't it still fail execution later?

builder/cmake_builder.rs
justsmth: This is a good question -- I think it can be removed. I tried to avoid making many behavioral changes in this PR to avoid potentially breaking consumers due to a refactoring, but this change seems safe enough as we have CI coverage for this platform.

builder/cmake_builder.rs
justsmth: Good point — I'm not sure if this fallback is ever hit in practice. To be safe, I'd rather keep the existing behavior, but I added comments and a better warning message so it's clear what's happening and what to check if the build fails.

skmcgrail (approved):

Issue #996, opened 2026-01-08
Disabling pre-built nasm binaries through feature flags (`aws-lc-rs/disable-prebuilt-nasm`)


kpcyrd:
### Problem:

I have some users in the Debian project (not Debian packaging, but Debian infrastructure), that _really_ dislike pre-built binaries.

I asked them to set `AWS_LC_SYS_PREBUILT_NASM=0` for now, but I'd rather have some cargo feature I can set in `Cargo.toml` to ensure the build should rather fail instead of using prebuilt binaries.

### Solution:

Making sure the `aws-lc-rs/prebuilt-nasm` feature flag never gets activated is somewhat tricky to ensure reliably, and since features are additive I can't turn it off again, so I'd need some feature like `aws-lc-rs/disable-prebuilt-nasm` that explicitly overrides `aws-lc-rs/prebuilt-nasm`.

### Out of scope:

If you're interested in integrating the pre-built binaries into reproducible-builds.org for independent verification from source, that's something I can also help with (feel free to email kpcyrd at archlinux dot org).

justsmth:
Hello! I posted a draft PR related to this. 

A couple of points to clarify:
* We only use NASM for Windows targets. All other targets use the GCC/Clang assembler.
* On Windows, the prebuilt objects will NOT be used if NASM is installed on the system. (So we only use prebuilt objects in a situation where the build would otherwise have failed.)

The PR is fairly simple. I'll see if we can get it merged for the next release.

kpcyrd:
hm, it might be something else then. It was noticed that on `ppc64el` the build failed because cmake was missing, so there were some concerns on the "more mainstream" platforms it would use a pre-compiled binary.

Thanks for clearing that up!

justsmth:
Hello -- we just release [aws-lc-rs v1.15.3](https://github.com/aws/aws-lc-rs/releases/tag/v1.15.3) 🎉, which provides a `disable-prebuilt-nasm ` feature as requested above. We also no longer require CMake for non-FIPS builds.

This issue will now be closed.  Feel free to comment here or open another issue if you continue to have problems after upgrading. Thanks for reporting and helping to improve our library! 😃

PR #997, opened 2026-01-08
disable-prebuilt-nasm feature

justsmth:
### Issues:
Addresses #996

### Description of changes: 
Provide a `disable-prebuilt-nasm` feature that prevents use of prebuilt-NASM regardless of any other factor.

### Call-out
* All this feature really does is force a build failure in the situation where the "nasm" executable is not found.
   * The logic has always been to NOT use prebuilt-nasm objects if we found the "nasm" executable. In other words -- if we find can find the NASM assembler, then we use the NASM assembler.

### Testing:
Added test.


kpcyrd:
This is exactly what I had in mind, thank you!

WillChilds-Klein (approved):

PR #1000, opened 2026-01-13
Support Windows 7

justsmth:
### Issues:
Addresses aws/aws-lc#1997

### Description of changes: 
When targeting Windows 7, ensure that there's a definition for `_WIN32_WINNT)`.
* Related: aws/aws-lc#2940

### DIFF
--- a/aws-lc-sys/builder/cc_builder.rs
+++ b/aws-lc-sys/builder/cc_builder.rs
@@ -406,9 +406,10 @@ impl CcBuilder {
             if target().ends_with("-win7-windows-msvc") {
                 // 0x0601 is the value of `_WIN32_WINNT_WIN7`
                 build_options.push(BuildOption::define("_WIN32_WINNT", "0x0601"));
-                emit_warning(
-                    "Setting _WIN32_WINNT to _WIN32_WINNT_WIN7 for x86_64-win7-windows-msvc target",
-                );
+                emit_warning(format!(
+                    "Setting _WIN32_WINNT to _WIN32_WINNT_WIN7 for {} target",
+                    target()
+                ));
             }
         }
     }


aws-lc-sys/builder/cc_builder.rs
```diff
@@ -402,6 +402,14 @@ impl CcBuilder {
                 "_STL_EXTRA_DISABLED_WARNINGS",
                 "4774 4987",
             ));
+
+            if target().ends_with("-win7-windows-msvc") {
```
0blu, line 406: Maybe just `-win7-windows-`.  
I also wonder, if string comparisons are the correct way to check for target tiples.

aws-lc-sys/builder/cc_builder.rs
```diff
@@ -402,6 +402,14 @@ impl CcBuilder {
...
+                // 0x0601 is the value of `_WIN32_WINNT_WIN7`
+                build_options.push(BuildOption::define("_WIN32_WINNT", "0x0601"));
+                emit_warning(
+                    "Setting _WIN32_WINNT to _WIN32_WINNT_WIN7 for x86_64-win7-windows-msvc target",
```
0blu: The text "[...] for **x86_64-win7-windows-msvc** target" might not be true for other targets, see `if` statement above.

aws-lc-sys/builder/cc_builder.rs
0blu: Regard my first comment. I just saw that this is inside `if is_like_msvc` so `gnu` is not even a realistic value at this path.

justsmth: > I also wonder, if string comparisons are the correct way to check for target tiples.

Yeah, I'm not sure whether there is any way other than searching `TARGET` to determine that it's Windows 7.

aws-lc-sys/builder/cc_builder.rs
```diff
@@ -402,6 +402,15 @@ impl CcBuilder {
                 "_STL_EXTRA_DISABLED_WARNINGS",
                 "4774 4987",
             ));
+
+            if target().ends_with("-win7-windows-msvc") {
+                // 0x0601 is the value of `_WIN32_WINNT_WIN7`
+                build_options.push(BuildOption::define("_WIN32_WINNT", "0x0601"));
```
WillChilds-Klein, line 408: confirmed: https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170

```
#define _WIN32_WINNT_WIN7                   0x0601 // Windows 7
```

WillChilds-Klein (approved):

skmcgrail (approved):

PR #1011, opened 2026-01-20
HKDF: Fewer allocations

justsmth:
### Issues:
Addresses: P369041323

### Description of changes: 
* Reduce number of required memory allocations in HKDF.
* Salt and Info values are public; no need to "zeroize" them.

### Call-out
On micro-benchmarks, I see a 2-10% latency reduction on both x86-64 and arm64.


aws-lc-rs/src/hkdf.rs
```diff
@@ -127,8 +120,10 @@ impl Salt {
     /// `new` panics if salt creation fails
     #[must_use]
     pub fn new(algorithm: Algorithm, value: &[u8]) -> Self {
-        let bytes = value.to_vec().into_boxed_slice();
-        Self { algorithm, bytes }
+        Self {
+            algorithm,
+            bytes: Arc::new(Box::from(value.to_owned())),
```
Mark-Simulacrum: This seems like it wants to be `Arc::from(value)` directly ([impl](https://doc.rust-lang.org/nightly/std/sync/struct.Arc.html#impl-From%3C%26%5BT%5D%3E-for-Arc%3C%5BT%5D%3E)) -- right now this allocates twice (Arc + Box) for not much reason I think?

justsmth: Thanks! 🤦‍♂️ I don't know why I didn't see that.

torben-hansen (approved):

aws-lc-rs/tests/hkdf_test.rs
```diff
@@ -95,7 +95,7 @@ fn hkdf_output_len_tests() {
 #[test]
 fn hkdf_info_len_tests() {
     for &alg in &[hkdf::HKDF_SHA256, hkdf::HKDF_SHA384, hkdf::HKDF_SHA512] {
-        for info_length in (50..300).step_by(7) {
+        for info_length in (40..400).step_by(13) {
```
skmcgrail, line 98: np: A comment to the intention of this range and step value would be good. Was unclear before with thinking about it, and changing it doesn't make it clearer either.

skmcgrail (approved):

PR #1030, opened 2026-02-02
Support cranelift backend on Windows

### LINKED ISSUE #537: Unable to build w/ rustc_codegen_cranelift compiler backend
### Problem:

From https://github.com/rustls/rustls/issues/2133:

On Windows 10 with Microsoft Build Tools latest version as of today I get this error:

```
error: failed to run custom build command for `aws-lc-sys v0.21.2`

Caused by:
  process didn't exit successfully: `C:\prj\target\debug\build\aws-lc-sys-19d1ffd36a8f3b39\build-script-main` (exit code: 0xc000001d, STATUS_ILLEGAL_INSTRUCTION)
  --- stdout
  cargo:rerun-if-env-changed=AWS_LC_SYS_NO_PREFIX
  cargo:rerun-if-env-changed=AWS_LC_SYS_INTERNAL_BINDGEN
  cargo:rerun-if-env-changed=AWS_LC_SYS_EXTERNAL_BINDGEN
  cargo:rerun-if-env-changed=AWS_LC_SYS_NO_ASM
  cargo:rerun-if-env-changed=AWS_LC_SYS_CFLAGS
  cargo:rerun-if-env-changed=AWS_LC_SYS_PREBUILT_NASM
  cargo:rerun-if-env-changed=AWS_LC_SYS_C_STD
  cargo:rustc-cfg=x86_64_pc_windows_msvc
  cargo:rerun-if-env-changed=AWS_LC_SYS_CMAKE_BUILDER
  cargo:rerun-if-env-changed=AWS_LC_SYS_STATIC
  default_for Target: 'x86_64-pc-windows-msvc'
  cargo:rerun-if-env-changed=AWS_LC_SYS_STATIC
  cargo:rerun-if-env-changed=CMAKE

  --- stderr
  Missing dependency: cmake
  thread 'main' panicked at C:\Users\fred\.cargo\registry\src\index.crates.io-6f17d22bba15001f\aws-lc-sys-0.21.2\builder/main.rs:315:40:
  called `Result::unwrap()` on an `Err` value: "Required build dependency is missing. Halting build."
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
```

`aws-lc-sys` is required by `aws-lc-rs@1.9.0` which is required by `rustls@0.23.13` which is required by `axum-server@0.7.1`.

Then I followed what is in here: https://aws.github.io/aws-lc-rs/requirements/windows.html.

I installed C/C++ Compiler and CMake but not NASM and it builds but at the end I get a lot of errors like these:

```
librustls-01ac9141ee708bf1.rlib(rustls-01ac9141ee708bf1.rustls.78218e420b8f09f3-cgu.13.rcgu.o) : error LNK2001: unresolved external symbol ␁aws_lc_0_21_2_EVP_PKEY_derive_init␍
librustls-01ac9141ee708bf1.rlib(rustls-01ac9141ee708bf1.rustls.78218e420b8f09f3-cgu.13.rcgu.o) : error LNK2001: unresolved external symbol ␁aws_lc_0_21_2_EVP_PKEY_derive_set_peer␍
librustls-01ac9141ee708bf1.rlib(rustls-01ac9141ee708bf1.rustls.78218e420b8f09f3-cgu.13.rcgu.o) : error LNK2001: unresolved external symbol ␁aws_lc_0_21_2_EVP_PKEY_derive␍
librustls-01ac9141ee708bf1.rlib(rustls-01ac9141ee708bf1.rustls.78218e420b8f09f3-cgu.02.rcgu.o) : error LNK2001: unresolved external symbol ␁aws_lc_0_21_2_CRYPTO_memcmp␍
... and a lot more
```

Everything is fine if I downgrade `axum-server` to 0.6.0.

#### Relevant details
 * CPU architecture: x86-64
 * CPU name: Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz
 * OS: Windows 10 x64

justsmth:
### Issues:
Resolves #537

### Description of changes: 
This PR extends support for the [rustc_codegen_cranelift](https://github.com/rust-lang/rustc_codegen_cranelift) compiler backend to Windows.

The previous release (v1.15.3) added cranelift support for Linux and macOS, but Windows was missing from the CI test matrix. When building with cranelift on Windows (`x86_64-pc-windows-msvc`), linking fails with "unresolved external symbol" errors because Windows targets use symbol prefixing (with a leading underscore `_`), but the existing `universal_no_u1` bindings don't account for this.

**Root Cause:**
The bindings generated by rust-bindgen include a `\u{1}` prefix in `#[link_name]` attributes to prevent symbol name mangling. This is an LLVM-specific feature that cranelift doesn't support. On Windows, symbols are additionally prefixed with `_aws_lc_...` instead of just `aws_lc_...`, so we need a separate set of bindings that:
1. Strips the `\u{1}` prefix (as we already do for non-prefixed bindings)
2. Also strips the leading underscore from the prefixed symbol names

**Changes:**
- Add new `universal_no_u1_prefixed` bindings file generated via sed from `universal_prefixed_crypto.rs`
- Update build script to emit `universal-no-u1-prefixed` cfg when cranelift is detected on targets that use prefixed symbols (e.g., Windows MSVC)
- Add Windows (`x86_64-pc-windows-msvc`) to the cranelift CI test matrix

### Call-outs:
- The new `universal_no_u1_prefixed_crypto.rs` file is large (~14k lines) because it's a copy of the bindings with sed transformations applied
- This is the same approach used for the existing `universal_no_u1_crypto.rs` file

### Testing:
CI tests will now run with the cranelift backend on Windows in addition to Linux and macOS.


aws-lc-sys/builder/main.rs
```diff
@@ -570,7 +570,11 @@ fn initialize() {
                         "Cranelift codegen backend detected. Using universal_no_u1 bindings.",
                     );
                 }
-                emit_rustc_cfg("universal-no-u1");
+                if target_has_prefixed_symbols() {
+                    emit_rustc_cfg("universal-no-u1-prefixed");
+                } else {
+                    emit_rustc_cfg("universal-no-u1");
+                }
```
skmcgrail, line 577: I feel like `soh` would have been a better choice, as the character name is "start of heading", with "SOH" being the common abbreviation and is what appears if you have a font that shows the glyph.

WillChilds-Klein (approved):

aws-lc-sys/builder/main.rs
justsmth: I agree. I'm refactoring this logic soon to unite the build scripts for the two crates. I'll plan to rename this file as part of that.

Issue #964, opened 2025-11-26
ML-DSA stabilization?


djc:
Given https://aws.amazon.com/about-aws/whats-new/2025/11/aws-private-ca-post-quantum-digital-certificates/ and the publication of [RFC 9881](https://datatracker.ietf.org/doc/rfc9881/), is there a plan for stabilization of ML-DSA?

Original issue:

- #773

justsmth:
Sorry, I believe the issue is that the "fips" branch of AWS-LC that we currently track for aws-lc-fips-sys doesn't have the necessary support for ML-DSA. We will only be able to stabilize this API when we update the branch that we track. 

I'll double check to verify my assumptions next week. I would anticipate being able to bump aws-lc-fips-sys to the latest branch sometime around the new year, but there's a few factors that might push it back a bit further.

DarkmatterVale:
Interested in this as well, what's the status on stabilizing the ML-DSA signature? Will it be released soon?

justsmth:
> Will it be released soon?

Sorry. We are still waiting as well. The most recent update I've heard anticipates a March switchover date.

DarkmatterVale:
Hi all - any update on this issue?

justsmth:
Sorry, this is taking much longer than any of us expected. 

On the [NIST CMVP modules in process page](https://csrc.nist.gov/Projects/cryptographic-module-validation-program/modules-in-process/modules-in-process-list) we still have "AWS-LC 3 Cryptographic Module" (dynamic and static) listed. We have been waiting for the certification process to complete before we cut `aws-lc-fips-sys` over to using our "FIPS 4.0" module (which is also now listed as in-process).

The "FIPS 4.0" module supports ML-DSA, which will allow us to stabilize the associated APIs in `aws-lc-rs`. Hopefully, this will be very soon.

1456055067:
@justsmth Looks like AWS-LC 3 moved to active status and LC 4 is now in process. Can you verify?

justsmth:
PR for the switch to the next AWS-LC FIPS branch:
* #1185

justsmth:
Hello! [aws-lc-rs v1.18.0](https://github.com/aws/aws-lc-rs/releases/tag/v1.18.0) is now released 🎉 -- the ML-DSA APIs are now stable with the switch to AWS-LC FIPS 4.0 (#1185, #1188).

[...]

Issue #773, opened 2025-04-10
Add support for ML-DSA


djc:
### Problem:

We've seen a few issues filed against rustls this week asking for support of post-quantum signature algorithms, possibly triggered by the OpenSSL release that included ML-KEM, ML-DSA and SLH-DSA support. It would be nice if aws-lc-rs supported ML-DSA (which seems to be the more popular of the post-quantum signature options at least in the TLS space, right now).

### Requirements / Acceptance Criteria:

https://datatracker.ietf.org/doc/draft-tls-westerbaan-mldsa/ might soon be adopted by the TLS WG. Should follow https://doi.org/10.6028/nist.fips.204. It appears aws-lc has [reasonably complete support](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/PQREADME.md#fips-204-module-lattice-based-digital-signature-algorithm-ml-dsa) for ML-DSA already?

justsmth:
Hello!

Yes, ML-DSA 44/65/87 should be support (although still unstable) as-of [v1.13.0](https://github.com/aws/aws-lc-rs/releases/tag/v1.13.0): https://docs.rs/aws-lc-rs/latest/aws_lc_rs/unstable/signature/index.html

Let us know if you have any concerns/suggestions regarding this new API.

justsmth:
I asked around about that last issue. Unless there's customer demand for SLH-DSA support, we do not have it listed on our (short-term) timeline. 

Feel free let us know if our lack of support negatively affects you or other consumers. Thanks!

djc:
> Feel free let us know if our lack of support negatively affects you or other consumers. Thanks!

So far the issues I've seen seem to be asking for any post-quantum signature algorithms. Will make sure issues get filed once folks start specifically asking about algorithms that are not ML-DSA.

I'll also try to make a proof of concept for ML-DSA support in rustls.

PR #1034, opened 2026-02-08
feat(pqdsa): Add deterministic seed-based key generation for ML-DSA

LatticeArc-Founder:
## Description

Add deterministic seed-based key generation for ML-DSA, enabling `PqdsaKeyPair` construction from a 32-byte seed per FIPS 204.

Related to #964 (ML-DSA stabilization)
Related to #773 (ML-DSA support)

## Changes

- Add `PqdsaKeyPair::from_seed(algorithm, seed)` constructor with size validation
- Add `AlgorithmID::seed_size_bytes()` constant (32 bytes for all ML-DSA variants)
- Enhance ACVP keygen test macro to validate seed-to-private-key mapping against NIST vectors

## Implementation Notes

This exposes the existing aws-lc C function:
- `EVP_PKEY_pqdsa_new_raw_private_key` with seed-length input, which internally calls `PQDSA_KEY_set_raw_keypair_from_seed()` to generate the full keypair

The method calls `EVP_PKEY_pqdsa_new_raw_private_key` directly with the algorithm NID, rather than going through `parse_raw_private_key`, because the latter routes by expanded private key length (2560/4032/4896) and would not match a 32-byte seed.

The seed is not retained in the returned `PqdsaKeyPair`; the expanded key material is stored instead. The expanded private key can be retrieved via `private_key()` and serialized via `to_pkcs8()` or `as_raw_bytes()`.

## Testing

- 12 unit tests in `src/pqdsa/key_pair.rs` (8 new for `from_seed`)
- 13 integration tests in `tests/pqdsa_test.rs` (6 new + 3 enhanced ACVP keygen tests)
- Tests cover: determinism, sign/verify, wrong size rejection, zeroed bytes, different seeds, serialization round-trip (PKCS#8 + raw), byte stability, functional equivalence, cross-algorithm behavior
- ACVP keygen tests now verify seed-based generation produces byte-for-byte identical private keys to NIST test vectors

## Checklist

- [x] Working against latest main branch
- [x] Local tests pass
- [x] Focused on specific change only
- [x] Clear commit message

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,20 +18,24 @@ resolver = "2"
 [workspace.dependencies]
 bindgen = "0.72.0"
 cc = "1.2.26"
-clap = "=4.4.18"# v4.5 is MSRV 1.74
 cmake = "0.1.54"
-criterion = "0.5.1" # v0.6.0 depends on clap v4.5
 dunce = "1.0.5"
 fs_extra = "1.3.0"
+regex = "1.11.1"
+untrusted = "0.7.1"
+zeroize = "1.8.1"
+
+# The dependencies below do not affect our MSRV. They are only used as
+# a dev-dependency or only used outside of our public crates.
+# These can be updated to the latest versions.
+clap = "4.4"
+criterion = "0.8.2"
 hex = "0.4.3"
 lazy_static = "1.5.0"
 openssl = "0.10.73"
 paste = "1.0.15"
-regex = "1.11.1"
 ring = "0.17.14"
-untrusted = "0.7.1"
-zeroize = "1.8.1"
-toml_edit = "0.24.0"
+toml_edit = "0.25.0"
 
 [profile.bench]
 lto = true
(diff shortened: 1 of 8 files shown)


aws-lc-rs/src/pqdsa/key_pair.rs
```diff
@@ -116,6 +118,61 @@ impl PqdsaKeyPair {
...
+            core::cmp::Ordering::Equal => {}
+        }
+        let nid = algorithm.0.id.nid();
+        let evp_pkey = LcPtr::new(unsafe {
+            EVP_PKEY_pqdsa_new_raw_private_key(nid, seed.as_ptr(), seed.len())
+        })
+        .map_err(|()| KeyRejected::unspecified())?;
+        validate_pqdsa_evp_key(&evp_pkey, algorithm.0.id)?;
+        let pubkey =
+            PublicKey::from_private_evp_pkey(&evp_pkey).map_err(|_| KeyRejected::unspecified())?;
```
dkostic, line 168: I guess we could have used `PublicKey::from_private_evp_pkey(&evp_pkey)?;` instead?

justsmth: Yeah, this explicit conversion is not needed as we have one defined [here](https://github.com/aws/aws-lc-rs/blob/d91c3af06fdf5d742a0ddd3f125ef548fe4dc555/aws-lc-rs/src/error.rs#L226-L230).

PR #1040, opened 2026-02-12
Speedup CI completion

justsmth:
### Description of changes: 
Set `AWS_LC_RS_DISABLE_SLOW_TESTS` for several slow running (>=30 minute) CI jobs that we have. (Setting this environment variable allows skipping slower tests, such as RSA 8192 key generation.)

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -239,6 +239,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CROSS_CONFIG: "./Cross.toml.x86_64-unknown-linux-gnu"
+      AWS_LC_RS_DISABLE_SLOW_TESTS: 1
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/cross.yml
```diff
@@ -239,6 +239,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CROSS_CONFIG: "./Cross.toml.x86_64-unknown-linux-gnu"
+      AWS_LC_RS_DISABLE_SLOW_TESTS: 1
```
WillChilds-Klein, line 242: so when or how would those tests ever get run? is the issue that they're only slow on _some_ platforms, so we disable them there but run them on others?

justsmth: They run for every job that doesn't specify `AWS_LC_RS_DISABLE_SLOW_TESTS=1`, which is essentially all jobs that don't run under emulation.

PR #1047, opened 2026-02-18
Check cranelift when using the cargo-clif executable

frederikhors:
Fixes https://github.com/aws/aws-lc-rs/issues/537#issuecomment-3921457859.

### DIFF
--- a/aws-lc-sys/builder/main.rs
+++ b/aws-lc-sys/builder/main.rs
@@ -271,7 +271,7 @@ fn is_cranelift_backend() -> bool {
     // CARGO_ENCODED_RUSTFLAGS contains flags separated by 0x1f (ASCII Unit Separator)
     if let Some(rustflags) = optional_env("CARGO_ENCODED_RUSTFLAGS") {
         for flag in rustflags.split('\x1f') {
-            if flag.contains("cranelift") {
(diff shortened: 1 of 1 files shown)


bjorn3:
This may be a bit too loose. It would also match for example `RUSTFLAGS="--cfg use_cranelift_for_wasm"` or something like that. Not that I could find any actual cases on github where that happens though.

justsmth:
From my testing, I believe most platforms were fine linking using the "no u1" (terrible name) bindings, even if it's not required.  So it might be ok for this to be overly broad in catching "cranelift".

Would this also catch for consumers using `codegen-backend = "cranelift"` in their `.cargo/config.toml`?

frederikhors:
> Would this also catch for consumers using `codegen-backend = "cranelift"` in their `.cargo/config.toml`?

I think this already works in the actual main branch code.

justsmth (approved):

frederikhors:
I can't wait to try it!

justsmth:
Hi @frederikhors -- We have a CI job that verifies the following text appears in the PR description:

If you agree, would you update the PR description with this text?

frederikhors:
Do you mean in the first post? Done.

samuel40791765 (approved):

PR #1058, opened 2026-03-09
Implement HMAC `sign_to_buffer`

fabrice102:
### Description of changes: 

Implement HMAC `sign_to_buffer`.

This follows the implementation of CMAC `sign_to_buffer` (https://github.com/aws/aws-lc-rs/pull/968), refactoring to make the HMAC code match the CMAC code.

### Call-outs:

Slightly changed the internal implementation to match CMAC.
In particular, there is a new `debug_assert` to increase robustness, and a new function `tag_len` to make it easier to get the tag length (and to match CMAC).

### Testing:

Added the same tests as for CMAC, and ran `cargo test --workspace`.


aws-lc-rs/src/hmac.rs
```diff
@@ -459,6 +491,43 @@ pub fn sign(key: &Key, data: &[u8]) -> Tag {
...
 }
 
+/// Calculates the HMAC of `data` using the key `key` in one step,
+/// writing the result into the provided `output` buffer.
+///
+/// The `output` buffer must be at least as large as the algorithm's
+/// tag length (i.e., `key.algorithm().tag_len()`). The returned slice will be a
+/// sub-slice of `output` containing exactly the tag bytes.
+///
+/// It is generally not safe to implement CMAC verification by comparing the
```
justsmth: > It is generally not safe to implement CMAC verification ...

HMAC

aws-lc-rs/src/hmac.rs
```diff
@@ -420,20 +427,45 @@ impl Context {
...
+
+    unsafe {
+        if 1 != indicator_check!(HMAC_Final(
+            ctx.key.get_hmac_ctx_ptr(),
+            output.as_mut_ptr(),
+            out_len.as_mut_ptr(),
+        )) {
+            return Err(Unspecified);
+        }
+        let actual_len = out_len.assume_init() as usize;
```
justsmth: I know much of the existing code isn't doing this properly, but would you narrow the scope of the `unsafe` blocks to wrap the specific function call where it's needed:
```rust
...
    if 1 != indicator_check!(unsafe {
        HMAC_Final(
            ctx.key.get_hmac_ctx_ptr(),
            output.as_mut_ptr(),
            out_len.as_mut_ptr(),
        )
    }) {
        return Err(Unspecified);
    }
    let actual_len = unsafe { out_len.assume_init() } as usize;
...
```

aws-lc-rs/src/hmac.rs
```diff
@@ -420,20 +427,45 @@ impl Context {
...
+        })
+    }
+}
+
+#[inline]
+pub(crate) fn internal_sign<'in_out>(
+    ctx: &mut Context,
+    output: &'in_out mut [u8],
+) -> Result<&'in_out mut [u8], Unspecified> {
+    let mut out_len = MaybeUninit::<c_uint>::uninit();
```
justsmth, line 451: Since its expected for the caller to verify the output length as a pre-condition, add a `debug_assert!` at the start:
```rust
pub(crate) fn internal_sign<'in_out>(
    ctx: &mut Context,
    output: &'in_out mut [u8],
) -> Result<&'in_out mut [u8], Unspecified> {
    debug_assert!(
        output.len() >= ctx.key.algorithm().tag_len(),
        "output buffer too small for HMAC tag"
    );
...
```

aws-lc-rs/src/hmac.rs
```diff
@@ -420,20 +427,45 @@ impl Context {
...
+        debug_assert!(
+            actual_len <= digest::MAX_OUTPUT_LEN,
+            "HMAC tag length {actual_len} exceeds maximum {}",
+            digest::MAX_OUTPUT_LEN
+        );
+        if actual_len != ctx.key.algorithm().tag_len() {
+            return Err(Unspecified);
         }
+
+        Ok(&mut output[..actual_len])
```
justsmth: NP: `Ok(&mut output[0..actual_len])`

aws-lc-rs/src/hmac.rs
```diff
@@ -538,4 +616,50 @@ mod tests {
...
+            let tag_len = algorithm.tag_len();
+
+            // Test with exact size buffer
+            let mut output = vec![0u8; tag_len];
+            let result = hmac::sign_to_buffer(&key, data, &mut output).unwrap();
+            assert_eq!(result.len(), tag_len);
+
+            // Verify the tag matches sign()
+            let tag = hmac::sign(&key, data);
+            assert_eq!(result, tag.as_ref());
```
justsmth: NP: 
* Verify that the contents of `output` match. (This is impossible to fail with the current implementation, but we should ensure that this behavior never changes.)
* Round-trip through `hmac::verify`:
```rust
            // Verify the tag matches sign()
            let tag = hmac::sign(&key, data);
            assert_eq!(result, tag.as_ref());
            assert_eq!(output.as_slice(), tag.as_ref());
            assert!(hmac::verify(&key, data, output.as_slice()).is_ok());
```

fabrice102: I'm also adding it for the `large_output` to be more thorough.

fabrice102:
Manually cherry-picking #1061 to fix the valgrind false positive and unlock CI.

PR #1059, opened 2026-03-10
Add `wasm32-unknown-emscripten` build and test support

justsmth:
### Description of changes:

Adds build and test support for `wasm32-unknown-emscripten`, following the addition of Emscripten support in AWS-LC.

### Call-outs:

- **`docker/emscripten/Dockerfile`:** The cross-rs emscripten base image bundles a Node.js too old to support the WebAssembly Exception Handling proposal that modern rustc emits for this target, so this Dockerfile replaces both the Emscripten SDK (3.1.74) and Node.js (20 LTS). `/emsdk` is made world-writable so the non-root cross-rs build user can write to the Emscripten cache.
- **`.cargo/config.toml` stack size bump:** Emscripten's 64 KB default stack overflows during regex compilation in our tests; raised to 1 MB.
- **`signature.rs` doctest:** Uses a runtime `cfg!(target_arch = "wasm32")` guard rather than `#[cfg(...)]` so the doctest still renders and compiles unchanged on every target — only the filesystem reads are skipped on WASM, where `tests/data/` isn't in the virtual FS.


builder/cc_builder.rs
```diff
@@ -295,7 +295,7 @@ impl CcBuilder {
         if !is_like_msvc {
             build_options.push(BuildOption::flag("-Wno-unused-parameter"));
             build_options.push(BuildOption::flag("-pthread"));
```
nhatnghiho: Should we exclude pthread for wasm?

nhatnghiho (dismissed):

builder/cc_builder.rs
justsmth: Good call. Yes -- we should drop `-pthread` for the wasm/emscripten target, and I've pushed a commit doing that.

aws-lc-rs/tests/aead_test.rs
```diff
@@ -582,6 +582,7 @@ fn test_aead_traits() {
 }
 
 #[test]
+#[cfg(not(target_arch = "wasm32"))]
```
nhatnghiho, line 586: minor nit: `#[cfg(not(target_arch = "wasm32"))]` doesn't communicate why threads are disabled, and future thread tests could miss this gate. Consider adding a comment, or some cfg flag emitted by the build script (something that says "no threads"?) to make it more intuitive than an architecture check.

justsmth: `std::thread::spawn` is a compile-error on `wasm32-unknown-emscripten` (there's no native threading unless you opt into `-pthread`), so these tests literally don't compile on the target. It's not a runtime-behavioral thing we might forget to check.

I'll add a `// wasm32 has no std::thread support` comment on each gate for clarity.

nhatnghiho (approved):

dougch (approved):

PR #1071, opened 2026-03-20
fix: follow symlinks when classifying include directory entries

cmtm:
`DirEntry::file_type()` on Linux uses `d_type` from `readdir(3)`, which returns `DT_LNK` for symlinks. This causes `is_file()` to return false for symlinked files, so they fall through to `fs_extra::dir::copy()` which panics with "Path is not a directory!".

This breaks builds under Bazel (and potentially other build systems) where source files in `CARGO_MANIFEST_DIR` are symlinks into a content-addressable store.

Replace `child.file_type().map_or(false, |t| t.is_file())` with `child.path().is_file()`, which calls `fs::metadata()` under the hood and follows symlinks.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -700,6 +700,34 @@ jobs:
       - name: Run tests
         run: cargo test -p aws-lc-rs --all-targets
 
+  symlinked-includes:
+    if: github.repository_owner == 'aws'
+    name: Symlinked includes
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: 'recursive'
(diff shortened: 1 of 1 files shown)


hnefatl:
> This breaks builds under Bazel (and potentially other build systems)

Also breaks builds under Nix, due to the same symlink-into-a-store behaviour.

justsmth:
Thanks for the PR!  I pushed another commit onto this branch that adds a CI job to detect future regressions.

We also have [a failing CI job](https://github.com/aws/aws-lc-rs/actions/runs/23442905366/job/68198601999?pr=1071) that requires that PRs contain the following statement in their description.

If you agree, would you add the statement above to the description of this PR?

PR #1081, opened 2026-03-31
MSVC Fix: use 8.3 short paths on Windows to avoid MAX_PATH limits in building script

walter-zeromatter:
When building aws-lc-sys with deeply nested source trees (e.g., Bazel external repository cache), the fully-qualified paths to manifest_dir and out_dir can exceed Windows' 260-character MAX_PATH limit, causing cl.exe invocations to fail.

Adds a `to_short_path()` helper that converts paths to their 8.3 short form via GetShortPathNameW on Windows (no-op on other platforms), and apply it to manifest_dir and out_dir in the build script's main().

### Testing:
I've tested this when building on windows via bazel with rules_rust & rules_rs

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,3 +12,7 @@ updates:
     ignore:
       - dependency-name: "untrusted"
         versions: ["0.7.1"]
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -28,7 +28,7 @@ jobs:
     if: github.repository_owner == 'aws'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
         with:
           submodules: 'recursive'
       - uses: dtolnay/rust-toolchain@master
@@ -55,7 +55,7 @@ jobs:
           - "--features bindgen,unstable,fips"
           - "--no-default-features --features aws-lc-sys"
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
         with:
           submodules: 'recursive'
       - uses: dtolnay/rust-toolchain@master
@@ -80,7 +80,7 @@ jobs:
         crate_dir: [ "aws-lc-sys", "aws-lc-fips-sys", "aws-lc-rs" ]
         diff_target: [ "branch", "published" ]
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
           submodules: 'recursive'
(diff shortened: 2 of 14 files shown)


builder/main.rs
```diff
@@ -818,7 +853,7 @@ fn main() {
     prepare_cargo_cfg();
 
     let manifest_dir = current_dir();
-    let manifest_dir = dunce::canonicalize(Path::new(&manifest_dir)).unwrap();
+    let manifest_dir = to_short_path(&dunce::canonicalize(Path::new(&manifest_dir)).unwrap());
```
justsmth: Could we conditionally compile this line and avoid the need to define `to_short_path` when targeting other platforms?
```rust
#[cfg(windows)]
let manifest_dir = to_short_path(&dunce::canonicalize(Path::new(&manifest_dir)).unwrap());
```

builder/main.rs
```diff
@@ -476,7 +476,42 @@ fn target_underscored() -> String {
...
+    if len == 0 {
+        return path.to_path_buf();
+    }
+    let mut buf = vec![0u16; len as usize];
+    let result = unsafe { GetShortPathNameW(wide.as_ptr(), buf.as_mut_ptr(), len) };
+    if result == 0 {
+        return path.to_path_buf();
+    }
+    buf.truncate(result as usize);
+    PathBuf::from(std::ffi::OsString::from_wide(&buf))
```
justsmth: We could potentially emit a warning if we see the build failure is likely:
```rust
    let short_path = PathBuf::from(std::ffi::OsString::from_wide(&buf));

    const MAX_PATH: usize = 260;
    if original_len >= MAX_PATH && (result as usize) >= MAX_PATH {
        emit_warning(format!(
            "Path length ({}) exceeds MAX_PATH ({}) and 8.3 short name conversion was ineffective. \
             8.3 short names may be disabled on this volume. \
             See: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/fsutil-8dot3name",
            original_len, MAX_PATH,
        ));
    }

    short_path
```

builder/main.rs
```diff
@@ -476,7 +476,42 @@ fn target_underscored() -> String {
 }
 
 fn out_dir() -> PathBuf {
-    PathBuf::from(cargo_env("OUT_DIR"))
+    to_short_path(&PathBuf::from(cargo_env("OUT_DIR")))
+}
```
justsmth, line 483: Conditional compilation for calling `to_short_path`?
```rust
fn out_dir() -> PathBuf {
    let out = PathBuf::from(cargo_env("OUT_DIR"));
    #[cfg(windows)]
    let out = to_short_path(&out);
    out
}
```

justsmth:
Thanks again for the PR! FYI -- I pushed a commit onto this branch to address my feedback above.

walter-zeromatter:
TY for all the reviews & fixes!

walter-zeromatter:
The failing check looks like a flake

justsmth (approved):

walter-zeromatter:
thanks for getting these merged!

PR #1086, opened 2026-03-31
MSVC Fix: Improve bad intrinsic check on msvc

walter-zeromatter:
### Issues:
None
### Description of changes: 
Fixes the cc_builder check for builtin swap intrinsics to better match cmake behavior and avoid link-time failures on MSVC. the `builtin_swap_check` passes, but AWS_LC_BUILTIN_SWAP_SUPPORTED should never be set for MSVC.

### Call-outs:
Only reproducible on Windows.

### Testing:
I've tested locally on windows when compiling via bazel

### DIFF
--- a/builder/cc_builder.rs
+++ b/builder/cc_builder.rs
@@ -750,13 +750,10 @@ impl CcBuilder {
         // try_compile_intermediates succeeds but __builtin_bswap* are unresolved at
         // link time. Without this guard the define is set and crypto/internal.h skips
         // the correct _MSC_VER path that uses _byteswap_* intrinsics.
+        if target_env() != "msvc"
+            && self.compiler_check("builtin_swap_check", Vec::<&'static str>::new())
         {
-            let compiler = cc::Build::default().get_compiler();
-            if (compiler.is_like_gnu() || compiler.is_like_clang())
-                && self.compiler_check("builtin_swap_check", Vec::<&'static str>::new())
-            {
-                cc_build.define("AWS_LC_BUILTIN_SWAP_SUPPORTED", Some("1"));
-            }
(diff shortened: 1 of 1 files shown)


builder/cc_builder.rs
```diff
@@ -746,8 +746,17 @@ impl CcBuilder {
...
+        // link time. Without this guard the define is set and crypto/internal.h skips
+        // the correct _MSC_VER path that uses _byteswap_* intrinsics.
+        {
+            let compiler = cc::Build::default().get_compiler();
+            if (compiler.is_like_gnu() || compiler.is_like_clang())
+                && self.compiler_check("builtin_swap_check", Vec::<&'static str>::new())
+            {
+                cc_build.define("AWS_LC_BUILTIN_SWAP_SUPPORTED", Some("1"));
+            }
         }
```
justsmth, line 757: I think it would be cleaner to condition the check on whether we are targeting an MSVC enviroment:
```
if target_env() != "msvc"
    && self.compiler_check("builtin_swap_check", Vec::<&'static str>::new())
{
    cc_build.define("AWS_LC_BUILTIN_SWAP_SUPPORTED", Some("1"));
}
```

justsmth (commented):
Thanks for the PR! I had one comment, but otherwise the change is good.

justsmth:
Thanks again for the PR.  I pushed a change onto the branch so the condition is on `target_env() != "msvc"`, which simplifies it a bit.

justsmth (approved):

walter-zeromatter:
Sorry for being slow getting back to you, and thanks for picking up the changes!

walter-zeromatter:
These build failures seem unrelated to my changes?

samuel40791765 (approved):

PR #1087, opened 2026-04-01
Strip LTO flags from CFLAGS for FIPS builds

skmcgrail:
Build environments like RPM mock chroots (e.g. AL2023) export CFLAGS with `-flto=auto -ffat-lto-objects`. These flags propagate through the cc crate into cmake-rs and ultimately into CMAKE_C_FLAGS. The FIPS delocator pipeline compiles C to assembly using `-S`, and the combination of `-S` with `-flto` causes GCC to produce a bcm.o containing R_X86_64_32 relocations that are incompatible with PIE executables.

The aws-lc CMakeLists.txt already acknowledges this class of issue: "-flto causes object files to contain LLVM bitcode. Mixing those with assembly output in the same static library breaks the linker" and disables assembly entirely when LTO is active for CFI builds.

Strip -flto* and -f{no-,}fat-lto-objects from CFLAGS before passing them to CMake for FIPS builds. Both the base CFLAGS and target-specific CFLAGS env vars must be overridden because the cc crate's envflags() accumulates flags from all prefix levels rather than replacing.

Add LTO CFLAGS to the hostile-environment CI matrix to prevent regressions.

### DIFF
--- a/builder/cmake_builder.rs
+++ b/builder/cmake_builder.rs
@@ -16,6 +16,25 @@ use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
+/// Strip LTO-related flags from a CFLAGS string. LTO flags interfere with the
+/// FIPS delocator pipeline, which compiles C to assembly (`-S`) and processes it
+/// with a Go tool. The combination of `-S` and `-flto` causes GCC to produce
+/// GIMPLE-annotated output that the delocator cannot handle. Additionally, LTO
+/// object files mixed with hand-written assembly in static libraries can cause
+/// linker errors (e.g. `R_X86_64_32` relocations incompatible with PIE).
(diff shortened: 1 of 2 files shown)


sgmenda (approved):
lgtm

torben-hansen (approved):

builder/cmake_builder.rs
```diff
@@ -178,7 +197,15 @@ impl CmakeBuilder {
...
-            set_env_for_target("CFLAGS", cflags);
+            let cflags = if is_fips_build() {
+                strip_lto_flags(&cflags)
+            } else {
+                cflags
+            };
+            // Set both the target-specific and base CFLAGS. cmake-rs reads
+            // the base CFLAGS env var directly, so we must override it too.
+            set_env_for_target("CFLAGS", &cflags);
+            set_env("CFLAGS", &cflags);
```
torben-hansen, line 208: Redundant copy for non-FIPS build. But nit.

Issue #1055, opened 2026-03-04
Add API to convert `rsa::PublicKeyComponents` into `signature::ParsedPublicKey` and/or DER encoded bytes


LukasKalbertodt:
### Problem:

If I have RSA public key components `n` and `e`, I can create a `rsa::PublicKeyComponents` value and also use it to verify signatures. Nice! However, I use `signature::ParsedPublicKey` everywhere else, so it would be nice to convert the former to the latter (of course, providing a [`VerificationAlgorithm`](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/signature/trait.VerificationAlgorithm.html) as well). 

### Solution:

Something like would work:

```
impl ParsedPublicKey {
    fn from_rsa_components(
        algorithm: &'static dyn VerificationAlgorithm,
        components: rsa::PublicKeyComponents,
    ) -> Result<Self, _>;
}
```

I expect that the actual implementation is already somewhere deep in the code base and this would add hardly any "real" code.

justsmth:
Hello! [aws-lc-rs v1.17.0](https://github.com/aws/aws-lc-rs/releases/tag/v1.17.0) has been released, which contains the API change requested here. 

This issue will now be closed. Please let us know if you have any other concerns after upgrading. Thanks again for your contribution to our library!

PR #1105, opened 2026-04-22
Add AsDer and ParsedPublicKey conversion for RSA PublicKeyComponents

MavenRain:
### Issues:
Addresses #1055

### Description of changes:
Implements `AsDer<PublicKeyX509Der<'static>>` for `rsa::PublicKeyComponents`, which builds an `EVP_PKEY` from the `(n, e)` components and marshals it as an X.509 `SubjectPublicKeyInfo`. Also adds `PublicKeyComponents::to_parsed_public_key`, which takes an `&'static RsaParameters` and returns a `signature::ParsedPublicKey` built directly from the same `EVP_PKEY`, skipping a DER encode/decode round-trip. Placing the conversion on the source type is symmetric with the existing `PublicKeyComponents::verify(params, ...)` method, keeps RSA-specific code out of `signature.rs`, and narrows `params` to `RsaParameters` so non-RSA algorithms are rejected at compile time rather than at runtime. Malformed components (empty or leading-zero `n`/`e`) are rejected by the existing `build_rsa()` helper and surface as `KeyRejected::inconsistent_components()` on both the `to_parsed_public_key` and `AsDer` paths.

### Call-outs:
Consistent with `ParsedPublicKey::new`, `to_parsed_public_key` does not enforce the algorithm's accepted bit-size range at construction — that check is deferred to `verify_sig`. This is documented on the new method and pinned down by a test.

### Testing:
unit tests (PKCS#1 v1.5 and PSS verification; non-RSA algorithm, empty-modulus, and below-range rejection; X.509 round-trip covered both in the default feature configuration and under `ring-io`)

### DIFF
--- a/aws-lc-rs/src/signature.rs
+++ b/aws-lc-rs/src/signature.rs
@@ -452,6 +452,50 @@ impl ParsedPublicKey {
         parse_public_key(bytes.as_ref(), algorithm)
     }
 
+    /// Creates a new `ParsedPublicKey` from the components of an RSA public key.
+    ///
+    /// `components` carries the public modulus `n` and public exponent `e` in
+    /// big-endian bytes without leading zeros. `algorithm` must be an RSA
+    /// verification algorithm such as [`RSA_PKCS1_2048_8192_SHA256`] or
+    /// [`RSA_PSS_2048_8192_SHA256`]; any other verification algorithm is
+    /// rejected.
+    ///
+    /// # Errors
+    /// `KeyRejected` if `algorithm` is not an RSA verification algorithm, or
+    /// if `components` do not form a valid RSA public key.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use aws_lc_rs::signature::{self, ParsedPublicKey, RsaPublicKeyComponents};
+    ///
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// # fn modulus_bytes() -> &'static [u8] { &[] }
+    /// # fn exponent_bytes() -> &'static [u8] { &[] }
+    /// # fn signature_bytes() -> &'static [u8] { &[] }
+    /// // Public key components (big-endian, no leading zeros) received
+    /// // out-of-band from a peer.
+    /// let components = RsaPublicKeyComponents {
+    ///     n: modulus_bytes(),
+    ///     e: exponent_bytes(),
+    /// };
+    /// let parsed = ParsedPublicKey::from_rsa_components(
+    ///     &signature::RSA_PKCS1_2048_8192_SHA256,
+    ///     &components,
+    /// )?;
+    /// parsed.verify_sig(b"hello, world", signature_bytes())?;
+    /// # Ok(()) }
+    /// ```
+    pub fn from_rsa_components<B: AsRef<[u8]> + Debug>(
+        algorithm: &'static dyn VerificationAlgorithm,
+        components: &rsa::PublicKeyComponents<B>,
+    ) -> Result<Self, KeyRejected> {
+        let der = components
+            .as_der()
+            .map_err(|Unspecified| KeyRejected::invalid_encoding())?;
+        parse_public_key(der.as_ref(), algorithm)
+    }
+
     /// Returns the algorithm used by this public key.
     #[must_use]
     pub fn algorithm(&self) -> &'static dyn VerificationAlgorithm {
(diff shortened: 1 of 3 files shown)


aws-lc-rs/src/signature.rs
```diff
@@ -452,6 +452,50 @@ impl ParsedPublicKey {
...
+    /// parsed.verify_sig(b"hello, world", signature_bytes())?;
+    /// # Ok(()) }
+    /// ```
+    pub fn from_rsa_components<B: AsRef<[u8]> + Debug>(
+        algorithm: &'static dyn VerificationAlgorithm,
+        components: &rsa::PublicKeyComponents<B>,
+    ) -> Result<Self, KeyRejected> {
+        let der = components
+            .as_der()
+            .map_err(|Unspecified| KeyRejected::invalid_encoding())?;
```
justsmth: I think "inconsistent components" would be more accurate:
```
.map_err(|Unspecified| KeyRejected::inconsistent_components())?;
```

aws-lc-rs/tests/rsa_test.rs
```diff
@@ -331,6 +331,125 @@ fn test_signature_rsa_primitive_verification() {
...
+fn test_parsed_public_key_from_rsa_components_rejects_non_rsa_algorithm() {
+    let n: [u8; 4] = [0x12, 0x34, 0x56, 0x79];
+    let e: [u8; 3] = [0x01, 0x00, 0x01];
+    let components = RsaPublicKeyComponents {
+        n: n.as_slice(),
+        e: e.as_slice(),
+    };
+    let err = ParsedPublicKey::from_rsa_components(&signature::ED25519, &components)
+        .expect_err("Ed25519 is not an RSA algorithm");
+    let _ = err;
```
justsmth: We can be more specific about the expected error:
```
assert_eq!(err.description_(), "WrongAlgorithm");
```

aws-lc-rs/tests/rsa_test.rs
```diff
@@ -331,6 +331,125 @@ fn test_signature_rsa_primitive_verification() {
...
+    let components = RsaPublicKeyComponents {
+        n: n.as_slice(),
+        e: e.as_slice(),
+    };
+    let err = ParsedPublicKey::from_rsa_components(
+        &signature::RSA_PKCS1_2048_8192_SHA256,
+        &components,
+    )
+    .expect_err("empty modulus should be rejected");
+    let _ = err;
```
justsmth: More specific error check:
```
assert_eq!(err.description_(), "InconsistentComponents");
```

aws-lc-rs/src/signature.rs
```diff
@@ -452,6 +452,50 @@ impl ParsedPublicKey {
...
+    ///     &signature::RSA_PKCS1_2048_8192_SHA256,
+    ///     &components,
+    /// )?;
+    /// parsed.verify_sig(b"hello, world", signature_bytes())?;
+    /// # Ok(()) }
+    /// ```
+    pub fn from_rsa_components<B: AsRef<[u8]> + Debug>(
+        algorithm: &'static dyn VerificationAlgorithm,
+        components: &rsa::PublicKeyComponents<B>,
+    ) -> Result<Self, KeyRejected> {
```
justsmth: Instead of this function, I think it would be better to have an `into_parsed` function on `PublicKeyComponents`.
```rust
impl<B: AsRef<[u8]> + Debug> PublicKeyComponents<B> {
    pub fn into_parsed(&self, algorithm: &'static dyn VerificationAlgorithm)
            -> Result<ParsedPublicKey, KeyRejected> { 
        ... 
    }
}
```

justsmth (commented):
Instead of `ParsedPublicKey::from_rsa_components`, I think a `PublicKeyComponents::into_parsed` would have a better aesthetic and achieve the same goal.

There are a few more edge cases to cover in testing, but it can wait until the public API is decided upon.

justsmth:
I wanted to let you know that I updated your PR based upon the feedback I provided.  Please let me know if you have any concerns with these changes. Thanks!

justsmth (approved):

justsmth:
Clippy CI failure addressed in separate PR: https://github.com/aws/aws-lc-rs/pull/1112

MavenRain:
> Clippy CI failure addressed in separate PR: #1112

Looks addressed now

WillChilds-Klein (approved):

Issue #1097, opened 2026-04-10
The CPU Jitter random number generator must not be compiled with optimizations.


yurivict:
### Problem:

Build of the qdrant project version 1.17.1 fails:
```
warning: aws-lc-sys@0.38.0: /wrkdirs/usr/ports/databases/qdrant/work/qdrant-1.17.1/cargo-crates/aws-lc-sys-0.38.0/aws-lc/third_party/jitterentropy/jitterentropy-library/src/jitterentropy-base.c:47:3: error: "The CPU Jitter random number generator must not be compiled with optimizations. See documentation. Use the compiler switch -O0 for compiling jitterentropy.c."
warning: aws-lc-sys@0.38.0:    47 |  #error "The CPU Jitter random number generator must not be compiled with optimizations. See documentation. Use the compiler switch -O0 for compiling jitterentropy.c."
warning: aws-lc-sys@0.38.0:       |   ^
warning: aws-lc-sys@0.38.0: 1 error generated.
```

#### Relevant details
AWS-LC for Rust versions or commit: (6b1bce0...)

System information: for linux, below info can be collected by running `uname -srvmp`
 * CPU architecture: x86-64
 * CPU name: AMD Ryzen 9 9950X 16-Core Processor
 * OS: FreeBSD 15

justsmth:
Hello! Sorry about this build failure.

As a quick workaround, you can export `AWS_LC_SYS_NO_JITTER_ENTROPY=1` in your environment to avoid this problem.

It looks like this is a FreeBSD ports build. We do have [mitigations](https://github.com/aws/aws-lc-rs/blob/main/builder/cc_builder.rs#L467-L487) that strip optimization flags from `CFLAGS` before compiling jitterentropy, but something in your build environment is apparently getting past that. Could you share the values of `CFLAGS` and any target-specific variants (e.g., `CFLAGS_x86_64_unknown_freebsd`, `TARGET_CFLAGS`) that are set during the build? That would help us figure out what we're missing.

Thanks for the report!

yurivict:
AWS_LC_SYS_NO_JITTER_ENTROPY=1 works, thanks!

This is a FreeBSD port build.

justsmth:
I don't have easy access to a FreeBSD host right now. Could you share the values of CFLAGS and any target-specific variants (e.g., CFLAGS_x86_64_unknown_freebsd, TARGET_CFLAGS) that are set during the build? That would help us figure out what we're missing.

[...]

PR #1106, opened 2026-04-22
Fix jitterentropy CFLAGS filtering for HOST_CFLAGS and TARGET_CFLAGS

justsmth:
### Issues:
* #1097

### Description of changes:
The cc crate (v1.2+) collects flags from ALL matching CFLAGS env vars — not just the highest-priority one. It checks `CFLAGS_{target}`, `HOST_CFLAGS` (native builds) or `TARGET_CFLAGS` (cross builds), and `CFLAGS`, appending flags from each in priority order. The previous guard only filtered and overrode `CFLAGS_{target}`, so optimization flags arriving via `HOST_CFLAGS` or `TARGET_CFLAGS` would land after our `-O0` on the command line and win.

The fix filters optimization flags from every variable the cc crate reads: `CFLAGS_{target}`, `HOST_CFLAGS`, `TARGET_CFLAGS`, and `CFLAGS`.

### Call-outs:
We filter both `HOST_CFLAGS` and `TARGET_CFLAGS` unconditionally rather than trying to mirror the cc crate's host-vs-cross selection logic. Filtering a variable the cc crate doesn't read for a given build is harmless, and this avoids coupling to the cc crate's internal branching.

The old `hostile-environment` CI matrix had an unused `os` dimension (`runs-on` was hardcoded to `ubuntu-latest` regardless of `matrix.os`). The new matrix drops that and instead parameterizes the env var name, covering `CFLAGS`, `HOST_CFLAGS`, and `TARGET_CFLAGS`.

### Testing:
The `hostile-environment` CI job now covers `HOST_CFLAGS="-O2"` and `TARGET_CFLAGS="-O2"` — both of which fail on `main` and pass with this change. Verified locally that all CFLAGS variants (`CFLAGS`, `HOST_CFLAGS`, `TARGET_CFLAGS`, and combinations) build successfully with `-O2`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -756,16 +756,19 @@ jobs:
 
   hostile-environment:
     if: github.repository_owner == 'aws'
-    name: Hostile environment
+    name: Hostile environment - ${{ matrix.env_name }}=${{ matrix.env_value }}
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, macos-latest ]
-        cflags:
-          - '-O3'
-          - '-flto=auto -ffat-lto-objects'
-    env:
-      CFLAGS: ${{ matrix.cflags }}
+        include:
+          - env_name: CFLAGS
+            env_value: '-O3'
+          - env_name: CFLAGS
+            env_value: '-flto=auto -ffat-lto-objects'
+          - env_name: HOST_CFLAGS
+            env_value: '-O2'
+          - env_name: TARGET_CFLAGS
+            env_value: '-O2'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -756,16 +756,19 @@ jobs:
...
+        include:
+          - env_name: CFLAGS
+            env_value: '-O3'
+          - env_name: CFLAGS
+            env_value: '-flto=auto -ffat-lto-objects'
+          - env_name: HOST_CFLAGS
+            env_value: '-O2'
+          - env_name: TARGET_CFLAGS
+            env_value: '-O2'
     runs-on: ubuntu-latest
```
WillChilds-Klein, line 772: why 02 instead of some other level of optimization?

builder/cc_builder.rs
```diff
@@ -470,21 +471,44 @@ impl CcBuilder {
...
+                .split_whitespace()
+                .filter(|flag| !flag.starts_with("-O") && !flag.starts_with("/O"))
+                .collect::<Vec<_>>()
+                .join(" ");
+            if is_like_msvc {
+                format!("{filtered} -Od").trim().to_string()
+            } else {
+                format!("{filtered} -O0 -U_FORTIFY_SOURCE")
+                    .trim()
+                    .to_string()
```
WillChilds-Klein, line 501: do either of these need to account for windows flag format `/0` as L493 does?

WillChilds-Klein (approved):

.github/workflows/tests.yml
justsmth: Any optimization flag would work for this test, it's just the one we picked..

builder/cc_builder.rs
justsmth: This step just appends onto the end of the value a flag to disable optimizations -- this is a "belt and suspenders" since optimization flags should have been removed already on the statement above.  This addition ensures that we've indicated that optimization should be disabled even if we missed the removal of a flag that would've implied otherwise.

The `/O` is only needed to detect parameters for MSVC, but (IIRC) MSVC also accepts the `-O` format for options, which is what we use for `-Od`.

dougch (approved):

PR #1109, opened 2026-04-23
Add opt-in 2TDEA and 3TDEA cipher support behind a `legacy-3des` feature

qzh0223:
### Description of changes:
Adds 2-key and 3-key Triple DES (`DES_EDE_FOR_LEGACY_USE_ONLY` and `DES_EDE3_FOR_LEGACY_USE_ONLY`) to the `cipher` module, gated behind a new opt-in Cargo feature `legacy-3des`. Only CBC and ECB are supported; CTR and CFB128 are rejected at key-construction time.

Everything about 3DES is hidden by default: the module, the `AlgorithmId` variants, the `SymmetricCipherKey` variants, the new `Iv64` cipher-context variant, and all tests are `#[cfg(feature = "legacy-3des")]`-gated. When the feature is enabled, every public item is also `#[deprecated]` so downstream use sites produce a compiler warning, and the naming follows the existing `cmac::TDES_FOR_LEGACY_USE_ONLY` convention.

Key validation is stricter than what the underlying `EVP_des_*` primitives would do on their own:
* Weak DES subkeys (the four SP 800-67 weak keys) are rejected via `DES_set_key` for every subkey.
* 2TDEA requires K1 ≠ K2.
* 3TDEA requires K1, K2, and K3 to be pairwise distinct. Callers who want the `K1 ‖ K2 ‖ K1` form must use the 2TDEA variant rather than encoding 2TDEA as a 24-byte 3TDEA key.

The streaming cipher path passes raw key bytes to the EVP layer (which internally uses `DES_set_key_unchecked`), so a new `UnboundCipherKey::validate_key_material` step runs the same checks before EVP initialization. For non-DES algorithms it is a no-op.

FIPS builds are supported: the native DES calls are wrapped in `indicator_check!`, which observes that the approved-operation counter does not advance and correctly marks the service indicator as unapproved.

### Call-outs:
* **Feature flag posture is intentionally stricter than `cmac`.** `cmac::TDES_FOR_LEGACY_USE_ONLY` is currently unconditional and not deprecation-marked. This PR opts for the opposite (feature-gated, `#[deprecated]`) because the stated expectation is that 3DES should be clearly legacy in the API surface. Happy to discuss whether `cmac` should follow in a separate change.
* **`all-bindings` propagation.** The low-level `DES_*` symbols live in `openssl/des.h`, which is outside the default bindgen allowlist, so `legacy-3des` forces `aws-lc-sys?/all-bindings` on. `aws-lc-fips-sys` always exposes the full binding set, so nothing extra is needed for the FIPS path. The `EVP_des_*` symbols used on the streaming/cmac paths live in `evp.h` and are always available.
* **Small internal error-type refactor.** `SymmetricCipherKey::aes128/192/256/chacha20` and `TryInto<SymmetricCipherKey> for UnboundCipherKey` now return `KeyRejected` instead of `Unspecified`, so the DES validation path can surface a more specific error internally. All of these items are `pub(crate)`, and the `From<KeyRejected> for Unspecified` conversion handles the boundary at the public API, so this is not an API break. Worth a look during review since it is technically unrelated to DES.
* **New `Algorithm::supports_mode`.** Used to reject `(algorithm, mode)` combinations (e.g. 3DES + CTR) at key-construction time rather than deferring to the first encrypt/decrypt call. This is called from `EncryptingKey::new`, `DecryptingKey::new`, the padded variants, and the streaming variants.
* **No 3DES support was added to `aead::quic`.** QUIC header protection doesn't use 3DES; `cipher_new_mask` treats the new `SymmetricCipherKey` variants as an error.

### Testing:
Tested only with the `legacy-3des` feature enabled (default builds are unchanged and covered by the existing test suite).

* SP 800-67 known-answer tests for 2TDEA and 3TDEA in CBC and ECB mode, both via the one-shot `EncryptingKey`/`DecryptingKey` API and via `StreamingEncryptingKey`/`StreamingDecryptingKey` with step sizes 2 through 9 to exercise the EVP update boundaries.
* Round-trip tests on non-block-aligned plaintext for PKCS#7 CBC/ECB.
* Negative tests for all four DES weak keys, rotated through each of the three subkey positions and padded with distinct non-weak fillers so the pairwise-distinctness check does not short-circuit the weak-key check.
* Negative tests for every degenerate 3TDEA configuration (K1 == K2, K2 == K3, K1 == K3) and the 2TDEA K1 == K2 case.
* Negative tests for CTR and CFB128 on both encrypt and decrypt paths.
* Full library test suite passes with and without `legacy-3des`, and under `--features fips,legacy-3des`.
* `cargo clippy` is clean with the feature enabled.

### DIFF
--- a/aws-lc-rs/README.md
+++ b/aws-lc-rs/README.md
@@ -124,6 +124,27 @@ It can be enabled in two ways:
 used in production builds. The `rand::unsealed` module and `mut_fill` method are not part of the
 stable public API and may change without notice.
 
+##### legacy-3des
+
+Enables Triple DES as an opt-in symmetric cipher under the `cipher` module, exposing
+both `DES_EDE_FOR_LEGACY_USE_ONLY` (2-key Triple DES / DES-EDE) and
+`DES_EDE3_FOR_LEGACY_USE_ONLY` (3-key Triple DES / DES-EDE3), together with the
+supporting key-length and IV-length constants. Only CBC and ECB operating modes are
+supported.
+
+**⚠️ Warning:** Triple DES is a legacy algorithm. It has been disallowed for
+encryption by [NIST SP 800-131A Rev. 2](https://csrc.nist.gov/publications/detail/sp/800-131a/rev-2/final):
+2-key Triple DES since 2015, and 3-key Triple DES after 2023. This feature exists
+solely to support interoperability with existing systems that cannot yet migrate to
+AES. All exposed items are marked `#[deprecated]` so that any use site produces a
+compiler warning. **Do not use Triple DES in new designs.** If you only need
+confidentiality, prefer AES-GCM or another AEAD from the `aead` module; if you
+specifically need a block cipher, prefer one of the AES algorithms in `cipher`.
+
+This feature also enables `aws-lc-sys?/all-bindings`, which is required to expose the
+`DES_*` bindings when the consuming build uses `bindgen` rather than the pre-generated
+bindings.
+
 ## Use of prebuilt NASM objects
 
 Prebuilt NASM objects are **only** applicable to Windows x86-64 platforms. They are **never** used on any other platform (Linux, macOS, etc.).
--- a/aws-lc-rs/README.md
+++ b/aws-lc-rs/README.md
@@ -134,16 +134,17 @@ supported.
 
 **⚠️ Warning:** Triple DES is a legacy algorithm. It has been disallowed for
 encryption by [NIST SP 800-131A Rev. 2](https://csrc.nist.gov/publications/detail/sp/800-131a/rev-2/final):
-2-key Triple DES since 2015, and 3-key Triple DES after 2023. This feature exists
+2-key Triple DES after 2015, and 3-key Triple DES after 2023. This feature exists
 solely to support interoperability with existing systems that cannot yet migrate to
 AES. All exposed items are marked `#[deprecated]` so that any use site produces a
 compiler warning. **Do not use Triple DES in new designs.** If you only need
 confidentiality, prefer AES-GCM or another AEAD from the `aead` module; if you
 specifically need a block cipher, prefer one of the AES algorithms in `cipher`.
 
-This feature also enables `aws-lc-sys?/all-bindings`, which is required to expose the
-`DES_*` bindings when the consuming build uses `bindgen` rather than the pre-generated
-bindings.
+The universal bindings used by default from `aws-lc-sys` do not expose the `DES_*` symbols. Thus,
+this feature enables `aws-lc-sys?/all-bindings` so that bindings are available for all AWS-LC
+libcrypto functions. On platforms where pre-generated bindings are unavailable, `bindgen` is
+invoked at build time and the generated bindings will include the full symbol set.
 
 ## Use of prebuilt NASM objects
 
--- a/aws-lc-rs/README.md
+++ b/aws-lc-rs/README.md
@@ -135,11 +135,11 @@ supported.
 **⚠️ Warning:** Triple DES is a legacy algorithm. It has been disallowed for
 encryption by [NIST SP 800-131A Rev. 2](https://csrc.nist.gov/publications/detail/sp/800-131a/rev-2/final):
 2-key Triple DES after 2015, and 3-key Triple DES after 2023. This feature exists
-solely to support interoperability with existing systems that cannot yet migrate to
-AES. All exposed items are marked `#[deprecated]` so that any use site produces a
-compiler warning. **Do not use Triple DES in new designs.** If you only need
-confidentiality, prefer AES-GCM or another AEAD from the `aead` module; if you
-specifically need a block cipher, prefer one of the AES algorithms in `cipher`.
+solely to support interoperability. All exposed items are marked `#[deprecated]`
+so that any use site produces a compiler warning. **Do not use Triple DES in new
+designs.** If you only need confidentiality, prefer AES-GCM or another AEAD from
+the `aead` module; if you specifically need a block cipher, prefer one of the AES
+algorithms in `cipher`.
 
 The universal bindings used by default from `aws-lc-sys` do not expose the `DES_*` symbols. Thus,
 this feature enables `aws-lc-sys?/all-bindings` so that bindings are available for all AWS-LC
--- a/aws-lc-rs/src/cipher.rs
+++ b/aws-lc-rs/src/cipher.rs
@@ -489,8 +489,7 @@ pub const AES_256: Algorithm = Algorithm {
 /// 2-key Triple DES is a legacy algorithm and has been disallowed for encryption
 /// by [NIST SP 800-131A Rev. 2](https://csrc.nist.gov/publications/detail/sp/800-131a/rev-2/final)
 /// since 2015 — several years earlier than 3-key Triple DES. It is retained here
-/// solely for interoperability with existing systems that cannot yet migrate.
-/// New designs must use an AES-based algorithm.
+/// solely for interoperability. New designs must use an AES-based algorithm.
 ///
 /// Only CBC and ECB operating modes are supported. K1 must differ from K2; if
 /// they are equal the cipher degenerates to single-DES and key construction
@@ -1529,11 +1528,10 @@ mod tests {
     #[test]
     #[allow(deprecated)]
     fn test_des_ede3_rejects_degenerate_keys() {
-        // SP 800-67r2 Appendix A: 3TDEA requires K1, K2, and K3 to be
-        // independently chosen. We enforce that all three subkeys are
-        // pairwise distinct so the cipher cannot degenerate to single-DES
-        // (K1 == K2 or K2 == K3) or to 2TDEA (K1 == K3). Callers who want
-        // 2TDEA must use `DES_EDE_FOR_LEGACY_USE_ONLY` with a 16-byte key.
+        // SP 800-67 §2 (Keying Option 1) requires K1, K2, and K3 to be pairwise
+        // independent. We enforce that all three subkeys are distinct so the cipher
+        // cannot degenerate to single-DES (K1==K2 or K2==K3) or to 2TDEA (K1==K3).
+        // Callers who want 2TDEA must use `DES_EDE_FOR_LEGACY_USE_ONLY` with a 16-byte key.
         let k1_eq_k2 = from_hex("0123456789abcdef0123456789abcdef456789abcdef0123").unwrap();
         assert!(
             UnboundCipherKey::new(&DES_EDE3_FOR_LEGACY_USE_ONLY, &k1_eq_k2)
--- a/aws-lc-rs/src/cipher.rs
+++ b/aws-lc-rs/src/cipher.rs
@@ -509,8 +509,8 @@ pub const DES_EDE_FOR_LEGACY_USE_ONLY: Algorithm = Algorithm {
 ///
 /// 3DES is a legacy algorithm and has been disallowed for encryption by
 /// [NIST SP 800-131A Rev. 2](https://csrc.nist.gov/publications/detail/sp/800-131a/rev-2/final)
-/// after 2023. It is retained here solely for interoperability with existing
-/// systems that cannot yet migrate. New designs must use an AES-based algorithm.
+/// after 2023. It is retained here solely for interoperability. New designs
+/// must use an AES-based algorithm.
 ///
 /// Only CBC and ECB operating modes are supported. K1, K2, and K3 must all
 /// be distinct; any pairwise equality causes 3TDEA to degenerate (to either
--- a/aws-lc-rs/src/cipher.rs
+++ b/aws-lc-rs/src/cipher.rs
@@ -262,6 +262,7 @@ pub use crate::cipher::aes::AES_256_KEY_LEN;
 /// 2-key Triple DES is a legacy algorithm and has been disallowed for encryption
 /// by NIST SP 800-131A Rev. 2 since 2015. It is retained for interoperability only.
 #[cfg(feature = "legacy-3des")]
+#[cfg_attr(aws_lc_rs_docsrs, doc(cfg(feature = "legacy-3des")))]
 #[deprecated(
     note = "2-key Triple DES is a legacy algorithm retained for interoperability only; NIST SP 800-131A Rev. 2 disallows its use for encryption. Prefer an AES-based algorithm."
 )]
@@ -272,6 +273,7 @@ pub use crate::cipher::des::DES_EDE_KEY_LEN;
 /// 3DES is a legacy algorithm and has been disallowed for encryption by
 /// NIST SP 800-131A Rev. 2 after 2023. It is retained for interoperability only.
 #[cfg(feature = "legacy-3des")]
+#[cfg_attr(aws_lc_rs_docsrs, doc(cfg(feature = "legacy-3des")))]
 #[deprecated(
     note = "3DES is a legacy algorithm retained for interoperability only; NIST SP 800-131A Rev. 2 disallows its use for encryption. Prefer an AES-based algorithm."
 )]
@@ -293,6 +295,7 @@ pub use crate::cipher::aes::AES_CFB_IV_LEN;
 /// 3DES is a legacy algorithm and has been disallowed for encryption by
 /// NIST SP 800-131A Rev. 2 after 2023. It is retained for interoperability only.
 #[cfg(feature = "legacy-3des")]
+#[cfg_attr(aws_lc_rs_docsrs, doc(cfg(feature = "legacy-3des")))]
 #[deprecated(
     note = "3DES is a legacy algorithm retained for interoperability only; NIST SP 800-131A Rev. 2 disallows its use for encryption. Prefer an AES-based algorithm."
 )]
@@ -379,6 +382,7 @@ macro_rules! define_cipher_context {
 
             /// A 64-bit Initialization Vector (used by 3DES).
             #[cfg(feature = "legacy-3des")]
+            #[cfg_attr(aws_lc_rs_docsrs, doc(cfg(feature = "legacy-3des")))]
             Iv64(FixedLength<IV_LEN_64_BIT>),
 
             /// No Cipher Context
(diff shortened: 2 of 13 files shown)


aws-lc-rs/README.md
```diff
@@ -124,6 +124,27 @@ It can be enabled in two ways:
...
+2-key Triple DES since 2015, and 3-key Triple DES after 2023. This feature exists
+solely to support interoperability with existing systems that cannot yet migrate to
+AES. All exposed items are marked `#[deprecated]` so that any use site produces a
+compiler warning. **Do not use Triple DES in new designs.** If you only need
+confidentiality, prefer AES-GCM or another AEAD from the `aead` module; if you
+specifically need a block cipher, prefer one of the AES algorithms in `cipher`.
+
+This feature also enables `aws-lc-sys?/all-bindings`, which is required to expose the
+`DES_*` bindings when the consuming build uses `bindgen` rather than the pre-generated
+bindings.
```
justsmth: This is correct --  but it could be more clear:
```
The universal bindings used by default from `aws-lc-sys` do not expose the `DES_*` symbols. Thus, this feature enables `aws-lc-sys?/all-bindings` so that bindings are available for all AWS-LC libcrypto functions. On platforms where pre-generated bindings are unavailable, `bindgen` is invoked at build time and the generated bindings will include the full symbol set."
```

aws-lc-rs/src/cipher.rs
```diff
@@ -797,22 +1024,38 @@ fn encrypt(
...
+                des::encrypt_cbc_mode(key, context, in_out)
+            }
         },
         OperatingMode::CTR => match algorithm.id() {
             AlgorithmId::Aes128 | AlgorithmId::Aes192 | AlgorithmId::Aes256 => {
                 aes::encrypt_ctr_mode(key, context, in_out)
             }
+            #[cfg(feature = "legacy-3des")]
+            AlgorithmId::DesEdeForLegacyUseOnly | AlgorithmId::DesEde3ForLegacyUseOnly => {
+                Err(Unspecified)
```
justsmth: We should use `unreachable!` here as well, since reaching this condition would indicate a bug in our logic:
```
unreachable!("DES does not support CTR mode."
```

aws-lc-rs/src/cipher.rs
```diff
@@ -797,22 +1024,38 @@ fn encrypt(
...
+            }
         },
         // TODO: Hopefully support CFB1, and CFB8
         OperatingMode::CFB128 => match algorithm.id() {
             AlgorithmId::Aes128 | AlgorithmId::Aes192 | AlgorithmId::Aes256 => {
                 aes::encrypt_cfb_mode(key, mode, context, in_out)
             }
+            #[cfg(feature = "legacy-3des")]
+            AlgorithmId::DesEdeForLegacyUseOnly | AlgorithmId::DesEde3ForLegacyUseOnly => {
+                Err(Unspecified)
```
justsmth: Also here:
```
unreachable!("DES does not support CFB128 mode."
```

aws-lc-rs/src/cipher.rs
```diff
@@ -363,6 +432,20 @@ pub enum AlgorithmId {
...
+    /// NIST SP 800-131A Rev. 2 since 2015.
+    #[cfg(feature = "legacy-3des")]
+    DesEdeForLegacyUseOnly,
+
+    /// 3TDEA (for legacy use only).
+    ///
+    /// 3DES has been disallowed for encryption by NIST SP 800-131A Rev. 2
+    /// after 2023.
+    #[cfg(feature = "legacy-3des")]
+    DesEde3ForLegacyUseOnly,
```
justsmth, line 465: We should also mark these as deprecated.

aws-lc-rs/src/cipher/key.rs
```diff
@@ -113,6 +144,107 @@ impl SymmetricCipherKey {
...
+    }
+
+    #[cfg(feature = "legacy-3des")]
+    fn des_set_key(key_bytes: &[u8; 8]) -> Result<DES_key_schedule, KeyRejected> {
+        let mut ks = MaybeUninit::<DES_key_schedule>::uninit();
+        // DES_set_key return values (see openssl/des.h):
+        //   0  => key is not weak and has odd parity (preferred)
+        //  -1  => key parity is not odd (schedule is still valid; most callers
+        //         do not maintain DES parity bits, so we accept this)
+        //  -2  => key is a weak DES key; reject it
```
justsmth: Pedantic -- it also rejects "semi-weak" keys.
```
//  -2  => key is a weak or semi-weak DES key; reject it
```

aws-lc-rs/tests/cipher_test.rs
```diff
@@ -1111,6 +1114,523 @@ unpadded_cipher_kat!(
...
+            assert_eq!($mode, decrypting_key.mode());
+            assert_eq!($alg, decrypting_key.algorithm());
+            let plaintext = decrypting_key.decrypt(&mut in_out, context).unwrap();
+            assert_eq!(input.as_slice(), plaintext);
+        }
+
+        des_streaming_cipher_rt!($name, $alg, $mode, $constructor, $key, $plaintext, 2, 9);
+    };
+}
+
```
justsmth, line 1413: * We should specify the source for these KATs. Ideally, we should be using all of the KATs for the supported modes specified here by NIST: https://csrc.nist.gov/projects/cryptographic-standards-and-guidelines/example-values
* Would it be better to read these out of a data file?
* Do we have any tests covering rejection of "semi-weak" keys?

aws-lc-rs/src/cipher.rs
```diff
@@ -838,22 +1081,38 @@ fn decrypt<'in_out>(
...
+                des::decrypt_cbc_mode(key, context, in_out)
+            }
         },
         OperatingMode::CTR => match algorithm.id() {
             AlgorithmId::Aes128 | AlgorithmId::Aes192 | AlgorithmId::Aes256 => {
                 aes::decrypt_ctr_mode(key, context, in_out)
             }
+            #[cfg(feature = "legacy-3des")]
+            AlgorithmId::DesEdeForLegacyUseOnly | AlgorithmId::DesEde3ForLegacyUseOnly => {
+                Err(Unspecified)
```
justsmth:
```
unreachable!("DES does not support CTR mode.")
```

aws-lc-rs/src/cipher.rs
```diff
@@ -838,22 +1081,38 @@ fn decrypt<'in_out>(
...
+            }
         },
         // TODO: Hopefully support CFB1, and CFB8
         OperatingMode::CFB128 => match algorithm.id() {
             AlgorithmId::Aes128 | AlgorithmId::Aes192 | AlgorithmId::Aes256 => {
                 aes::decrypt_cfb_mode(key, mode, context, in_out)
             }
+            #[cfg(feature = "legacy-3des")]
+            AlgorithmId::DesEdeForLegacyUseOnly | AlgorithmId::DesEde3ForLegacyUseOnly => {
+                Err(Unspecified)
```
justsmth:
```
unreachable!("DES does not support CFB128 mode.")
```

aws-lc-rs/README.md
```diff
@@ -124,6 +124,27 @@ It can be enabled in two ways:
...
+
+Enables Triple DES as an opt-in symmetric cipher under the `cipher` module, exposing
+both `DES_EDE_FOR_LEGACY_USE_ONLY` (2-key Triple DES / DES-EDE) and
+`DES_EDE3_FOR_LEGACY_USE_ONLY` (3-key Triple DES / DES-EDE3), together with the
+supporting key-length and IV-length constants. Only CBC and ECB operating modes are
+supported.
+
+**⚠️ Warning:** Triple DES is a legacy algorithm. It has been disallowed for
+encryption by [NIST SP 800-131A Rev. 2](https://csrc.nist.gov/publications/detail/sp/800-131a/rev-2/final):
+2-key Triple DES since 2015, and 3-key Triple DES after 2023. This feature exists
```
justsmth: NP: I believe saying "after 2015" would be more accurate.

aws-lc-rs/src/cipher/des.rs
```diff
@@ -0,0 +1,187 @@
...
+    // The caller (`encrypt`/`decrypt` in cipher.rs) validates block alignment
+    // for CBC mode; assert here as a safety net.
+    debug_assert_eq!(in_out.len() % DES_BLOCK_LEN, 0);
+
+    // DES is not FIPS-approved; `indicator_check!` will observe that the
+    // underlying call does not increment the approved-operation counter and
+    // correctly mark the service indicator as unapproved in FIPS builds.
+    //
+    // SAFETY: `DES_ede3_cbc_encrypt` supports in-place operation (in == out).
+    indicator_check!(unsafe {
```
justsmth, line 155: NP: As a sanity check, can we have a test that verifies this results in `FipsServiceStatus::NonApproved`?  (You should be able to find a few relevant examples of this.)

aws-lc-rs/src/cipher/key.rs
```diff
@@ -113,6 +144,107 @@ impl SymmetricCipherKey {
...
+            return Err(KeyRejected::unspecified());
+        }
+        // `as_chunks` is only stable since Rust 1.88.0, so use explicit slicing
+        // instead to stay within the crate's MSRV.
+        let first_key: &[u8; 8] = key_bytes[0..8]
+            .try_into()
+            .map_err(|_| KeyRejected::unspecified())?;
+        let second_key: &[u8; 8] = key_bytes[8..16]
+            .try_into()
+            .map_err(|_| KeyRejected::unspecified())?;
```
justsmth: NP: instead of `map_err`, these could be `expect` as we've already verified that the length is valid.

aws-lc-rs/src/cipher/des.rs
qzh0223: should I wrote those tests in aws-lc-rs/src/cipher/tests/fips.rs? But they tests will be run with fips feature only?

aws-lc-rs/src/cipher/key.rs
qzh0223: I actually prefer to use map_err, it is not harmful to use it, as you said we already check the length, so doesn't matter map_err or expect it should not reach here.

aws-lc-rs/src/cipher/key.rs
```diff
@@ -113,6 +144,107 @@ impl SymmetricCipherKey {
...
+
+        // SP 800-67 §3.1 requires K1 != K2 for 2-Key TDEA; if they are equal
+        // the cipher degenerates to single-DES (56-bit effective security).
+        if first_key == second_key {
```
torben-hansen: Do in constant-time, just for the sake of it.

aws-lc-rs/src/cipher/key.rs
```diff
@@ -113,6 +144,107 @@ impl SymmetricCipherKey {
...
+            .try_into()
+            .map_err(|_| KeyRejected::unspecified())?;
+        let second_key: &[u8; 8] = key_bytes[8..16]
+            .try_into()
+            .map_err(|_| KeyRejected::unspecified())?;
+        let third_key: &[u8; 8] = key_bytes[16..24]
+            .try_into()
+            .map_err(|_| KeyRejected::unspecified())?;
+
+        // SP 800-67r2 Appendix A: for 3-Key TDEA, K1, K2 and K3 should be
```
torben-hansen: > P 800-67r2 Appendix A: for 3-Key [...] independently chosen

True requirement, but it's not specified in that appendix.

aws-lc-rs/src/cipher/key.rs
```diff
@@ -113,6 +144,107 @@ impl SymmetricCipherKey {
...
+        // independently chosen. We enforce that all three subkeys are
+        // distinct. Callers who want the `K1 ‖ K2 ‖ K1` (2TDEA) form must
+        // use `DES_EDE_FOR_LEGACY_USE_ONLY` with a 16-byte key instead of
+        // encoding 2TDEA as a 24-byte 3TDEA key.
+        if first_key == second_key || second_key == third_key || first_key == third_key {
```
torben-hansen: constant-time, for the sake of it?

aws-lc-rs/tests/cipher_test.rs
qzh0223: the source is : NIST SP 800-67 Rev 2 Appendix B which is the same one in https://csrc.nist.gov/projects/cryptographic-standards-and-guidelines/example-values. I used all of the KATs for all the supported modes.

I don't think write them in a file is needed, these tests follow the aes tests exactly.

I will add tests to cover semi-weak keys.

aws-lc-rs/src/cipher/key.rs
justsmth: The `expect` is better for ensuring the sanity of our code -- as it stands, the `map_err` is unreachable, and were it ever reached there's a serious bug for which we'd need more visibility than a simple `Err` return value would provide.

aws-lc-rs/src/cipher.rs
```diff
@@ -394,6 +477,53 @@ pub const AES_256: Algorithm = Algorithm {
...
+    id: AlgorithmId::DesEdeForLegacyUseOnly,
+    key_len: DES_EDE_KEY_LEN,
+    block_len: DES_BLOCK_LEN,
+};
+
+/// 3TDEA (DES-EDE3, 3-key Triple DES) cipher, for legacy interoperability only.
+///
+/// 3DES is a legacy algorithm and has been disallowed for encryption by
+/// [NIST SP 800-131A Rev. 2](https://csrc.nist.gov/publications/detail/sp/800-131a/rev-2/final)
+/// after 2023. It is retained here solely for interoperability with existing
```
torben-hansen: Delete
> with existing systems that cannot yet migrate

aws-lc-rs/README.md
```diff
@@ -124,6 +124,27 @@ It can be enabled in two ways:
...
+solely to support interoperability with existing systems that cannot yet migrate to
```
torben-hansen: Delete
> with existing systems that cannot yet migrate to AES

aws-lc-rs/src/cipher/des.rs
```diff
@@ -0,0 +1,187 @@
...
+        }
+        // Sanity check: `decrypt` validates that `in_out.len() % block_len == 0`
+        // for ECB mode before dispatching here.
+        debug_assert!(in_out_iter.into_remainder().is_empty());
+    }
+
+    Ok(in_out)
+}
+
+fn des_ede_cbc_encrypt(
```
torben-hansen: You abstract `2TDEA` and`3TDEA` to the same interface, which is okay. In a few comments you distinguished by using terms like `ede` and `ede3`. But in the implementation, `des.rs`, that gets quite confusing when the function name code path is e.g. `encrypt_cbc_mode` mapping to `des_ede_cbc_encrypt` which then call `DES_ede3_cbc_encrypt`. That's quite a good mix of no `ede`, with `ede, and `ede3`. Took me a bit to unpack that.

Not sure I have an action on it. But if one wants to abstract in this way, better to use naming not directly related.

aws-lc-rs/src/cipher/key.rs
justsmth: To clarify how this might be done:
```
if constant_time::verify_slices_are_equal(first_key, second_key).is_ok() {
    return Err(KeyRejected::inconsistent_components());
}
```

aws-lc-rs/src/cipher/key.rs
justsmth: This is a little trickier to get constant time:
```
        let mut any_equal: u8 = 0;
        for (a, b) in [
            (first_key, second_key),
            (second_key, third_key),
            (first_key, third_key),
        ] {
            any_equal |= u8::from(constant_time::verify_slices_are_equal(a, b).is_ok());
        }
        if any_equal != 0 {
            return Err(KeyRejected::inconsistent_components());
        }
```

I'm not totally sure the compiler couldn't find a way to short-circuit, but it's a step in the right direction.

aws-lc-rs/src/cipher/key.rs
```diff
@@ -113,6 +146,108 @@ impl SymmetricCipherKey {
...
+        // or to 2TDEA (K1==K3). Callers who want 2TDEA must use
+        // `DES_EDE_FOR_LEGACY_USE_ONLY` with a 16-byte key instead of
+        // encoding 2TDEA as a 24-byte 3TDEA key.
+        let mut any_equal: u8 = 0;
+        for (a, b) in [
+            (first_key, second_key),
+            (second_key, third_key),
+            (first_key, third_key),
+        ] {
+            any_equal |= u8::from(constant_time::verify_slices_are_equal(a, b).is_ok());
```
justsmth, line 231: We could make this slightly more resistant to compiler optimizations using `std::hint::blackbox`:
```rust
    any_equal |= u8::from(constant_time::verify_slices_are_equal(a, b).is_ok());
    any_equal = core::hint::black_box(any_equal);
```

aws-lc-rs/src/cipher/streaming.rs
```diff
@@ -132,6 +132,7 @@ impl StreamingEncryptingKey {
         // going through `SymmetricCipherKey` construction.  Validate
         // algorithm-specific key constraints (e.g. DES weak-key / K1!=K2
         // checks) that would otherwise be missed.
+        #[cfg(feature = "legacy-3des")]
         key.validate_key_material()?;
```
justsmth: Why is this build `cfg` needed? `validate_key_material` is available regardless.

justsmth: Ok, I see what changed. I'd rather provide a `validate_key_material` implementation for all key types.

justsmth (dismissed):

aws-lc-rs/src/cipher/tests/fips.rs
```diff
@@ -7,6 +7,9 @@ use crate::cipher::{
     DecryptingKey, EncryptingKey, PaddedBlockDecryptingKey, PaddedBlockEncryptingKey,
     StreamingDecryptingKey, StreamingEncryptingKey, UnboundCipherKey, AES_128, AES_192, AES_256,
 };
+#[cfg(feature = "legacy-3des")]
```
torben-hansen, line 10: If everything is feature gated, are tests actually running without changes to the test runner scripts?

justsmth: Good point! We should update several of our CI jobs to ensure this feature is enabled for their runs.

justsmth (approved):

torben-hansen (approved):

PR #1119, opened 2026-05-11
Reduce AEAD and RSA test iterations under DISABLE_SLOW_TESTS

justsmth:
### Description of changes:

The MSan CI (`aws-lc-rs msan`) and `cross v0.2.5 tests aarch64-unknown-linux-gnu` jobs are regularly exceeding 30 minutes. `AWS_LC_RS_DISABLE_SLOW_TESTS=1` was already wired into the build but wasn't reducing the iteration count enough on its own.

Three changes:

- **AEAD tests** — two reductions. (1) Unconditional: the `bad_tag`/`bad_nonce`/`bad_aad` checks now run once per test vector instead of once per `in_prefix_len`, since those failure paths don't depend on prefix length. (2) Under `disable_slow_tests`: shrink `in_prefix_lengths` from 36 entries to `{0, 1, 16, 17}`. This is by far the fattest axis of the inner loop, and the four offsets cover the alignment cases MSan actually cares about (in-place, byte-misaligned, AES block boundary, and one past).

- **RSA sign tests** — under `disable_slow_tests`, exercise one representative vector per digest via a `HashSet<String>`. DER parsing and the `Fail-Invalid-Key` early-return still run on every vector, so input-validation coverage is preserved; only the expensive signing path is sampled. Stable under test-file reordering.

- **`cross.yml`** — the cross v0.2.5 job referenced `Cross.toml.x86_64-unknown-linux-gnu`, a file that was deleted in #852. Repoint it at `Cross.toml.passthrough-only` (which is what the sibling `aws-lc-rs-cross-prebuilt-nasm` job already uses).

### Call-outs:

The `bad_func` hoist is unconditional, not gated on `disable_slow_tests`. It's a pure redundancy removal — the three failure paths don't read `in_prefix_len`, so running them 36 times per vector was giving us no additional coverage for the same inputs. Faster non-MSan CI jobs as a side benefit.

### Testing:

Covered by the existing CI suite. The MSan job is the primary target; the non-sampling CI jobs continue to exercise every test vector and the full 36-entry prefix-length set.

### DIFF
--- a/aws-lc-rs/tests/rsa_test.rs
+++ b/aws-lc-rs/tests/rsa_test.rs
@@ -17,6 +17,7 @@ use aws_lc_rs::signature::{
 };
 use aws_lc_rs::test::to_hex_upper;
 use aws_lc_rs::{digest, rand, signature, test, test_file};
+use std::collections::HashSet;
 
 #[test]
 fn rsa_traits() {
@@ -60,6 +61,11 @@ fn rsa_from_pkcs8_test() {
 #[test]
 fn test_signature_rsa_pkcs1_sign() {
     let rng = rand::SystemRandom::new();
+    // Under `disable_slow_tests`, one representative vector per digest is
+    // enough to exercise the signing path for MSan.  The `Fail-Invalid-Key`
+    // early-return above still runs on every vector, so input-validation
+    // coverage is preserved.
+    let mut signed_digests = HashSet::new();
     test::run(
         test_file!("data/rsa_pkcs1_sign_tests.txt"),
         |section, test_case| {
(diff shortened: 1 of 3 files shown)


aws-lc-rs/tests/rsa_test.rs
```diff
@@ -177,10 +189,13 @@ fn test_signature_rsa_pss_sign() {
             if key_pair.is_err() && result == "Fail-Invalid-Key" {
                 return Ok(());
             }
+            let msg = test_case.consume_bytes("Msg");
+            if cfg!(disable_slow_tests) && !signed_digests.insert(digest_name) {
+                return Ok(());
+            }
```
prasden, line 195: Non-blocking: same operations here and L104-L106. Worth a helper?

prasden (approved):

nhatnghiho (approved):

PR #1129, opened 2026-05-15
Support linking against a system-installed AWS-LC

justsmth:
### Description of changes:

Adds opt-in support for linking `aws-lc-sys` against an existing AWS-LC CMake installation instead of building from the bundled source. Activated by setting `AWS_LC_SYS_SYSTEM_DIR` to the install prefix. When unset, the build behaves exactly as before.

The new path validates the installation (confirms AWS-LC headers, version ≥ bundled), locates libcrypto/libssl honoring `AWS_LC_SYS_STATIC`, and uses pre-generated bindings from either `AWS_LC_SYS_SYSTEM_BINDINGS` or the conventional `share/rust/aws_lc_bindings.rs` location populated by AWS-LC's CMake (`-DGENERATE_RUST_BINDINGS=ON`, see [aws-lc#2999](https://github.com/aws/aws-lc/pull/2999)).

FIPS builds (`aws-lc-fips-sys`) are not yet supported on this path and will fail with a clear error directing users to build from source.

### Call-outs:

- Library resolution handles MSVC import-library vs static-archive disambiguation via sibling DLL detection (`lib/foo.lib` + `bin/foo.dll` → dynamic; `lib/foo.lib` alone → static). This is the only reliable heuristic since both forms share the `.lib` extension.
- The `-awslc` SONAME suffix used by `ENABLE_DIST_PKG` installs is supported as a fallback candidate name (`crypto-awslc`, `ssl-awslc`).
- `AWS_LC_SYS_STATIC` when set explicitly is treated as a hard requirement — no silent fallback to the other linkage form.

### Testing:

- Unit tests cover version parsing, header validation, library resolution (Unix + MSVC), and bindings lookup — run via `cargo test -p builder-test`.
- New CI workflow (`.github/workflows/system-lib-tests.yml`) builds AWS-LC from the submodule into static, shared, and symbol-prefixed install prefixes on Linux, macOS, and Windows; tests `aws-lc-sys` and `aws-lc-rs` against those installs. The submodule is intentionally left uninitialized in test jobs so any silent fallback to building from source hard-fails.
- `scripts/tests/test_system_lib.sh` provides the same fixture for local development.


samuel40791765 (dismissed):

builder/main.rs
```diff
@@ -541,9 +564,24 @@ fn get_builder(prefix: &Option<String>, manifest_dir: &Path, out_dir: &Path) ->
     };
 
     if is_fips_build() {
+        assert!(
+            get_system_dir_path().is_none(),
+            "System-library linking is not yet supported for FIPS builds.\n\
```
WillChilds-Klein, line 593: why is this not supported for FIPS?

builder/main.rs
```diff
@@ -709,6 +755,11 @@ fn initialize() {
 }
 
 fn is_bindgen_required() -> bool {
+    // The system library path uses pre-generated bindings shipped with the
+    // install; bindgen is never needed.
+    if get_system_dir_path().is_some() {
```
WillChilds-Klein, line 786: nice side effect of this feature!

builder/main.rs
```diff
@@ -617,7 +655,10 @@ static mut SYS_NO_JITTER_ENTROPY: Option<bool> = None;
 static mut SYS_NO_U1_BINDINGS: Option<bool> = None;
 static mut SYS_INCLUDES: Option<Vec<PathBuf>> = None;
 static mut SYS_SANITIZER: Option<String> = None;
-
+static mut SYS_STATIC: Option<bool> = None;
+static mut SYS_SYSTEM_DIR: Option<PathBuf> = None;
+static mut SYS_SYSTEM_BINDINGS: Option<PathBuf> = None;
+static mut SYS_SYSTEM_SKIP_VERSION_CHECK: bool = false;
```
WillChilds-Klein, line 686: why default this to false?

or, more generally, why enforce that system bindings are newer than bundled bindings?

WillChilds-Klein (dismissed):

builder/main.rs
justsmth: The logic required to support FIPS is slightly more complicated. I plan to post a PR for that next.

builder/main.rs
justsmth: We assume that any AWS-LC behavior that aws-lc-rs depend upon will be present in the bundled version of AWS-LC, as well as to be present in all later AWS-LC versions. We only allow skipping this to avoid unnecessarily failing a build where the consumer is accepting the risk that an older AWS-LC version is acceptable.

WillChilds-Klein (approved):

samuel40791765 (approved):

PR #1133, opened 2026-05-26
ci: replace dtolnay/rust-toolchain with direct rustup invocations

justsmth:
### Description of changes:
All CI jobs started failing because the `dtolnay/rust-toolchain` GitHub Action can no longer be fetched (`An action could not be found at the URI ...`). Per the [GitHub Actions Rust documentation](https://docs.github.com/en/actions/tutorials/build-and-test-code/rust), the recommended approach is to drive `rustup` directly — it's already installed on GitHub-hosted runners.

Every `dtolnay/rust-toolchain@*` invocation across 12 workflow files is replaced with the equivalent `rustup toolchain install ... --profile minimal --no-self-update` + `rustup default` pair. The redundant `rustup override set ${{ steps.toolchain.outputs.name }}` follow-ups are dropped since `rustup default` covers the same need.

For the container-based jobs in `compilers.yml` (bare `ubuntu:18.04` / `ubuntu:20.04` / `arm64v8/ubuntu:*` images that don't have rustup pre-installed), we bootstrap rustup via `rustup-init` from `sh.rustup.rs` — replicating what the dtolnay action was doing internally.

### Call-outs:
- The matrix-driven conditional toolchain in `cross.yml`'s `aws-lc-rs-cross-test` job (stable+target vs. nightly+rust-src, keyed on `matrix.target[2]`) became an `if`/`else` shell block since rustup arguments aren't trivially expressible as a single matrix-templated command.
- `pregen-bindings.yml` had a latent issue: both jobs installed `RUST_SCRIPT_NIGHTLY_TOOLCHAIN` and then ran `rustup override set $RUST_NIGHTLY_TOOLCHAIN` — overriding to a variable that wasn't used for the install. It happened to work because both env vars resolve to `nightly`. The replacement uses `RUST_SCRIPT_NIGHTLY_TOOLCHAIN` consistently.
- The bindings-generator workflows previously SHA-pinned `dtolnay/rust-toolchain@e97e2d8...`. SHA pinning no longer applies since we're invoking the runner-installed `rustup` binary directly.

### Testing:
CI itself is the test. All 13 workflow YAML files parse cleanly via `ruby -ryaml`.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -31,13 +31,10 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: 'recursive'
-      - uses: dtolnay/rust-toolchain@master
-        id: toolchain
-        with:
-          toolchain: ${{ env.RUST_NIGHTLY_TOOLCHAIN }}
-          components: rustfmt
-      - name: Set Rust toolchain override
-        run: rustup override set ${{ steps.toolchain.outputs.name }}
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install ${{ env.RUST_NIGHTLY_TOOLCHAIN }} --profile minimal --component rustfmt --no-self-update
+          rustup default ${{ env.RUST_NIGHTLY_TOOLCHAIN }}
 
       - name: Run cargo fmt
         run: cargo fmt -- --check --verbose
@@ -58,13 +55,10 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: 'recursive'
-      - uses: dtolnay/rust-toolchain@master
-        id: toolchain
-        with:
-          toolchain: ${{ env.RUST_NIGHTLY_TOOLCHAIN }}
-          components: clippy
-      - name: Set Rust toolchain override
-        run: rustup override set ${{ steps.toolchain.outputs.name }}
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install ${{ env.RUST_NIGHTLY_TOOLCHAIN }} --profile minimal --component clippy --no-self-update
+          rustup default ${{ env.RUST_NIGHTLY_TOOLCHAIN }}
       - name: Run cargo clippy
         working-directory: ${{ matrix.crate_dir }}
         run: cargo clippy ${{ matrix.features }} --all-targets -- -W clippy::all -W clippy::pedantic -D warnings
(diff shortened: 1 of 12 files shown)


crlorentzen (approved):

.github/workflows/compilers.yml
```diff
@@ -222,10 +222,10 @@ jobs:
           git fetch origin ${{ github.sha }}
           git checkout --recurse-submodules -b ci-job ${{ github.sha }}
           gcc --version
-      - uses: dtolnay/rust-toolchain@master
-        id: toolchain
-        with:
-          toolchain: stable
+      - name: Install Rust toolchain
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable --profile minimal
```
crlorentzen, line 227: Nit: Could be worth considering what [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain/blob/master/action.yml#L76) does with `--retry 10 --retry-connrefused`, for resilience against transient network failures and reducing manual interventions. (I'm thinking 3–5 rather than 10 — curl's exponential backoff means 10 retries can stall for ~17 min before failing, whereas 3–5 caps the worst case at 7–31 seconds.)

justsmth: Yeah, that might be needed. For now, I'm fine with waiting to see how frequently such failures occur that would've been avoided by using a `--retry`.

.github/workflows/cross.yml
```diff
@@ -69,19 +69,21 @@ jobs:
...
-          components: ${{ (matrix.target[2] == '0' && '') || 'rust-src' }}
-      - name: Set Rust toolchain override
-        run: rustup override set ${{ steps.toolchain.outputs.name }}
+      - name: Install Rust toolchain for ${{ matrix.target[0] }}
+        run: |
+          if [[ '${{ matrix.target[2] }}' == '0' ]]; then
+            rustup toolchain install stable --profile minimal --target ${{ matrix.target[0] }} --no-self-update
+            rustup default stable
+          else
+            rustup toolchain install nightly --profile minimal --component rust-src --no-self-update
```
WillChilds-Klein, line 84: why rust-src?

.github/workflows/integration.yml
```diff
@@ -351,12 +345,10 @@ jobs:
...
       # versions compatible with our declared rust-version.
       - name: Install stable Rust
-        uses: dtolnay/rust-toolchain@stable
+        run: rustup toolchain install stable --profile minimal --no-self-update
 
       - name: Install MSRV Rust version
-        uses: dtolnay/rust-toolchain@stable
-        with:
-          toolchain: 1.71.0
+        run: rustup toolchain install 1.71.0 --profile minimal --no-self-update
```
WillChilds-Klein, line 351: is this version used across multiple jobs? if so, is there some way we can set MSRV centrally?

WillChilds-Klein (approved):

.github/workflows/analysis.yml
```diff
@@ -254,10 +241,10 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: 'recursive'
-      - uses: dtolnay/rust-toolchain@master
-        id: toolchain
-        with:
-          toolchain: nightly
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install nightly --profile minimal --no-self-update
```
nebeid, line 246: why some jobs `install nightly` and some `install ${{ env.RUST_SCRIPT_NIGHTLY_TOOLCHAIN }}`

justsmth: These should probably all be `${{ env.RUST_SCRIPT_NIGHTLY_TOOLCHAIN }}`.  The reason the environment variable was added is that sometimes the current "nightly" build is broken, and we need to select some previous "nightly" build in order to unblock our CI.

PR #1138, opened 2026-06-01
ci(compilers): retry-tolerant curl, latest Go, drop unused PowerShell

### LINKED ISSUE #1136: ci: compilers.yml uses fragile curl-based tool installs inside raw containers
---
  name: AWS-LC for Rust General Issue
  about: Template
  title: 'ci: compilers.yml uses fragile curl-based tool installs inside raw containers'
  labels: ''
  assignees: ''
  ---

  ### Problem:

  The `compilers.yml` workflow runs inside upstream `ubuntu:20.04` / `ubuntu:18.04` containers and manually installs Go (and PowerShell) via `curl` from external URLs on every run. No retries, no caching, no fallback. Any transient network
  issue or upstream outage causes CI failures unrelated to code changes.

  Affected lines in `compilers.yml`: 55, 140, 205, 266 (Go), plus PowerShell `.deb` download from GitHub Releases.

  This pattern is inconsistent with other workflows in the repo that use `actions/setup-go@v6`.

  ### Solution:

  1. **Use `actions/setup-go`** — should work with `container:` jobs since GHA mounts the tool cache automatically. Needs testing.
  2. **OR `--retry 3 --retry-delay 5`** to curl commands — minimal change, reduces transient failures.

  * **Does this change any public APIs?** No.
  * **Which algorithm(s) will this impact?** None — CI-only change.

  * Go installation in container jobs should not fail due to transient network issues

justsmth:
### Issues:
Resolves #1136

### Description of changes:
The four container-based jobs in `compilers.yml` fetch Go, PowerShell, and rustup over the network with no retries and no `-f`, and the rustup install uses `curl | sh` (which silently executes a truncated script on a dropped connection). Any blip on `go.dev`, `github.com/releases`, or `sh.rustup.rs` fails unrelated PRs.

This PR makes those downloads retry-tolerant, drops the PowerShell install (it isn't invoked anywhere on the Linux build/test path — every `pwsh` reference in the tree is in upstream aws-lc's Windows-only CI), and bumps Go to latest stable while we're touching the file.

### Call-outs:

- I skipped the `actions/setup-go` option from the issue. It requires Node 20 inside the container, which needs glibc ≥ 2.28; the `ubuntu:18.04` rows are on 2.27 and would break. The rest of the repo uses `setup-go` because it doesn't run inside ancient containers — the inconsistency is intentional.
- PowerShell removal is the part most worth a careful look. PS 7.2 is EOL and PS 7.4+ drops 18.04 support, so "drop it" is also the only forward-compatible option if it turned out we did need it — but as best I can tell, we don't.
- `${CURL_RETRY_FLAGS}` is intentionally unquoted at call sites (relies on shell word-splitting). Don't add `set -u` or quote the expansion.

### Testing:
CI on this PR. Behavioral surface of the jobs is unchanged.

### DIFF
--- a/.github/workflows/compilers.yml
+++ b/.github/workflows/compilers.yml
@@ -18,6 +18,19 @@ env:
   # Otherwise, we test against the latest version.
   RUST_NIGHTLY_TOOLCHAIN: nightly
   DEBIAN_FRONTEND: noninteractive
+  # Pinned versions of tools we install via curl in container jobs.
+  # Hoisted here so updates are a single edit instead of four.
+  GO_VERSION: 1.26.3
+  # Standard curl flags for fetching tool tarballs/scripts in container jobs.
+  # -f: fail (non-zero exit) on HTTP errors instead of writing an HTML error
+  #     page to disk and confusingly failing the next tar/sh step.
+  # --retry / --retry-delay: ride out transient blips on go.dev, github.com
+  #     releases, and sh.rustup.rs. curl's default --retry behavior already
+  #     covers timeouts and HTTP 408/429/5xx, which is the failure mode we
+  #     actually see. (Note: --retry-all-errors would be nice-to-have but
+  #     was added in curl 7.71.0 and is unavailable on the ubuntu:18.04 and
+  #     ubuntu:20.04 containers used here.)
+  CURL_RETRY_FLAGS: "-fL --retry 5 --retry-delay 5"
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


dougch (approved):
compilers workflow green...

nhatnghiho (approved):

.github/workflows/compilers.yml
```diff
@@ -18,6 +18,19 @@ env:
...
+  # Standard curl flags for fetching tool tarballs/scripts in container jobs.
+  # -f: fail (non-zero exit) on HTTP errors instead of writing an HTML error
+  #     page to disk and confusingly failing the next tar/sh step.
+  # --retry / --retry-delay: ride out transient blips on go.dev, github.com
+  #     releases, and sh.rustup.rs. curl's default --retry behavior already
+  #     covers timeouts and HTTP 408/429/5xx, which is the failure mode we
+  #     actually see. (Note: --retry-all-errors would be nice-to-have but
+  #     was added in curl 7.71.0 and is unavailable on the ubuntu:18.04 and
+  #     ubuntu:20.04 containers used here.)
+  CURL_RETRY_FLAGS: "-fL --retry 5 --retry-delay 5"
```
crlorentzen, line 33: nit (personal preference): Let's uplift all calls to use the same features including protocol controls, ensuring common settings are easily maintained:
```
CURL_RETRY_FLAGS: "--retry 5 --retry-delay 5"
CURL_OPTS: "-fsSL --proto '=https' --tlsv1.2"
```
Then call sites become:
```
curl ${CURL_OPTS} ${CURL_RETRY_FLAGS} -O "https://go.dev/dl/..."
curl ${CURL_OPTS} ${CURL_RETRY_FLAGS} https://sh.rustup.rs -o /tmp/rustup-init.sh
```
Two changes from current:
1. Split retry flags from general options so each can be updated independently.
2. Uplift --proto '=https' --tlsv1.2 and -sS from the rustup-only calls — all targets are HTTPS, and silent mode makes sense for all CI fetches (it's unlikely someone is watching progress bars).

PR #1139, opened 2026-06-01
fix: Apply -ffile-prefix-map to jitterentropy builder

### LINKED ISSUE #1137: jitterentropy embeds absolute paths to files
### Problem:

The jitterentropy build is embedding the absolute paths to the source files in the generated `.o`. This means the build output is dependent on the system it was produced on. The following patch helps work around the issue:

```diff
--- a/builder/cc_builder.rs
+++ b/builder/cc_builder.rs
@@ -474,6 +474,26 @@
                 .flag("-fwrapv")
                 .flag("-Wconversion");
         }
+
+        // The jitter-entropy builder does not inherit the flags set in
+        // collect_universal_build_options on the main cc_build. Without
+        // -ffile-prefix-map, jitterentropy objects bake the absolute manifest
+        // & break build reproducibility under Bazel (path includes the output base).
+        if !is_like_msvc {
+            let path = self.manifest_dir.display().to_string();
+            let file_flag = format!("-ffile-prefix-map={path}=");
+            if je_builder.is_flag_supported(&file_flag).unwrap_or(false) {
+                je_builder.flag(&file_flag);
+            } else {
+                je_builder.flag_if_supported(format!("-fdebug-prefix-map={path}="));
+            }
+            // -ffile-prefix-map covers __FILE__ and GCC's UBSan source locations
+            // but clang stores UBSan source locations in separate type-descriptor
+            // records that bypass file-prefix-map. At -O0 those records are not
+            // dead-code-eliminated and bake absolute paths into .rodata. Strip
+            // all but the basename. Silently ignored if UBSan is not enabled.
+            je_builder.flag_if_supported("-fsanitize-undefined-strip-path-components=-1");
+        }
         je_builder
     }
```

justsmth:
### Issues:
Resolves #1137

### Description of changes:
The jitter-entropy builder constructs a fresh `cc::Build` and bypasses `collect_universal_build_options`, so `-ffile-prefix-map` never reaches jitter object files. Stock release builds are unaffected (jitter sources don't use `__FILE__`), but consumers that enable debug info on C compilation — Bazel, Nix, distro packagers — see absolute paths leak into the archive via DWARF.

Commit 1 fixes the builder. Commit 2 adds a CI step that builds with debug info enabled, since the existing release path-check ran without DWARF and couldn't catch this class of regression.

### Call-outs:
Review the commits in order — the CI commit alone would fail without the fix.

Path-stripping is applied to the jitter builder unconditionally (vs. the main builder, which gates it on `OPT_LEVEL`), since jitter is always compiled at `-O0` regardless of profile.

The helper also includes `-fsanitize-undefined-strip-path-components=-1` for defensive coverage; the new CI check deliberately does not exercise UBSan (rationale in the commit message).

### Testing:
Local A/B on `aarch64-apple-darwin` with `CARGO_PROFILE_RELEASE_DEBUG=true`: archive goes from 10 leaked path strings (all six `jitterentropy-*.c` paths) to zero. The new CI step is the codified equivalent.

### DIFF
--- a/.github/workflows/compilers.yml
+++ b/.github/workflows/compilers.yml
@@ -118,6 +118,25 @@ jobs:
           else
             exit 0; # SUCCESS
           fi
+      # Verify the same path stripping in a release build that keeps DWARF.
+      # This catches sub-builders that miss the shared prefix-map flags and
+      # exercises the assembler-specific `-Wa,--debug-prefix-map` path.
+      - if: ${{ matrix.gcc_version == '8' && matrix.fips == '0' }}
+        working-directory: /tmp/aws-lc-rs/
+        name: Verify paths not found in release build with debug info
+        env:
+          CARGO_PROFILE_RELEASE_DEBUG: 'true'
+          CARGO_TARGET_DIR: /tmp/aws-lc-rs-target-debuginfo
+        run: |
+          cargo build -p aws-lc-sys --release
+          RELEASE_LIBCRYPTO=$(find ${CARGO_TARGET_DIR}/release -name "libaws_lc_*_crypto.a")
+          if strings ${RELEASE_LIBCRYPTO} | grep '/tmp/aws-lc-rs/'; then
+            echo "FAIL: build paths leaked into archive (DWARF source paths):"
+            strings ${RELEASE_LIBCRYPTO} | grep '/tmp/aws-lc-rs/' | sort -u | head -20
+            exit 1; # FAIL - we did not expect to find "/tmp/aws-lc-rs/" (i.e., a path)
+          else
+            exit 0; # SUCCESS
+          fi
 
   aws-lc-rs-2004-arm-gcc:
     if: github.repository_owner == 'aws'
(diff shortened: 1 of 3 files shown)


builder/cc_builder.rs
```diff
@@ -380,6 +381,23 @@ impl CcBuilder {
...
+        // Probe first because clang's integrated assembler rejects the flag,
+        // and skip paths with spaces because `-Wa,...` must stay a bare token.
+        let opt_level = cargo_env("OPT_LEVEL");
+        if (target_os() == "linux" || target_os().ends_with("bsd"))
+            && !compiler_is_msvc
+            && !matches!(opt_level.as_str(), "0" | "1" | "2")
+            && !self.manifest_dir.to_string_lossy().contains(' ')
+        {
+            let path_str = self.manifest_dir.display().to_string();
+            let asm_flag = format!("-Wa,--debug-prefix-map={path_str}=");
```
WillChilds-Klein, line 395: is the trailing `=` intentional?

builder/cc_builder.rs
```diff
@@ -380,6 +381,23 @@ impl CcBuilder {
...
         }
 
+        // `-ffile-prefix-map` does not reach GNU `as` for `.S` sources, so in
+        // release-style builds mirror it with `-Wa,--debug-prefix-map=...`.
+        // Probe first because clang's integrated assembler rejects the flag,
+        // and skip paths with spaces because `-Wa,...` must stay a bare token.
+        let opt_level = cargo_env("OPT_LEVEL");
+        if (target_os() == "linux" || target_os().ends_with("bsd"))
+            && !compiler_is_msvc
+            && !matches!(opt_level.as_str(), "0" | "1" | "2")
```
WillChilds-Klein, line 391: why don't we do this for O3? what if `OPT_LEVEL` isn't set?

.github/workflows/compilers.yml
```diff
@@ -118,6 +118,25 @@ jobs:
...
+      - if: ${{ matrix.gcc_version == '8' && matrix.fips == '0' }}
+        working-directory: /tmp/aws-lc-rs/
+        name: Verify paths not found in release build with debug info
+        env:
+          CARGO_PROFILE_RELEASE_DEBUG: 'true'
+          CARGO_TARGET_DIR: /tmp/aws-lc-rs-target-debuginfo
+        run: |
+          cargo build -p aws-lc-sys --release
+          RELEASE_LIBCRYPTO=$(find ${CARGO_TARGET_DIR}/release -name "libaws_lc_*_crypto.a")
+          if strings ${RELEASE_LIBCRYPTO} | grep '/tmp/aws-lc-rs/'; then
```
WillChilds-Klein: nit: can you reference `working-directory` here and on L135?

.github/workflows/compilers.yml
```diff
@@ -174,6 +193,51 @@ jobs:
       - name: Run cargo test (release)
         working-directory: /tmp/aws-lc-rs/
         run: cargo test -p aws-lc-rs --release --all-targets --no-default-features --features ${{ (matrix.fips == '0' && 'unstable,aws-lc-sys') || 'unstable,fips' }}
+      # The steps below verify that we're successfully using `-ffile-prefix-map`
+      # to remove build environment paths from the resulting library.
+      - if: ${{ matrix.gcc_version == '8' }}
```
WillChilds-Klein: for this and release build check, should we make this ≥ ? 

perhaps not worth it for RelWithDebugInfo as that costs another `cargo build`.

builder/cc_builder.rs
justsmth: Yeah, intentional. The flag syntax is `--debug-prefix-map=OLD=NEW`, so `{path_str}=` sets `OLD=<manifest_dir>` and `NEW=` (empty). 

Effectively, it strips the prefix entirely rather than rewriting it to another path.

builder/cc_builder.rs
justsmth: We do. The condition is the negation `!matches!(opt_level, "0" | "1" | "2")`, so it fires for `3`/`s`/`z` (O3 included).

If `OPT_LEVEL` is unset `cargo_env` panics on a missing var... but Cargo always sets `OPT_LEVEL` during build-script execution. See list of environment variables set by Cargo here: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts

WillChilds-Klein (dismissed):

WillChilds-Klein (approved):

samuel40791765 (approved):

PR #1147, opened 2026-06-12
fix(builder): select flag dialect by compiler driver mode, not target ABI

### LINKED ISSUE #1146: aws-lc-sys 0.41.0 fails to build for x86_64-pc-windows-msvc under cargo-xwin (clang-cl): jitterentropy "--param ssp-buffer-size=4"
### Problem:

Building `aws-lc-sys` (via `aws-lc-rs`, pulled by `rustls`) for the
`x86_64-pc-windows-msvc` target with **cargo-xwin** (which uses `clang-cl`) fails
while compiling the **jitterentropy** C sources.

`CcBuilder` selects the non-MSVC / GNU flag branch for the jitterentropy build and
passes `--param ssp-buffer-size=4`
(see `builder/cc_builder.rs`, the `else` branch of `if is_like_msvc { ... }`,
around the `je_builder` flags). Under `clang-cl`:

1. `--param` is an unknown clang-cl argument and is dropped (warning).
2. its value `ssp-buffer-size=4` is then treated as a **stray positional input
   file** ("`ssp-buffer-size=4`: 'linker' input unused").
3. clang-cl now sees two inputs (`ssp-buffer-size=4` + the real `.c`) with a single
   `-Fo<file>.o`, and errors:
   `cannot specify '-Fo...' when compiling multiple source files`.

Root cause looks like `is_like_msvc` being **false** for the cargo-xwin `clang-cl`
toolchain even though the target is `*-pc-windows-msvc`, so the GNU flag branch is
taken for a compiler that is actually in cl (MSVC) mode. The flag is emitted **only**
in the jitterentropy sub-build (`ssp-buffer-size=4` appears exactly once in
`cc_builder.rs`); the main AWS-LC build compiles fine.

**Workaround that works:** `AWS_LC_SYS_NO_JITTER_ENTROPY=1` — skips the offending
sub-build, and the rest builds and links cleanly for windows-msvc under cargo-xwin.

**Minimal reproduction of the flag handling** (no aws-lc needed; any clang in cl
driver mode, here Apple clang 17, same as LLVM clang-cl):
$ printf 'int dummy(void){return 0;}\n' > foo.c
$ clang --driver-mode=cl -nologo -Fofoo.o --param ssp-buffer-size=4 -c -- foo.c
clang: warning: unknown argument ignored in clang-cl: '--param'
clang: error: cannot specify '-Fofoo.o' when compiling multiple source files
clang: warning: ssp-buffer-size=4: 'linker' input unused

Full build reproduction:
rustup target add x86_64-pc-windows-msvc
cargo xwin build --release --target x86_64-pc-windows-msvc
fails in aws-lc-sys on jitterentropy-*.c

#### Relevant details
AWS-LC for Rust versions or commit: `aws-lc-rs 1.17.0`, `aws-lc-sys 0.41.0`
(also `cc 1.2.55`, `rustls 0.23` with the `aws_lc_rs` feature, `cargo-xwin` with
xwin 0.21.4). Resolved features: `aws-lc-sys, prebuilt-nasm`.

System information (cross-compile):
 * CPU architecture: host x86-64 (Linux), target **x86_64-pc-windows-msvc**
 * CPU name: CI runner (Debian bookworm-slim container)
 * OS: Debian 12 (bookworm-slim) build host, cross-compiling to Windows via cargo-xwin / clang-cl

Build log:
 * compiler: clang-cl (from cargo-xwin)
```text
$ cargo xwin build --release --target x86_64-pc-windows-msvc
   Compiling aws-lc-sys v0.41.0
  cargo:warning=clang: warning: unknown argument ignored in clang-cl: '--param' [-Wunknown-argument]
  cargo:warning=clang: error: cannot specify '-Fo.../aws-lc-sys-.../out/...-jitterentropy-gcd.o' when compiling multiple source files
  cargo:warning=clang: warning: ssp-buffer-size=4: 'linker' input unused [-Wunused-command-line-argument]
  exit status: 1
  error occurred in cc-rs: command did not execute successfully (status code exit status: 1): LC_ALL="C" "clang-cl" "-nologo" "-MD" "-Brepro" "-m64" ... "-W4" "-fwrapv" "--param" "ssp-buffer-size=4" "-fvisibility=hidden" ... "-Fo.../jitterentropy-gcd.o" "-c" "--" ".../jitterentropy/jitterentropy-library/src/jitterentropy-gcd.c"
```

Suggested fix: detect the cl-mode/clang-cl compiler (or gate on it) before emitting
GNU-only flags like --param ssp-buffer-size=4 for the jitterentropy build, so the
windows-msvc cross-compile via cargo-xwin works without AWS_LC_SYS_NO_JITTER_ENTROPY.

justsmth:
### Issues:
Resolves #1146

### Description of changes:
`aws-lc-sys` was deciding several Windows compiler flag choices from the wrong signal.

The bug in #1146 is about compiler driver mode, not just the target ABI: a `*-pc-windows-msvc` build can still use plain `clang` in GNU mode, while `clang-cl` can be misdetected by `cc-rs` and still requires cl-style flags. This change introduces `compiler_is_cl_like()` and routes dialect-sensitive decisions through that helper, while keeping `target_is_msvc()` only for true ABI/CRT/SDK decisions.

That fixes the jitterentropy `--param ssp-buffer-size=4` failure under cargo-xwin / `clang-cl`, and it also preserves the inverse case where plain `clang` targets the MSVC ABI in GNU mode.

The PR also adds CI coverage for both Windows clang driver modes on the MSVC ABI and documents the distinction in the Windows requirements guide.

### Call-outs:
- Review focus is the split between `compiler_is_cl_like()` and `target_is_msvc()`: dialect-sensitive flag selection now uses driver mode, while ABI-specific behavior still keys off the target.
- `prepare_jitter_entropy_builder()` now derives its own dialect choice instead of accepting a caller-supplied bool.
- The original cargo-xwin failure is environment-specific, so the regression is guarded with deterministic unit tests around flag selection and driver detection, plus explicit CI lanes for `CC=clang` and `CC=clang-cl`.

### Testing:
- `cargo test -p builder-test --lib`
- `cargo test -p aws-lc-rs --tests`

### DIFF
--- a/builder/cc_builder.rs
+++ b/builder/cc_builder.rs
@@ -453,10 +453,8 @@ impl CcBuilder {
         }
     }
 
-    fn prepare_jitter_entropy_builder(&self) -> cc::Build {
+    fn prepare_jitter_entropy_builder(&self, is_cl_like: bool) -> cc::Build {
         // See: https://github.com/aws/aws-lc/blob/2294510cd0ecb2d5946461e3dbb038363b7b94cb/third_party/jitterentropy/CMakeLists.txt#L19-L35
-        // Resolve the flag dialect here so the caller cannot pass the wrong mode.
-        let is_cl_like = compiler_is_cl_like(&cc::Build::new().get_compiler());
         let mut build_options: Vec<BuildOption> = Vec::new();
         self.add_includes(&mut build_options);
         self.add_defines(&mut build_options, is_cl_like);
@@ -618,7 +616,7 @@ impl CcBuilder {
         s2n_bignum_builder.define("S2N_BN_HIDE_SYMBOLS", "1");
 
         // CPU Jitter Entropy is compiled separately due to needing specific flags
-        let mut jitter_entropy_builder = self.prepare_jitter_entropy_builder();
+        let mut jitter_entropy_builder = self.prepare_jitter_entropy_builder(is_cl_like);
         jitter_entropy_builder.flag(format!(
             "{}{}",
             force_include_option,
--- a/builder/cc_builder.rs
+++ b/builder/cc_builder.rs
@@ -458,11 +458,13 @@ impl CcBuilder {
         }
     }
 
-    fn prepare_jitter_entropy_builder(&self, is_msvc_target: bool) -> cc::Build {
+    fn prepare_jitter_entropy_builder(&self) -> cc::Build {
         // See: https://github.com/aws/aws-lc/blob/2294510cd0ecb2d5946461e3dbb038363b7b94cb/third_party/jitterentropy/CMakeLists.txt#L19-L35
+        // Resolve the flag dialect here so the caller cannot pass the wrong mode.
+        let is_cl_like = compiler_is_cl_like(&cc::Build::new().get_compiler());
         let mut build_options: Vec<BuildOption> = Vec::new();
         self.add_includes(&mut build_options);
-        self.add_defines(&mut build_options, is_msvc_target);
+        self.add_defines(&mut build_options, is_cl_like);
 
         let mut je_builder = cc::Build::new();
         for option in build_options {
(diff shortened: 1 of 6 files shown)


builder/main.rs
```diff
@@ -498,6 +498,52 @@ fn target_env() -> String {
...
+/// This reflects the target ABI only and is the right signal for ABI/CRT/SDK
+/// decisions (MSVC CRT linkage, `_CRT_SECURE_NO_WARNINGS`, the `clang-cl`
+/// dependency checks). It is *not* a reliable signal for the compiler's flag
+/// *dialect*: a `*-pc-windows-msvc` target can be built with plain `clang` in
+/// GNU driver mode (e.g. `CC=clang`), which needs GNU-style flags. For dialect
+/// decisions use `compiler_is_cl_like` instead.
+#[allow(unused)]
+fn target_is_msvc() -> bool {
+    target_env() == "msvc"
+}
```
prasden: I see this called in multiple places, do we need `#[allow(unused)]`?

builder/cc_builder.rs
```diff
@@ -452,17 +447,19 @@ impl CcBuilder {
...
                 build_options.push(BuildOption::define("AWSLC_WINDOWS_7_COMPAT", ""));
             }
         }
     }
 
-    fn prepare_jitter_entropy_builder(&self, is_like_msvc: bool) -> cc::Build {
+    fn prepare_jitter_entropy_builder(&self) -> cc::Build {
         // See: https://github.com/aws/aws-lc/blob/2294510cd0ecb2d5946461e3dbb038363b7b94cb/third_party/jitterentropy/CMakeLists.txt#L19-L35
+        // Resolve the flag dialect here so the caller cannot pass the wrong mode.
+        let is_cl_like = compiler_is_cl_like(&cc::Build::new().get_compiler());
```
prasden: Why a new `cc::Build here`? We get the value of `is_cl_like` at L604 right before calling into this function, could we pass it in as a parameter instead?

builder/main.rs
justsmth: Good point. I'll remove it.

prasden (approved):

WillChilds-Klein (approved):

Issue #1082, opened 2026-03-31
BUG: aws-lc-rs fails to load 1024 bit private key with KeyRejected("Unspecified") error


liyongjing:
Problem:
We have a key we use for testing that is not parseable by aws-lc-rs, but is parseable by the rsa crate.

Solution:
aws-lc-rs should be able to parse valid RSA keys.

Does this change any public APIs? No
Which algorithm(s) will this impact? RSA parsing
Requirements / Acceptance Criteria:
What must a solution address in order to solve the problem? How do we know the solution is complete?

Will the Usage Guide or other documentation need to be updated? No
Testing: How will this change be tested? Call out new integration tests, functional tests, or particularly
Test against our example key:
```rust
pub fn test() {
    use base64::Engine;
[...]

PR #1151, opened 2026-06-15
Return TooSmall/TooLarge for out-of-range RSA private keys

satyakwok:
### Issues:
Addresses #1082

### Description of changes:
`RsaKeyPair::validate_private_key` rejected any RSA private key outside the supported `2048..=8192` bit range with the generic `KeyRejected("Unspecified")`, which doesn't tell the caller what's wrong. As noted in #1082, a too-small key (e.g. 1024-bit) should report that it's too small.

`KeyRejected::too_small()` and `too_large()` already exist, so this returns those reasons instead — `TooSmall` for keys below 2048 bits and `TooLarge` for keys above 8192 bits. No behavior change beyond the error reason; out-of-range keys are still rejected.

### Call-outs:
This only changes the error *reason*, not which keys are accepted. The 2048–8192 bit support range is unchanged.

### Testing:
Added two regression tests in `basic_rsa_test.rs` using the existing `tests/data/rsa_test_private_key_1024.p8` (too small) and `rsa_test_private_key_16384.p8` (too large) fixtures, asserting the rejection reason. Verified they fail against the previous `Unspecified` behavior and pass with the change.

### DIFF
--- a/aws-lc-rs/src/rsa/key.rs
+++ b/aws-lc-rs/src/rsa/key.rs
@@ -177,7 +177,8 @@ impl KeyPair {
         }
         match key.as_const().key_size_bits() {
             2048..=8192 => Ok(()),
-            _ => Err(KeyRejected::unspecified()),
+            bits if bits < 2048 => Err(KeyRejected::too_small()),
+            _ => Err(KeyRejected::too_large()),
         }
     }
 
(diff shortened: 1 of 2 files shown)


justsmth:
Thanks for the PR!  There was a slightly larger change that needed to be made, which I posted as the PR here: https://github.com/aws/aws-lc-rs/pull/1153

I will close this PR, since that other one covers this and few other places needing the same change.  Thanks again!

satyakwok:
No worries, that’s fine. Thanks as well.

PR #1152, opened 2026-06-15
fix(system-lib): gate system AWS-LC on a declared minimum version

justsmth:
### Description of changes:

The system-library version check previously required the installed AWS-LC to be `>=` the bundled submodule's version. That floor ratcheted up on every submodule bump, rejected older-but-compatible installs, and couldn't run in CI without the submodule checked out (it read the submodule headers).

This replaces it with a declared minimum that's independent of the submodule:
- `aws-lc-sys` gates on mainline `1.68.0` — the first AWS-LC release that emits Rust bindings (`-DGENERATE_RUST_BINDINGS=ON`), which the system-library path consumes, so it's the genuine floor for this path.
- `aws-lc-fips-sys` gates on the FIPS *module* version (`AWSLC_FIPS_VERSION_NUMBER`, falling back to the version-string major on legacy `<=3.x` branches). aws/aws-lc#3211 decoupled the FIPS version from the library version, so the library version string is no longer a reliable ordering across the mainline/LTS split.

### Call-outs:

- The declared minimums are hand-maintained constants, *not* derived from the submodule. CI is the source of truth: `test-minimum-version` builds the mainline floor (`v1.68.0`) and `test-fips-minimum-version` builds the oldest FIPS branch that both reports `MINIMUM_FIPS_VERSION` and emits Rust bindings (`fips-2024-09-27`, which resolves to FIPS version 3 via the legacy fallback); both run the suite with the check enabled. A unit test guards that the declared minimums never exceed what we bundle.
- A *present-but-unparseable* `AWSLC_FIPS_VERSION_NUMBER` is a hard error, not a silent fallback to the library-version major — post-decoupling that value isn't a trustworthy substitute, so masking it would hide a real problem. An *absent* macro still falls back (the legacy `<=3.x` path).
- `MINIMUM_FIPS_VERSION` is `3` today, with a TODO to bump to `4` when the crate moves to `fips-2025-09-12-lts` (which would also move the FIPS floor job's pin).
- The FIPS workflow fans out over a configurable set of refs (FIPS branches + a FIPS build of `main`), edited in one place in the `setup` job. Listed refs must support `-DGENERATE_RUST_BINDINGS=ON` and be plain branch/tag names (they're embedded in artifact names).
- The system-library test jobs `rm -rf` the submodule tree after checkout (rather than only skipping init), so a from-source fallback stays physically impossible even if a future `actions/checkout` inits submodules by default.

### Testing:

- `cargo test -p builder-test` — FIPS-version resolution (including the malformed-macro path), the minimum-vs-bundled guards, and existing system-lib tests. Passing locally.
- CI: existing system-library jobs now run with the version check enabled (skip removed); `test-minimum-version` and `test-fips-minimum-version` exercise the non-FIPS and FIPS floors end-to-end. Needs a CI run to confirm `v1.68.0`, `fips-2024-09-27`, and a FIPS build of `main` build cleanly with Rust bindings.

### DIFF
--- a/.github/workflows/system-lib-tests-fips.yml
+++ b/.github/workflows/system-lib-tests-fips.yml
@@ -254,10 +254,16 @@ jobs:
             cargo_args: "-p aws-lc-rs --no-default-features --features fips --lib"
             cargo_cmd: "test"
     steps:
-      # Do not initialize the aws-lc submodule; that makes fallback-to-source a
-      # hard failure instead of a silent regression. The version check still
-      # runs (it compares against a declared minimum, not submodule headers).
+      # Do not initialize the aws-lc submodule, then remove its working tree
+      # outright; that makes fallback-to-source a hard failure instead of a
+      # silent regression, and stays robust even if a future actions/checkout
+      # starts initializing submodules by default. The version check still runs
+      # (it compares against a declared minimum, not submodule headers).
       - uses: actions/checkout@v7
+
+      - name: Remove the AWS-LC submodule working tree
+        shell: bash
+        run: rm -rf aws-lc-fips-sys/aws-lc
 
       - uses: dtolnay/rust-toolchain@stable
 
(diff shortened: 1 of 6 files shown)


.github/workflows/system-lib-tests.yml
```diff
@@ -275,6 +274,61 @@ jobs:
...
+        run: |
+          cmake -S aws-lc-min-src -B build-min \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DBUILD_TESTING=OFF \
+            -DBUILD_TOOL=OFF \
+            -DDISABLE_GO=ON \
+            -DBUILD_SHARED_LIBS=OFF \
+            -DGENERATE_RUST_BINDINGS=ON \
+            -DCMAKE_INSTALL_PREFIX="${PWD}/install-min"
+          cmake --build build-min --target install -j
```
prasden, line 330: `@stable` is the minimal profile, but `-DGENERATE_RUST_BINDINGS=ON` needs it. do we need to add `rustfmt` to components like the other jobs?

  https://github.com/aws/aws-lc-rs/blob/130a0b20bc452df4288994b9c19508d9a253bbeb/.github/workflows/system-lib-tests.yml#L49 
  https://github.com/aws/aws-lc-rs/blob/130a0b20bc452df4288994b9c19508d9a253bbeb/.github/workflows/system-lib-tests.yml#L125 
  https://github.com/aws/aws-lc-rs/blob/130a0b20bc452df4288994b9c19508d9a253bbeb/.github/workflows/system-lib-tests-fips.yml#L73

.github/workflows/system-lib-tests-fips.yml
```diff
@@ -42,12 +58,13 @@ jobs:
       matrix:
         # Cover both Apple architectures.
         os: [ ubuntu-latest, macos-15-intel, macos-latest ]
+        aws_lc_ref: ${{ fromJSON(needs.setup.outputs.refs) }}
```
WillChilds-Klein, line 61: nit: why not just inline the list here instead of parsing JSON?

.github/workflows/system-lib-tests.yml
```diff
@@ -244,8 +244,8 @@ jobs:
       # Intentionally do NOT initialize the aws-lc/ submodule. This makes the
```
WillChilds-Klein: is there something stronger we can do here than simply not init'ing the submodule? perhaps run `rm -rf aws-lc/*` or something?

if some later version of `actions/checkout` automatically init's submodules by default, we may get unexpected/unwanted behavior. perhaps i'm being overly paranoid here.

builder/system_library.rs
```diff
@@ -216,13 +229,29 @@ impl SystemLib {
...
+fn read_and_validate_base_h(include_dir: &Path) -> Result<(PathBuf, String), String> {
     let base_h = include_dir.join("openssl").join("base.h");
     let content = std::fs::read_to_string(&base_h)
         .map_err(|e| format!("Failed to read {}: {}", base_h.display(), e))?;
 
-    // Verify this is AWS-LC (not OpenSSL or BoringSSL). Tokenize each line so
-    // that an unrelated comment that mentions the macro can't false-match.
     let has_awslc_marker = content.lines().any(|line| {
         let mut tokens = line.split_whitespace();
         tokens.next() == Some("#define") && tokens.next() == Some("OPENSSL_IS_AWSLC")
```
prasden: We do this parsing here and in `extract_fips_version_number.` Helper?

.github/workflows/system-lib-tests.yml
```diff
@@ -275,6 +274,61 @@ jobs:
...
           fi
 
+  # ---------------------------------------------------------------------------
+  # Proves the declared minimum: build AWS-LC at MINIMUM_AWS_LC_VERSION (see
+  # builder/system_library.rs) and run the full test suite against it with the
+  # version check enabled. A failure means the minimum is unsupported and must
+  # be raised together with this pin.
+  # ---------------------------------------------------------------------------
+  test-minimum-version:
+    if: github.repository_owner == 'aws'
```
WillChilds-Klein, line 291: you don't want this check to run on forks?

builder/system_library.rs
prasden: Vestigial?

aws-lc-sys/README.md
```diff
@@ -123,9 +123,9 @@ If neither is available the build fails with guidance on how to proceed.
 
 ### Version compatibility
 
-The version embedded in the installed headers must be greater than or equal to
-the AWS-LC version bundled with this crate. To bypass this check (not
-recommended), set `AWS_LC_SYS_SYSTEM_SKIP_VERSION_CHECK=1`.
+The installed AWS-LC version must be at least the minimum supported by this
+crate (currently `1.68.0`), independent of the bundled version. To bypass this
```
WillChilds-Klein: suggest referencing the constant. `1.68.0` seems like it will get stale quickly here

.github/workflows/system-lib-tests.yml
```diff
@@ -275,6 +274,61 @@ jobs:
             exit 1
           fi
 
+  # ---------------------------------------------------------------------------
+  # Proves the declared minimum: build AWS-LC at MINIMUM_AWS_LC_VERSION (see
+  # builder/system_library.rs) and run the full test suite against it with the
+  # version check enabled. A failure means the minimum is unsupported and must
+  # be raised together with this pin.
+  # ---------------------------------------------------------------------------
+  test-minimum-version:
```
prasden, line 290: Should we also add an equivalent test for the FIPS versions?

builder/system_library.rs
```diff
@@ -216,13 +229,29 @@ impl SystemLib {
...
+/// bundled submodule) whose support is proven by CI; bump it and the CI pin
+/// together. See `.github/workflows/system-lib-tests.yml`.
+const MINIMUM_AWS_LC_VERSION: &str = "1.68.0";
+
+/// Minimum AWS-LC FIPS *module* version supported via the system-library path,
+/// for `aws-lc-fips-sys`. The FIPS version (aws/aws-lc#3211) is decoupled from
+/// the library version and increases monotonically across FIPS branches, so it
+/// is a stable basis for comparison.
+///
+/// TODO: bump to `4` when switching to the `fips-2025-09-12-lts` branch.
```
WillChilds-Klein, line 245: we test this branch in CI today, right (L43 system-lib-tests-fips.yml)? is now the time to bump to 4?

builder/system_library.rs
```diff
@@ -294,32 +336,55 @@ fn version_at_least(installed: &str, required: &str) -> Result<bool, String> {
...
+        if tokens.next() != Some("AWSLC_FIPS_VERSION_NUMBER") {
+            continue;
+        }
+        let raw = tokens.next();
+        return raw
+            .and_then(|v| v.parse::<u32>().ok())
+            .map(Some)
+            .ok_or_else(|| {
+                format!(
+                    "Malformed AWSLC_FIPS_VERSION_NUMBER: expected an integer, found {}",
```
WillChilds-Klein:
```suggestion
                    "Malformed AWSLC_FIPS_VERSION_NUMBER: expected an unsigned integer, found {}",
```

aws-lc-sys/README.md
justsmth: Updating the text to just reference the constant.

builder/system_library.rs
justsmth: I updated the wording to align with current logic.

builder/system_library.rs
justsmth: Not yet -- I'd like this mechanism compatible with the FIPS 3.x module for this next release. After that, we'll change over to the FIPS 4.0 branch and bump this as well.

.github/workflows/system-lib-tests.yml
justsmth: We have a very small subset of tests that we expect forks to run -- this helps to minimize the cost that might be incurred for contributors that fork our repo. Of course, people are free to remove this CI job filter condition if they'd like.

.github/workflows/system-lib-tests-fips.yml
justsmth: GH Actions matrices can't read `env`/`vars`, so the `setup` + `fromJSON` pattern is the cleanest way to keep one edit point. 

This list feeds 7 other jobs here — without it we'd be hand-syncing 7 locations every time these branches change.

WillChilds-Klein (approved):

prasden (approved):

PR #1156, opened 2026-06-23
feat(system-lib): auto-detect system AWS-LC from OPENSSL_* and pkg-config

justsmth:
### Description of changes:
Adds automatic system AWS-LC detection for `aws-lc-sys` and `aws-lc-fips-sys` when the explicit `*_SYSTEM_DIR` variable is not set.

The build script can now discover and adopt a system-installed AWS-LC through:

- `OPENSSL_DIR`
- `OPENSSL_INCLUDE_DIR` + `OPENSSL_LIB_DIR`
- pkg-config on Unix: `openssl`, `aws-lc`, `libcrypto`, `libcrypto-awslc`

A detected install is only adopted if it is actually AWS-LC, satisfies the required version checks, provides usable pre-generated Rust bindings, and, for FIPS, passes FIPS verification. Otherwise detection is ignored and the build falls back to the bundled source unless `*_USE_SYSTEM=1` requires a system install.

This also documents the behavior, including binding lookup rules, static/dynamic library selection, FIPS verification requirements, and the `*_USE_SYSTEM` controls.

### Call-outs:
- Auto-detection is best-effort; explicit `*_SYSTEM_DIR` remains strict and fails on misconfiguration.
- `*_SYSTEM_SKIP_VERSION_CHECK=1` skips only the version floor. It still requires AWS-LC headers.
- Split include/lib detection infers bindings from the include prefix, including AWS-LC’s cohabiting `<prefix>/include/aws-lc` layout.
- pkg-config detection is Unix-only; Windows auto-detection is covered through `OPENSSL_DIR`.

### Testing:
- CI covers Unix, macOS, Windows, FIPS, pkg-config, split-prefix installs, dist-package/cohabiting AWS-LC names, and required-system failure behavior.


builder/system_library_tests.rs
samuel40791765, line 1: I'm fairly certain that we would fail if one tried to link a system OpenSSL using this mechanism, but I'm curious what error would be emitted and whether we should have tests to cover this case that people could often fall into.

That might require an additional set up beyond the scope of this PR, so not a blocker.

.github/workflows/system-lib-tests.yml
```diff
@@ -102,13 +102,29 @@ jobs:
...
+        run: |
+          cmake -S aws-lc-sys/aws-lc -B build-dist-pkg \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DBUILD_TESTING=OFF \
+            -DBUILD_TOOL=OFF \
+            -DDISABLE_GO=ON \
+            -DBUILD_SHARED_LIBS=OFF \
+            -DGENERATE_RUST_BINDINGS=ON \
+            -DENABLE_DIST_PKG=ON \
+            -DENABLE_DIST_PKG_OPENSSL_SHIM=OFF \
```
samuel40791765, line 116: I was under the impression this was OFF by default, or is this just to be explicit? 

Also curious as to what `build-dist-pkg` does, I couldn't quite find any documentation on it.

samuel40791765 (dismissed):

builder/system_library_tests.rs
justsmth: The detection logic checks for the `OPENSSL_IS_AWSLC` marker in `base.h` and rejects anything that lacks it. It fails slightly later if we find the AWS-LC headers, but then attempt to link to an OpenSSL libcrypto:
* FIPS builds would fail early during the `verify_fips_install` step;
* non-FIPS builds would fail at final linkage due to undefined symbols (since several functions used by aws-lc-rs are unique to AWS-LC).

.github/workflows/system-lib-tests.yml
justsmth: Right -- `ENABLE_DIST_PKG_OPENSSL_SHIM` is OFF by default when `ENABLE_DIST_PKG` is ON. We make this explicit here so intent is obvious and unaffected by upstream default changes.

`build-dist-pkg` is just the CMake build directory name (`-B build-dist-pkg`).

aws-lc-sys/README.md
```diff
@@ -73,6 +73,61 @@ Set `AWS_LC_SYS_SYSTEM_DIR` to the install prefix:
...
+A discovered install is **adopted only if it is genuinely AWS-LC** (carries the
+`OPENSSL_IS_AWSLC` marker), satisfies the [version requirement](#version-compatibility),
+and ships usable [bindings](#bindings-resolution). Anything else — a system
+OpenSSL, an AWS-LC older than the minimum, or one without bindings — is ignored
+and the build falls back to source. Auto-detection never fails the build on its
+own; it only ever *adds* the option of linking a suitable system install.
+
+Use `AWS_LC_SYS_USE_SYSTEM` to control this behavior:
+
+* **unset** — detect; link a usable system AWS-LC if found, else build from source.
```
geedo0, line 116: Can you speak to the preference for the system libraries? I can see the case for either route being preferred.

aws-lc-sys/README.md
```diff
@@ -73,6 +73,61 @@ Set `AWS_LC_SYS_SYSTEM_DIR` to the install prefix:
...
+
+1. `AWS_LC_SYS_SYSTEM_DIR` — explicit install prefix (above). Any problem with
+   an explicit prefix is a hard error, never a silent fallback.
+2. `OPENSSL_DIR` — install prefix.
+3. `OPENSSL_INCLUDE_DIR` + `OPENSSL_LIB_DIR` — independent header and library
+   directories (both must be set).
+4. pkg-config (Unix only) — probes `openssl`, `aws-lc`, `libcrypto`, then
+   `libcrypto-awslc`, used only when none of the variables above are set.
+   Cross-compilation follows pkg-config's own rule (set
+   `PKG_CONFIG_ALLOW_CROSS=1` to allow it).
```
geedo0, line 95: Any reason why we decide to support so many options here (especially the OpenSSL variables)? What is the most conventional/expected case? I worry about the complexity this all adds. Over just a controlled/predictable single path of defining our own variable and/or .pc files.

builder/system_detect.rs
```diff
@@ -0,0 +1,190 @@
...
+//!
+//! Unlike the explicit `<crate>_SYSTEM_DIR` path (which treats any problem as a
+//! hard error), detection is best-effort: a candidate that isn't a usable
+//! AWS-LC is ignored with a diagnostic and the build falls back to compiling
+//! AWS-LC from source. The `<crate>_USE_SYSTEM` control variable (handled by the
+//! caller in `get_builder`) can turn "nothing found" into a hard error, or
+//! disable detection entirely.
+//!
+//! Discovery mirrors the mechanism `openssl-sys` uses, so that an environment
+//! which provides AWS-LC to `openssl-sys` (notably an Amazon Brazil version
```
geedo0: What's an Amazon Brazil version set? 🙃

geedo0 (dismissed):
Just some comments/curiosities, nothing blocking.

builder/system_detect.rs
justsmth: 🤦

aws-lc-sys/README.md
justsmth: For environments that already have AWS-LC in the build closure, linking it avoids a redundant from-source build and a second embedded copy of libcrypto. It feels safe as a default because detection is conservative: we only adopt a real AWS-LC that meets the version floor and ships bindings, otherwise we fall back to source. Anyone who wants the from-source build can force it with `AWS_LC_SYS_USE_SYSTEM=0`.

geedo0 (dismissed):

aws-lc-sys/README.md
justsmth: We're mirroring `openssl-sys`'s discovery of a `libcrypto` in the build environment — which works really well at finding AWS-LC. With this change, we should be able to link AWS-LC in all the places `openssl-sys` currently can. On the complexity of supporting multiple ways to specify locations: the header/library lookup lives in a single function, so it stays contained.

(I went back to compare against `openssl-sys` and found a subtle difference, which will fixed in the next commit.)

samuel40791765 (approved):

geedo0 (approved):

Issue #745, opened 2025-03-21
`aws-lc-rs` causes substantial binary size increase


landonxjames:
### Problem:

The AWS Rust SDK recently updated our default version of `rustls` to one that uses `aws-lc-rs` as the crypto provider. This has caused a substantial (consistently ~4MB in my tests) increase in binary size. This has caused some pain points for customers: https://github.com/smithy-lang/smithy-rs/issues/4049. In our debugging using [cargo-bloat](https://crates.io/crates/cargo-bloat) it appears that much of this size comes from `aws-lc` functions as can be seen in the output below from a small test application.

```sh
$ cargo bloat --release -n 50
 File  .text     Size                  Crate Name
 1.0%   2.1% 173.1KiB             aws_sdk_s3 aws_sdk_s3::config::endpoint::internals::resolve_endpoint
 0.3%   0.6%  47.2KiB              [Unknown] edwards25519_scalarmulbase_alt_edwards25519_gtable
 0.3%   0.6%  47.2KiB              [Unknown] edwards25519_scalarmulbase_edwards25519_gtable
 0.3%   0.6%  47.2KiB              [Unknown] curve25519_x25519base_byte_alt_edwards25519_gtable
 0.3%   0.6%  47.2KiB              [Unknown] curve25519_x25519base_byte_edwards25519_gtable
 0.2%   0.4%  30.7KiB             aws_lc_sys _aws_lc_0_27_0_pqcrystals_kyber512_ref_indcpa_keypair_derand
 0.2%   0.4%  28.7KiB             aws_lc_sys _aws_lc_0_27_0_pqcrystals_kyber768_ref_indcpa_keypair_derand
 0.2%   0.3%  25.7KiB             aws_lc_sys _aws_lc_0_27_0_pqcrystals_kyber768_ref_indcpa_enc
 0.1%   0.3%  24.9KiB              [Unknown] p384_montjscalarmul_p384_montjadd
 0.1%   0.3%  23.1KiB             aws_lc_sys _aws_lc_0_27_0_pqcrystals_kyber512_ref_indcpa_enc
 0.1%   0.3%  21.8KiB             aws_config aws_config::loader::ConfigLoader::load::{{closure}}
 0.1%   0.3%  21.8KiB             aws_config aws_config::loader::ConfigLoader::load::{{closure}}
 0.1%   0.3%  21.8KiB             aws_config aws_config::loader::ConfigLoader::load::{{closure}}
[...]

PR #1159, opened 2026-06-25
Automatically optimize aws-lc for size when opt-level is "s" or "z"

justsmth:
### Issues:
Addresses #745, P353434599

### Description of changes:

Consumers targeting constrained environments (embedded, Lambda, size-sensitive containers) currently have no straightforward way to reduce the AWS-LC footprint. This PR has the builder opt into AWS-LC's size-optimized configuration automatically whenever Cargo is already building for size (`opt-level=s` or `z`), so the smaller footprint comes along with the profile the consumer has already chosen. It can also be forced on or off explicitly, independent of opt-level.

Under the hood this enables AWS-LC's `OPENSSL_SMALL` configuration, which trades large, fast implementations (most notably the AVX-512 assembly on x86_64) for smaller generic ones. No algorithms are removed and outputs are unchanged -- the cost is slower elliptic-curve performance. The optimization is wired through both the cc and cmake builders across Linux, macOS, and Windows.

### Call-outs:

- The size defines are applied differently depending on the compilation path (C/`.S`, raw NASM, and FIPS), which can look inconsistent in the diff but is intentional. For C/`.S` the AVX-512 gate is derived automatically from `OPENSSL_SMALL` (via `target.h`), whereas raw NASM never includes that header and has to be told explicitly; prebuilt NASM objects cannot be reconfigured at all and instead emit a warning.
- FIPS builds are not excluded but emit a warning, consistent with how we treat other user-controlled settings that could affect the module.
- Related AWS-LC PRs: 
  - aws/aws-lc#3319
  - aws/aws-lc#3320

### Testing:

Full test suite passes with size-optimized paths active (`opt-level=z`), builder unit tests cover the new option handling, and a non-required CI job asserts a meaningful size reduction on x86_64 Linux.

### DIFF
--- a/.github/workflows/binary-size.yml
+++ b/.github/workflows/binary-size.yml
@@ -1,0 +1,109 @@
+name: binary-size
+
+on:
+  push:
+    branches:
+      - '*'
+      - '!generate/aws-lc-*'
+  pull_request:
+    branches:
+      - '*'
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+env:
+  RUST_BACKTRACE: 1
+  # Minimum expected size reduction percentage when opt-level = "z" vs "3".
+  # Measured at ~67% on x86_64 Linux; threshold set conservatively.
+  EXPECTED_REDUCTION_PERCENT: 40
+
+permissions:
+  contents: read
+
+jobs:
+  binary-size-check:
+    if: github.repository_owner == 'aws'
+    name: Binary size reduction (x86_64 Linux)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          submodules: 'recursive'
+
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install stable --profile minimal --no-self-update
+          rustup default stable
+
+      - name: Build with opt-level=3 (performance, no size optimization)
+        working-directory: tests/binary-size
+        env:
+          CARGO_PROFILE_RELEASE_OPT_LEVEL: "3"
+        run: |
+          cargo build --release
+          cp target/release/binary-size binary-size-default
+
+      - name: Build with opt-level=z (size-optimized, triggers OPENSSL_SMALL)
+        working-directory: tests/binary-size
+        env:
+          CARGO_PROFILE_RELEASE_OPT_LEVEL: "z"
+        run: |
+          cargo clean
+          cargo build --release
+          cp target/release/binary-size binary-size-small
+
+      - name: Verify small binary runs correctly
+        working-directory: tests/binary-size
+        run: ./binary-size-small
+
+      - name: Compare sizes
+        working-directory: tests/binary-size
+        run: |
+          size_default=$(stat --format="%s" binary-size-default)
+          size_small=$(stat --format="%s" binary-size-small)
+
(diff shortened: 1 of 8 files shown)


builder/main.rs
```diff
@@ -869,6 +871,27 @@ fn is_no_asm() -> bool {
     unsafe { SYS_NO_ASM }
 }
 
+pub(crate) fn is_small() -> bool {
```
nebeid, line 933: **FIPS: require explicit opt-in instead of inferring from `opt-level`.**

  `is_small()` returns `true` on any `opt-level=s/z` build with no override — including FIPS. Since `aws-lc-fips-sys` compiles the module from source, this changes
  the validated module's object code. `opt-level` is a general profile setting consumers set to shrink *their own* binary, so a FIPS consumer can silently land off a
  validated configuration with no crypto-specific intent — and the only guard is a `cargo:warning=` line, which is easily missed (scrolled past, or suppressed under
  `--quiet`/many CI setups).

  For the non-FIPS crate the opt-level trigger is great — reversible, behavior-preserving, well-documented. For FIPS I'd make it deliberate: keep auto-detection off
  unless `AWS_LC_FIPS_SYS_SMALL=1` is explicitly set. This reuses the override knob already in this PR (`SYS_SMALL`), no new mechanism:

  ```rust
  None => !is_fips_build() && matches!(opt_level.as_str(), "z" | "s"),
  ```

  `AWS_LC_FIPS_SYS_SMALL=1` still enables it; `opt-level` alone no longer mutates the FIPS module.
  
  (Separately: the new `binary-size` and `opt-level=z` test jobs are non-FIPS only, so the size-optimized FIPS module is currently neither size-checked nor tested —
  worth a follow-up if FIPS+small is a supported config.)

justsmth: * We generally do not enforce a specific configuration for FIPS builds -- the consumer is responsible for building in a way that satisfies their specific compliance requirements.
* AWS-LC's own CMake build scripts do not prevent use of `CMAKE_BUILD_TYPE` or `OPENSSL_SMALL` when there's also `FIPS=1`. If there were to be enforcement, I think we would want to enforce it in AWS-LC first.

nebeid (dismissed):

nebeid (approved):

builder/main.rs
```diff
@@ -775,6 +775,7 @@ static mut SYS_NO_ASM: bool = false;
 static mut SYS_PREBUILT_NASM: Option<bool> = None;
 static mut SYS_CMAKE_BUILDER: Option<bool> = None;
 static mut SYS_NO_PREGENERATED_SRC: bool = false;
+static mut SYS_SMALL: Option<bool> = None;
```
sgmenda, line 778: nit: we might want to refactor this `static mut` pattern into more idiomatic `std::sync` at some point.

.github/workflows/binary-size.yml
```diff
@@ -0,0 +1,141 @@
...
+
+jobs:
+  binary-size-check:
+    if: github.repository_owner == 'aws'
+    name: Binary size reduction (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # Minimum expected size reduction for opt-level "z" vs "3", set conservatively
```
sgmenda, line 28: how conservative would you say?

sgmenda (approved):
lgtm

Issue #1166, opened 2026-07-01
Public API for FIPS status


ctz:
### Problem:

Down in rustls we have the following sort of API (simplified):

```
enum FipsStatus {
  Unvalidated,
  Pending,
  Certified { certificate_url: &'static str }
}
```

Over in https://github.com/rustls/rustls/pull/3114/changes/78288e5ff04dbb9d51fdc7d4e39aa45b9bc525d7 I am considering determining the applicable FIPS certificate based on the first digit of [`aws_lc_fips_sys::AWSLC_VERSION_NUMBER_STRING`](https://docs.rs/aws-lc-fips-sys/latest/aws_lc_fips_sys/constant.AWSLC_VERSION_NUMBER_STRING.html) but this isn't a stable ABI, and requires taking a direct dependency on a crate which is otherwise private to `aws-lc-rs`.

### Solution:

It would be great if `aws-lc-rs` itself had a public API for obtaining this information. This API doesn't necessarily have to work in const-context for `rustls`, but that may be beneficial for other users.

One possibility is an enhanced version of [`try_fips_mode()`](https://docs.rs/aws-lc-rs/1.17.1/aws_lc_rs/fn.try_fips_mode.html) which returns similar information to above (albeit in a different type). Another is exposing the version of the FIPS module, or certificate URL, somewhere in the `aws-lc-rs` public API.

* **Does this change any public APIs?** This would add an additional public API surface.
* **Which algorithm(s) will this impact?** n/a

### Requirements / Acceptance Criteria:

**What must a solution address in order to solve the problem? How do we know the solution is complete?**

A crate downstream of `aws-lc-rs` can determine the FIPS status of the `aws-lc` version it is using, at runtime.

That API is available to both `feature = "fips"` builds, and not.

The data is accurate over internal evolution of `aws-lc` versions.

justsmth:
Hello! Thanks for reaching out.

I agree we need better visibility into the "FIPS module" being used. We recently changed how we version AWS-LC, which adds some complexity here (see AWS-LC's [VERSIONING.md](https://github.com/aws/aws-lc/blob/main/VERSIONING.md)).

> I am considering determining the applicable FIPS certificate based on the first digit of [aws_lc_fips_sys::AWSLC_VERSION_NUMBER_STRING](https://docs.rs/aws-lc-fips-sys/latest/aws_lc_fips_sys/constant.AWSLC_VERSION_NUMBER_STRING.html) ...

[...]

PR #1167, opened 2026-07-02
Add public API to check AWS-LC and FIPS versions

prasden:
### Issues:

Addresses #1166 

Added two public version APIs:

- `awslc_version()` returns the linked AWS-LC library version number, such as `"5.1.0"` or `"3.4.0"`, resolved at runtime.
- `fips_version()` returns the AWS-LC FIPS release-branch version as `Some(version)`, or `None` when the sys crate reports `0`.

For `aws-lc-fips-sys`, the FIPS version is resolved at build time from the AWS-LC headers. For `aws-lc-sys`, it is always `0`, including FIPS builds, because the crate does not track a supported AWS-LC FIPS release branch.

`fips_version()` is independent of `try_fips_mode()`: a build may have FIPS mode enabled while returning `None` from `fips_version()`.

### Call-outs:

The pinned AWS-LC FIPS branch does not provide a usable `FIPS_version()` API. The module version is therefore generated from the headers at build time. This can move to the runtime API once `aws-lc-fips-sys` tracks a FIPS 5+ branch.

### Testing:

Added `test_awslc_version`, which verifies the version starts with a numeric major version. `test_fips_version` verifies non-FIPS builds return `None` and supported FIPS builds return a module version of at least `3`.


aws-lc-rs/src/lib.rs
```diff
@@ -329,6 +333,44 @@ pub fn try_fips_mode() -> Result<(), &'static str> {
     }
 }
 
+/// The version string of the linked AWS-LC library (e.g. `"AWS-LC 3.4.0"`).
```
justsmth: Doc comment is wrong -- actual outputs are one of:
- non-FIPS: `"AWS-LC 5.1.0"`
- FIPS: `"AWS-LC FIPS 3.4.0"

aws-lc-rs/src/lib.rs
```diff
@@ -329,6 +333,44 @@ pub fn try_fips_mode() -> Result<(), &'static str> {
...
+///
+/// This identifies the exact release you are running against. For FIPS builds it
+/// can disambiguate builds that share a FIPS module number, but it must not be
+/// used to infer FIPS certification status directly.
+///
+/// # Panics
+/// Panics if AWS-LC returns a version string that is not valid UTF-8.
+#[must_use]
+pub fn awslc_version() -> &'static str {
```
justsmth, line 341: I think this should return the raw version number -- e.g, "3.4.0" or "5.1.0" -- not the unfiltered string.
```
pub fn awslc_version() -> &'static str {
    init();
    let full = unsafe { CStr::from_ptr(OpenSSL_version(OPENSSL_VERSION)) }
        .to_str()
        .expect("AWS-LC version string is not valid UTF-8");
    full.rsplit(' ').next().unwrap_or(full)
}
```

aws-lc-rs/src/lib.rs
```diff
@@ -329,6 +333,44 @@ pub fn try_fips_mode() -> Result<(), &'static str> {
...
+    init();
+    #[cfg(feature = "fips")]
+    let ptr = unsafe { awslc_version_string() };
+    #[cfg(not(feature = "fips"))]
+    let ptr = unsafe { OpenSSL_version(OPENSSL_VERSION) };
```
justsmth: This `cfg` split is not needed.

aws-lc-rs/src/lib.rs
```diff
@@ -421,4 +463,21 @@ mod tests {
             crate::fips_cpu_jitter_entropy();
         }
     }
+
+    #[test]
+    fn test_awslc_version() {
+        assert!(!crate::awslc_version().is_empty());
```
justsmth: This test is too easy to pass -- can we have it check the major-version number? (It'll need to be updated, but it'll be infrequent -- only after we cut an LTS branch.)

aws-lc-rs/src/lib.rs
prasden: Updated

aws-lc-rs/src/lib.rs
prasden: Yes, updated

builder/main.rs
```diff
@@ -1372,6 +1372,10 @@ fn main() {
 /// (`CMake` and CC). This sets up include paths, exports library and
 /// configuration names for downstream crates, and registers rerun triggers.
 pub(crate) fn emit_source_build_metadata(manifest_dir: &Path) {
+    if is_fips_build() {
+        system_library::emit_fips_version(&get_aws_lc_include_path(manifest_dir)).unwrap();
+    }
```
justsmth: I think this should instead be conditioned on `is_fips_crate()` -- we plan to support a "FIPS" build from `aws-lc-sys`, but when used `aws_lc_rs::fips_version()` should always produce a 0 (since such a FIPS build does not correspond to a supported FIPS branch -- however, `try_fips_mode()` would return `Ok`).

aws-lc-rs/src/lib.rs
```diff
@@ -329,6 +329,38 @@ pub fn try_fips_mode() -> Result<(), &'static str> {
...
+/// For FIPS builds, this reflects the module number of the pinned AWS-LC FIPS branch.
+#[must_use]
+pub fn fips_version() -> Option<u32> {
+    init();
+
+    #[cfg(feature = "fips")]
+    let version = Some(aws_lc::fips_version());
+    #[cfg(not(feature = "fips"))]
+    let version = None;
+
```
justsmth: Instead of conditional compilation -- this should unconditionally call `aws_lc::fips_version()` and return `None` if the value is `0`.

aws-lc-rs/src/lib.rs
```diff
@@ -421,4 +453,27 @@ mod tests {
...
+    #[test]
+    fn test_fips_version() {
+        assert_eq!(crate::fips_version(), None);
+    }
+
+    #[cfg(feature = "fips")]
+    #[test]
+    fn test_fips_version() {
+        assert_eq!(crate::fips_version(), Some(aws_lc::fips_version()));
+    }
```
justsmth, line 481: This can't fail -- not a good "test".  What about something like `crate::fips_version().unwrap() >= 3`?

aws-lc-rs/src/lib.rs
```diff
@@ -329,6 +329,38 @@ pub fn try_fips_mode() -> Result<(), &'static str> {
...
+        .expect("AWS-LC version string is not valid UTF-8");
+    full.rsplit(' ').next().unwrap_or(full)
+}
+
+/// The FIPS module number this build corresponds to, or `None` for non-FIPS builds.
+///
+/// For FIPS builds, this reflects the module number of the pinned AWS-LC FIPS branch.
+#[must_use]
+pub fn fips_version() -> Option<u32> {
+    init();
```
justsmth: The `init()` is pointless (besides being a no-op) here since neither implementation of `aws_lc::fips_version()` actually invokes something in the C library.

justsmth (commented):
The PR description needs to be updated based on the latest changes.

builder/system_library.rs
```diff
@@ -355,6 +362,16 @@ impl SystemLib {
             self.layout.marker_dir.display()
         ));
 
+        if is_fips_crate() {
+            // Only aws-lc-fips-sys consumes the generated constant; a
+            // FIPS-flavored aws-lc-sys build reports 0. Resolve here only if
+            // `validate` skipped it (skip_version_check).
+            let fips_version = match *self.fips_version.borrow() {
+                Some(version) => version,
+                None => validate_and_resolve_fips_version(include_dir)?,
```
justsmth: Since this branch is only reachable when the consumer set `SYSTEM_SKIP_VERSION_CHECK`, we should honor the skip here too rather than parse the headers anyway -- just emit a warning and report `0`:
```
        None => {
            emit_warning(
                "AWS-LC version check skipped; aws_lc_rs::fips_version() will report None.",
            );
            0
        }
```

We should also note in the `fips_version()` doc comments (both crates) that skipped-version-check builds report `None`/`0`.

prasden: Done, updated with comments

justsmth (dismissed):

justsmth (approved):

aws-lc-sys/src/lib.rs
```diff
@@ -106,3 +106,9 @@ pub fn BIO_get_mem_data(b: *mut BIO, pp: *mut *mut c_char) -> c_long {
 pub fn init() {
     unsafe { CRYPTO_library_init() }
 }
+
+/// Always `0`: `aws-lc-sys` does not track an AWS-LC FIPS release branch.
+#[must_use]
+pub const fn fips_version() -> u32 {
+    0
+}
```
samuel40791765, line 114: Nit: I understand we can't do this FIPS now, but Is there a particular reason why we can't call `FIPS_version` here and avoid hard-coding the value in two places?
```
  pub fn fips_version() -> u32 {
      unsafe { FIPS_version() }
  }
```

justsmth: We really do want `aws-lc-sys` to always return 0 here. The semantics of `fips_version()` are "which supported AWS-LC FIPS release branch does this build correspond to" -- and `aws-lc-sys` always builds from mainline, which is never a FIPS release branch.

Note that in the default (non-FIPS) build, `FIPS_version()` returns 0 anyway, so the two approaches only differ in one case: we plan to support building `aws-lc-sys` as a FIPS build, where [`FIPS_version()` would return `AWSLC_FIPS_VERSION_NUMBER`](https://github.com/aws/aws-lc/blob/7ed868ec4e4e2acc7ec238e1550d3d9f78df06f2/crypto/fipsmodule/self_check/fips.c#L29-L35) (currently 5). That's exactly the value we don't want -- such a build doesn't correspond to a supported FIPS branch, and returning 5 would make it hard to distinguish from a genuine FIPS-branch build. Consumers of that future build can verify FIPS mode via `try_fips_mode()` instead.

Hardcoding also keeps this a `const fn`, matching the `aws-lc-fips-sys` signature (whose value is likewise a build-time constant, since the pinned FIPS branch doesn't provide a usable `FIPS_version()` API).

aws-lc-rs/src/lib.rs
```diff
@@ -421,4 +455,28 @@ mod tests {
...
+    #[test]
+    fn test_fips_version() {
+        assert_eq!(crate::fips_version(), None);
+    }
+
+    #[cfg(feature = "fips")]
+    #[test]
+    fn test_fips_version() {
+        // Module versions are monotonic across FIPS branches; the pinned branch is 3.
+        assert!(crate::fips_version().unwrap() >= 3);
```
WillChilds-Klein, line 480: FIPS version is never less than 4, right? introduced in this commit at 4

https://github.com/aws/aws-lc/commit/8e71c38410811d6872f2a32eaf30a178f275b9bc


```suggestion
        assert!(crate::fips_version().unwrap() >= 4);
```

aws-lc-rs/src/lib.rs
```diff
@@ -329,6 +329,40 @@ pub fn try_fips_mode() -> Result<(), &'static str> {
...
+/// The FIPS module version this build corresponds to, or `None` if the build
+/// does not correspond to an AWS-LC FIPS release branch.
+///
+/// Unlike [`awslc_version()`], this is a build-time constant resolved from
+/// the AWS-LC headers; the library linked at runtime is not consulted. It is
+/// also not equivalent to [`try_fips_mode()`] and does not by itself imply
+/// FIPS certification status. Returns `None` when the system-library version
+/// check is skipped.
+// TODO: Resolve at runtime via the `FIPS_version()` C API once
+// aws-lc-fips-sys tracks a FIPS 5+ branch.
```
WillChilds-Klein, line 359: ` FIPS_version` [will return 0](https://github.com/aws/aws-lc/blob/532ff233935aa07b45f8e1fd83f5dc0c9b95c049/crypto/fipsmodule/self_check/fips.c#L29-L34) in the non-FIPS case, so unless we promise external users `*-sys` API stability, we can just call `FIPS_version` directly here and delete the `fips_version()` impls altogether when the time comes.

builder/system_library.rs
```diff
@@ -456,6 +476,23 @@ fn validate_and_resolve_fips_version(include_dir: &Path) -> Result<u32, String>
...
+pub(crate) fn emit_fips_version(include_dir: &Path) -> Result<(), String> {
+    let (base_h, content) = read_and_validate_base_h(include_dir)?;
+    let fips_version = resolve_fips_version(&base_h, &content)?;
+    write_fips_version(fips_version)
+}
+
+/// Writes `OUT_DIR/fips_version.rs`: a bare `u32` literal that `include!`
+/// expands in const context, keeping `fips_version()` a `const fn`.
+pub(crate) fn write_fips_version(fips_version: u32) -> Result<(), String> {
+    let dest = out_dir().join("fips_version.rs");
```
WillChilds-Klein, line 491: nit: is a single `u32` well-formed rust deserving of a `.rs` extension?

aws-lc-rs/src/lib.rs
prasden: https://github.com/aws/aws-lc-rs/blob/f031b0645c924d61cc6e943d187060c36380923d/.github/workflows/system-lib-tests-fips.yml#L316-L326

CI tests against our minimum supported FIPS 3 branch so this needs to remain `>= 3`

aws-lc-rs/src/lib.rs
justsmth: `FIPS_version()` won't return 0 when performing a "FIPS build" of our mainline branch. See my response [here](https://github.com/aws/aws-lc-rs/pull/1167#discussion_r3603026151).

builder/system_library.rs
justsmth: Not really -- but the filename provides an indication of its purpose. We can always change this filename later if there's a need to.

aws-lc-rs/src/lib.rs
justsmth: We will be switching to the "FIPS 4 branch" next week. We should bump this then.

builder/system_library.rs
WillChilds-Klein: nitpick: saving off the filename as a constant would be preferable.

WillChilds-Klein (approved):

PR #1170, opened 2026-07-06
Fix memcmp compiler check on AIX

thieman:
### Issues:
No open issues related to this as far as I know

### Description of changes: 
Fixes an unconditional error during crate compilation on `powerpc64`. The `memcmp_check` build was missing a required flag to build against 64-bit on AIX. This PR slots that in for the known architecture tuple of `aix` and `powerpc64`.

Following this change, I'm able to build the crate successfully on AIX 7.3. 🎉 

### Call-outs:
None that I'm aware of

### Testing:
No tests contributed in this PR. It doesn't seem like the repo has any ability to test against AIX machines, but if it does then please direct me to where I could add a smoke test for this compilation step working.

### DIFF
--- a/builder/cc_builder.rs
+++ b/builder/cc_builder.rs
@@ -826,6 +826,10 @@ impl CcBuilder {
         // requires -fuse-ld=lld). This matches the CMake build which only passes
         // CMAKE_C_FLAGS_RELEASE (-O3) to check_run().
         let mut memcmp_compile_args: Vec<std::ffi::OsString> = vec!["-O3".into()];
+        // AIX GCC defaults to 32-bit output; -maix64 selects the 64-bit ABI for the probe executable.
+        if target_os() == "aix" && target_arch() == "powerpc64" {
+            memcmp_compile_args.push("-maix64".into());
+        }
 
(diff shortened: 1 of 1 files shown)


samuel40791765 (approved):

justsmth:
Thanks for the PR!  After further thought on this problem, we'd like to take a different approach to resolve the build failure. 

For context -- in #1134, we went in (what I believe to be) the wrong direction. We stopped leveraging the build flags computed by cc-rs completely. This has led to us needing to recreate build logic from `cc-rs` (such as the logic to provide [proper build flags for AIX](https://github.com/rust-lang/cc-rs/blob/fa031a077aca2b19a31de895eafb17e965f35c89/src/lib.rs#L2239-L2244).

So, in #1173, we are partly reverting that approach and again leveraging the flags produced by `cc-rs` -- but isolating it from CFLAGS provided by the environment.  We're also narrowing the use of the memcmp compiler check to only those compilers potentially affected by it.

This PR will be closed. We would welcome feedback and/or testing on #1173 to confirm it resolves the issue you found. Thanks again for this contribution. We appreciate your help!

Issue #1168, opened 2026-07-03
aws-lc-sys 0.42.0: memcmp_invalid_stripped_check fails under hardened build flags (compile ignores CFLAGS, link respects LDFLAGS)


jgrund:
## Problem

`aws-lc-sys` 0.42.0 fails to build on x86_64 Linux under RPM hardened build flags (Fedora/RHEL/Rocky `rpmbuild` defaults):

```
thread 'main' panicked at aws-lc-sys-0.42.0/builder/cc_builder.rs:855:9:
COMPILER: gcc
ARGS: ["-O3", "-Wl,-z,relro", "-Wl,--as-needed", "-Wl,-z,pack-relative-relocs", "-Wl,-z,now", "-specs=/usr/lib/rpm/redhat/redhat-hardened-ld", "-specs=/usr/lib/rpm/redhat/redhat-hardened-ld-errors", "-specs=/usr/lib/rpm/redhat/redhat-annobin-cc1", "-Wl,--build-id=sha1", "-specs=/usr/lib/rpm/redhat/redhat-package-notes", ".../memcmp_invalid_stripped_check.c", "-Wno-unused-parameter", "-o", ".../memcmp_invalid_stripped_check"]
ERROR: /usr/bin/ld: /tmp/cchLSu7f.o: relocation R_X86_64_32 against `.rodata' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: failed to set dynamic section sizes: bad value
collect2: error: ld returned 1 exit status
OUTPUT: Failed to compile memcmp_invalid_stripped_check
```

0.41.0 builds fine in the same environment.

## Root cause

In 0.42.0 the `memcmp_invalid_stripped_check` probe was changed to compile with only `-O3`, intentionally ignoring `CFLAGS` (to address #1132), while still appending `LDFLAGS` (behavior from #958).

RPM hardened builds split PIE across the two variables:

- `CFLAGS` carries `-specs=/usr/lib/rpm/redhat/redhat-hardened-cc1`, which adds `-fPIE` at compile time — **now dropped**.
- `LDFLAGS` carries `-specs=/usr/lib/rpm/redhat/redhat-hardened-ld`, which forces `-pie` at link time — **still applied**.

The probe therefore compiles a non-PIE object and links it into a PIE executable, which fails on x86_64 (absolute `R_X86_64_32` relocations). aarch64 is unaffected since its codegen is PIC-compatible by default.

[...]

Issue #1132, opened 2026-05-23
aws-lc-sys fails to build by clang on Windows when installing cargo-audit


AngelicosPhosphoros:
### Problem:

I use clang as C, C++ compiler, lllvm-ar as archiver and rust-lld as linker when installing cargo binaries. That allows crosslanguage LTO.

However, in such setup, the `aws-lc-sys` crate doesn't compile on Windows and breaks installation of cargo-audit. It had successfully installed before inclusion of this crate (I don't remember version though and I had already deleted old one).

#### Relevant details
LLVM and rustc versions:

```
> rustc --version --verbose
rustc 1.95.0 (59807616e 2026-04-14)
binary: rustc
commit-hash: 59807616e1fa2540724bfbac14d7976d7e4a3860
commit-date: 2026-04-14
host: x86_64-pc-windows-msvc
release: 1.95.0
LLVM version: 22.1.2
> clang --version
clang version 22.1.6 (https://github.com/llvm/llvm-project fc4aad7b5db3fff421df9a9637605b9ca5667881)
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\LLVM\bin
```

OS version:

```
Get-ComputerInfo | Select-Object WindowsProductName, OsVersion, OsBuildNumber

WindowsProductName    OsVersion  OsBuildNumber
------------------    ---------  -------------
Windows 10 Enterprise 10.0.19041 19041
```

Command used to build using Clang + Rustc LTO;

```powershell
$env:CC='clang'
$env:CXX='clang++'
$env:CFLAGS='-flto=thin -march=native'
$env:CXXFLAGS='-flto=thin -march=native'
$env:AR='llvm-lib'
$env:RUSTFLAGS=' -Ctarget-cpu=native -Ccodegen-units=1 '
$env:CARGO_PROFILE_RELEASE_LTO='thin'
$env:CARGO_TARGET_X86_64_PC_WINDOWS_MSVC_LINKER='rust-lld'
cargo install --locked --target=x86_64-pc-windows-msvc cargo-audit
```

Build log:
<details><summary>Error message (huge)</summary>

```
    Updating crates.io index
  Installing cargo-audit v0.22.1
    Updating crates.io index
    Updating crates.io index
   Compiling proc-macro2 v1.0.106
   Compiling unicode-ident v1.0.22
   Compiling quote v1.0.44
   Compiling memchr v2.7.6
   Compiling windows-link v0.2.1
[...]

PR #1173, opened 2026-07-13
fix: restore cc default flags in memcmp probe; only fail build on confirmed bug

justsmth:
### Issues:
Addresses #1168, supersedes #1170

### Description of changes:
The memcmp probe (GCC bug 95189 check) has broken builds in unusual-but-legitimate environments three times (#1132, #1168, #1170). The root cause of the last two: #1134 dropped *all* cc-computed compiler flags to keep user CFLAGS out of the probe, which also discarded cc's target default flags (`-fPIC`, AIX's `-maix64`). This restores the computed defaults while still excluding user CFLAGS, by suppressing the CFLAGS env vars around `get_compiler()`.

The check is also now layered so only a confirmed bug can fail the build: it runs only for GCC (clang never had the bug), only for versions that may contain it (9.0-9.3, 10.0-10.1; the probe makes the final call since distros backport fixes), and a probe that fails to *build* only warns -- matching AWS-LC's CMake `check_run`. A probe that runs and demonstrates the miscompilation still fails the build.

### Call-outs:
* We remain intentionally stricter than CMake in one place: upstream only warns on a confirmed bug; we fail the build.
* No existing CI job exercises the detection path (all runners have fixed GCCs), so this adds a job that builds in `gcc:9.3`/`gcc:10.1` containers and asserts the build fails with `COMPILER BUG DETECTED`.

### Testing:
* New unit tests for the version gate and CFLAGS suppression (`cargo test -p builder-test`).
* Verified end-to-end in a `gcc:10.1` container: probe builds, detects the bug, build fails with the expected message.
* `cargo test -p aws-lc-sys`, clippy (pedantic), and rustfmt clean.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -953,7 +953,8 @@ jobs:
     # Reproduces the default Fedora/RHEL `rpmbuild` environment: %build exports
     # CFLAGS/CXXFLAGS/LDFLAGS from the redhat-rpm-config macros evaluated below.
     # These split PIE hardening across CFLAGS (-fPIE) and LDFLAGS (-pie), which
-    # broke the memcmp probe (it drops CFLAGS but honors LDFLAGS).
+    # historically broke the memcmp probe: it dropped all cc-computed compile
+    # flags (losing the default -fPIC) while honoring LDFLAGS.
     # See: https://github.com/aws/aws-lc-rs/issues/1168
     name: rpmbuild hardened environment
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -987,6 +988,47 @@ jobs:
...
+          - '10.1'
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          submodules: 'recursive'
+      # The gcc:9.3/10.1 images are Debian Buster based -- too old to host
+      # modern actions (node) in a job container, so run the build via docker.
+      - name: Build must fail with COMPILER BUG DETECTED
+        run: |
+          docker run --rm -v "$PWD:/repo" gcc:${{ matrix.gcc_version }} bash -c '
```
WillChilds-Klein, line 1014: who maintains these images? can we trust them to remain stable? i.e. not receive bug fix backports?

WillChilds-Klein (approved):

samuel40791765 (approved):

PR #1175, opened 2026-07-15
Add `rsa::KeyPair::from_components` for constructing RSA key pairs from raw components

### LINKED ISSUE #791: Construct an RSA Key pair from raw private components
### Problem:

It is not possible to construct an `rsa::KeyPair` from it's raw private components (d, p, q, etc).
In ring I'm able to use [`KeyPair::from_components`](https://docs.rs/ring/latest/ring/rsa/struct.KeyPair.html#method.from_components) method to create an RSA Key from the raw numbers.

I need this API to parse a JWK into a `rsa::KeyPair` to sign a JWS. 

### Solution:

Implement the same API `ring` has, using a `PrivateKeyComponents` struct and new function

justsmth:
### Issues:
Resolves #791

### Description of changes:
Adds `rsa::KeyPairComponents` and `rsa::KeyPair::from_components`, matching ring 0.17's API for constructing RSA key pairs from raw components. This allows callers to construct a signing key directly from formats such as JWK without first encoding the components as DER.

This also removes unnecessary generic bounds from `PublicKeyComponents` and applies the existing canonical encoding requirements for `n` and `e` consistently across RSA construction paths.

### Call-outs:
- Unlike ring, this implementation validates `d` and the CRT parameters at construction using `RSA_check_key`. Keys with inconsistent or placeholder values that ring accepts may therefore be rejected.
- Leading zeros are rejected for `n` and `e`, consistent with their existing API contract and ring's behavior. Leading zeros remain permitted for private components.
- `p` and `q` are not primality tested, matching ring.
- The use of `EVP_PKEY_set1_RSA` and the ownership handling around `RSA_set0_*` are intentional to avoid leaks or double frees.

### Testing:
Adds coverage for supported key-size boundaries, public exponent and encoding rules, corrupted components, DER parity, signing and verification, trait bounds, and `Debug` redaction.

Verified with `cargo test` under the default and `fips` features, `make clippy`, `make format`, and rustdoc with `-D warnings`.


aws-lc-rs/src/rsa/key.rs
```diff
@@ -97,6 +97,45 @@ impl Sealed for KeyPair {}
...
+    pub q: Private,
+
+    /// `p`'s CRT exponent: `d mod (p - 1)`.
+    pub dP: Private,
+
+    /// `q`'s CRT exponent: `d mod (q - 1)`.
+    pub dQ: Private,
+
+    /// The CRT coefficient: `q**-1 mod p`.
+    pub qInv: Private,
```
skmcgrail, line 123: I believe these components are optional? Could we reflect that in the API to allow creation from d, p, and q alone? Or are we trying to be strictly compatible with a ring API here?

skmcgrail (approved):

prasden (approved):

PR #1179, opened 2026-07-17
Publish script fixes: cargo clean failure, and verify aws-lc-rs against minimum published sys crates

justsmth:
### Description of changes:
Two fixes to the publishing scripts:

1. **Remove unnecessary `cargo clean` calls from prepublish checks.** Newer cargo (1.97+) refuses to clean a directory lacking a `CACHEDIR.TAG` file, which the first `cargo clean` tripped by running against a fresh mktemp dir. Both calls were unnecessary anyway.

2. **Verify aws-lc-rs builds against the minimum published sys crate versions before publishing.** This closes the gap behind the v1.17.2 yank: local checks resolve the sys crates via `path` dependencies, and `cargo package` verification only builds default features -- so the `fips` build was never tested against a registry-resolved aws-lc-fips-sys. The new check extracts the packaged `.crate` (path deps stripped), pins direct deps to their minimum allowed versions via `cargo +nightly update -Zdirect-minimal-versions`, and builds/tests both the default and `fips` configurations.

### Call-outs:
* `-Zdirect-minimal-versions` is nightly-only, but these scripts already assume nightly for `clippy --fix`. It pins only direct deps; full `-Zminimal-versions` is too fragile due to understated bounds in transitive crates.
* This adds a full AWS-LC FIPS build (~minutes) to the aws-lc-rs publish flow.

### Testing:
* Manually ran the new check against v1.17.3: sys crates resolve from the registry (0.43.0 / 0.13.16) and both configurations build.
* Ran it against the yanked v1.17.2: aws-lc-fips-sys resolves to v0.13.1 and the fips build fails -- it would have blocked the bad release.

### DIFF
--- a/scripts/publish/_publish_tools.sh
+++ b/scripts/publish/_publish_tools.sh
@@ -70,6 +70,64 @@ function sanity_check_sys_crate {
   echo Sanity check: SUCCESS
 }
 
+# Verifies that the packaged crate builds and tests against the dependency
+# versions currently published on crates.io. The manifest in the packaged
+# .crate has all `path` entries stripped, so dependency resolution comes from
+# the registry rather than the local workspace. Direct dependencies are pinned
+# to the *minimum* versions allowed by Cargo.toml to verify that the declared
+# version requirements are actually sufficient.
+#
+# This catches the failure that occurred with aws-lc-rs v1.17.2, which
+# depended on symbols only available in a not-yet-published aws-lc-fips-sys.
+# Local checks passed (path deps), but the published crate failed to build
+# with the `fips` feature.
+#
+# Usage: verify_crate_with_published_deps RELATIVE_CRATE_PATH CARGO_ARGS...
+#   Each CARGO_ARGS element is a (space-separated) set of arguments appended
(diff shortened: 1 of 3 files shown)


samuel40791765 (approved):

scripts/publish/_publish_tools.sh
```diff
@@ -70,6 +70,64 @@ function sanity_check_sys_crate {
...
 
+# Verifies that the packaged crate builds and tests against the dependency
+# versions currently published on crates.io. The manifest in the packaged
+# .crate has all `path` entries stripped, so dependency resolution comes from
+# the registry rather than the local workspace. Direct dependencies are pinned
+# to the *minimum* versions allowed by Cargo.toml to verify that the declared
+# version requirements are actually sufficient.
+#
+# This catches the failure that occurred with aws-lc-rs v1.17.2, which
+# depended on symbols only available in a not-yet-published aws-lc-fips-sys.
```
WillChilds-Klein, line 81: is ther an Issue associated with this?

WillChilds-Klein (approved):

PR #1183, opened 2026-07-19
Add out-of-place AEAD sealing

### LINKED ISSUE #1182: Expose out-of-place AEAD sealing
### Problem:

Every sealing entry point in `aws_lc_rs::aead` is in-place: the plaintext buffer is overwritten with the ciphertext. A caller whose plaintext is borrowed or shared therefore has to copy it into a scratch buffer purely to make it mutable — even though AWS-LC itself does not require that.

`EVP_AEAD_CTX_seal_scatter` takes distinct `in` and `out` pointers, and [the header is explicit that they need not alias](https://github.com/aws/aws-lc/blob/683ebde4bf3bcc016a9a710ad6b49c0c91b59161/include/openssl/aead.h#L360-L361):

> If `in` and `out` alias then `out` must be `== in`.

But [`UnboundKey::seal_in_place_separate_scatter`](https://github.com/aws/aws-lc-rs/blob/9232f4df24627662243a1a7cf71d885b2ff6ce72/aws-lc-rs/src/aead/unbound_key.rs#L199) [passes the same buffer to both](https://github.com/aws/aws-lc-rs/blob/9232f4df24627662243a1a7cf71d885b2ff6ce72/aws-lc-rs/src/aead/unbound_key.rs#L222-L236) — `in_out.as_mut_ptr()` as `out` and `in_out.as_ptr()` as `in`.

So out-of-place sealing is a gap in the Rust binding rather than a missing capability, and closing it needs no changes to AWS-LC.

For decryption, [`open_separate_gather`](https://github.com/aws/aws-lc-rs/blob/9232f4df24627662243a1a7cf71d885b2ff6ce72/aws-lc-rs/src/aead.rs#L812) *is* genuinely out-of-place, so encrypt and decrypt are currently inconsistent in their capabilities. [`seal_in_place_scatter`](https://github.com/aws/aws-lc-rs/blob/9232f4df24627662243a1a7cf71d885b2ff6ce72/aws-lc-rs/src/aead.rs#L924) is not a substitute — its `in_out` is still the aliased in-place payload buffer, and only the `extra_in` tail is scattered.

#### Use case:

`rustls` gathers each TLS record's payload into a freshly allocated buffer before sealing it, only because the current contract is in-place-only. This forces every byte to be copied.

`rustls` tracks the surface API for output buffer reuse as [#3068](https://github.com/rustls/rustls/issues/3068) and [#3127](https://github.com/rustls/rustls/pull/3127); the latter states the constraint directly: "both `ring` and `aws-lc-rs` only offer in-place sealing, so one plaintext copy is inherent regardless."

The equivalent request against *ring* ([#436](https://github.com/briansmith/ring/issues/436)) was opened by that project's own author in 2017 and closed `not_planned` in 2025.

### Solution:

Add three new API methods:

```rust
// Tier 1 -- primitive. Any caller-supplied buffer, including uninitialised.
// Returns the two outputs as initialised slices, so the caller needs no `unsafe`
// of its own to treat them as written.
pub fn seal_separate_out_of_place_uninit<'o, A: AsRef<[u8]>>(
    &self, nonce: Nonce, aad: Aad<A>,
    plaintext: &[u8],
    ciphertext: &'o mut [MaybeUninit<u8>],
    extra_in: &[u8],
    extra_out_and_tag: &'o mut [MaybeUninit<u8>],
) -> Result<(&'o mut [u8], &'o mut [u8]), Unspecified>;

// ... and the same for callers whose buffers are already initialised.
pub fn seal_separate_out_of_place<A: AsRef<[u8]>>(
    &self, nonce: Nonce, aad: Aad<A>,
    plaintext: &[u8],
    ciphertext: &mut [u8],
    extra_in: &[u8],
    extra_out_and_tag: &mut [u8],
) -> Result<(), Unspecified>;

// Tier 2 -- ergonomic. Appends ciphertext, extra ciphertext and tag to `out`,
// growing it into spare capacity without zeroing. The caller writes no `unsafe`,
// cannot mis-size the output, and can supply a recycled buffer.
pub fn seal_out_of_place_append<A: AsRef<[u8]>>(
    &self, nonce: Nonce, aad: Aad<A>,
    plaintext: &[u8],
    extra_in: &[u8],
    out: &mut Vec<u8>,
) -> Result<(), Unspecified>;
```

All three should be defined for both `LessSafeKey` and `TlsRecordSealingKey`. The `extra_in` / `extra_out_and_tag` pair mirrors the existing `seal_in_place_scatter`; TLS 1.3 uses it for the inner content-type byte.

The `MaybeUninit` tier exists because forming a `&mut [u8]` over uninitialised memory is undefined behaviour even when every byte is about to be written, so a caller wanting to skip the zero-fill needs an entry point that speaks `MaybeUninit`. Returning the initialised slices rather than `()` keeps the one unverifiable FFI claim — "the C function wrote every byte" — inside the crate that documents it, instead of pushing it out to every call site.

`seal_out_of_place_append` is what the rustls integration actually uses, and it is the reason the rustls-side change contains **no `unsafe` at all**: the crate owns the uninitialised region, and `out.len()` afterwards is a durable record of what was written, where a returned slice's evidence would expire with the borrow. It also computes the output size internally, so the caller cannot get it wrong.

The `LessSafeKey` methods carry the same FIPS note as the rest of that type.

* **Does this change any public APIs?** Yes, but purely additively: three new methods on each of `LessSafeKey` and `TlsRecordSealingKey`. No existing signature changes, and nothing is deprecated or removed, so it is not a breaking change.
* **Which algorithm(s) will this impact?** Any AEAD reachable through those two types. The mechanism is algorithm-agnostic — the same `EVP_AEAD_CTX_seal_scatter` call with `in` and `out` no longer aliased — so it applies uniformly rather than per-algorithm. Tested against AES-128-GCM, AES-256-GCM and CHACHA20_POLY1305.

#### Measurements:

I implemented this and wired rustls's TLS 1.3 AES-GCM encrypter onto it, with no change to `MessageEncrypter::encrypt`'s signature — the plaintext stays immutable and borrowed, so nothing above rustls has to change.

Measured on a dedicated bare-metal `c7i.metal-24xl` (Xeon Platinum 8488C, turbo disabled, `performance` governor, SMT sibling of the pinned core offlined, benchmark pinned to one core), using rustls's own bulk send benchmark with a single pinned key type. All three binaries ran on the same box; `n=11` runs each, median reported:

| build | bulk send | vs baseline |
|---|---|---|
| baseline: gather, then seal in place | 7528.7 MB/s | — |
| out-of-place seal, **zero-filled** output | 7559.4 MB/s | **+0.4%** |
| out-of-place seal, **appending** into spare capacity | 9377.5 MB/s | **+24.6%** |

Run-to-run spread was 0.18–0.49% stdev. A second, independently provisioned box measured the outer pair at +24.9%, so the result reproduces across instances.

**Skipping the zero-fill is required to see an improvement.** Sealing out-of-place into a zero-filled buffer buys essentially nothing: it swaps one full write pass for another; the gather `memcpy` disappears but a `memset` of the destination replaces it. The win comes from eliminating the extra pass over the record altogether, which needs out-of-place sealing *and* an uninitialised destination together. Neither alone moves the number.

A solution is complete when a caller can seal from a borrowed plaintext into a separate, uninitialised buffer — without copying the plaintext first, and without writing `unsafe` to do so.

* **RFC links:** none. This is a binding-level gap rather than a protocol change; the relevant contract is the `EVP_AEAD_CTX_seal_scatter` documentation linked above.
* **Related Issues:** rustls [#3068](https://github.com/rustls/rustls/issues/3068) and [#3127](https://github.com/rustls/rustls/pull/3127) are the motivating consumer; *ring* [#436](https://github.com/briansmith/ring/issues/436) is the equivalent request there, closed `not_planned`.
* **Will the Usage Guide or other documentation need to be updated?** Rustdoc on the new methods only. No Usage Guide changes — this adds an alternative to existing entry points rather than changing recommended usage.
* **Testing:** unit tests covering the new methods:
  - Byte-identical to `seal_in_place_separate_tag` for AES-128-GCM, AES-256-GCM and CHACHA20_POLY1305 at lengths 0, 1, 15, 16, 17, 1024, 4096, 16384, 16385
  - Round-trips through `open_in_place`
  - The plaintext is provably unmodified
  - The `MaybeUninit` and initialised variants produce identical output
  - Mismatched buffer lengths are refused
  - A caller sealing into an uninitialised buffer and using the result compiles under `#![forbid(unsafe_code)]`, so the "no `unsafe` needed" claim is enforced by the compiler rather than asserted

### Out of scope:

* `RandomizedNonceKey` and `SealingKey<N>` also seal in place and could take the same treatment. I've left them out because the TLS use case doesn't exercise them, but I can include them in a PR if you'd prefer the whole surface covered.
* A fully-allocating variant returning a fresh `Vec` would be two lines on top of `seal_out_of_place_append`, but it would mandate an allocation per call and foreclose buffer reuse, which is the direction rustls is moving. Happy to add one if you disagree.
* Out-of-place *opening* — `open_separate_gather` already provides it.

---

Happy to open a PR — [the branch is ready](https://github.com/iainmcgin/aws-lc-rs/tree/iain/out-of-place-seal). Filing this first per `CONTRIBUTING.md`.

iainmcgin:
### Issues:

Resolves #1182

### Description of changes:

Every sealing entry point in `aws_lc_rs::aead` is in-place, so a caller whose plaintext is borrowed or shared must copy it into a scratch buffer purely to make it mutable. `EVP_AEAD_CTX_seal_scatter` takes distinct `in` and `out` pointers and documents that they need only be equal *if* they alias, so the copy is a property of this binding rather than of the AEAD. #1182 has the full background.

Adds one method to `LessSafeKey` and one to `TlsRecordSealingKey`:

```rust
pub fn seal_separate_out_of_place<A: AsRef<[u8]>>(
    &self,                      // &mut self on TlsRecordSealingKey
    nonce: Nonce,
    aad: Aad<A>,
    plaintext: &[u8],           // borrowed immutably, left untouched
    ciphertext: &mut [u8],      // caller-provided, exactly plaintext.len()
    extra_in: &[u8],
    extra_out_and_tag: &mut [u8],
) -> Result<(), Unspecified>
```

This mirrors `open_separate_gather`, which already takes a caller-provided `&mut [u8]` for the opening direction, so the two directions now match. `extra_in` / `extra_out_and_tag` carry TLS 1.3's inner content-type byte and follow the existing `seal_in_place_scatter` signature. No existing signature changes; nothing is deprecated or removed.

**Narrowed from the original three methods**, per @justsmth's review. The earlier revision also had `seal_separate_out_of_place_uninit` (a `MaybeUninit` output) and `seal_out_of_place_append` (a `&mut Vec<u8>` grown into its spare capacity). Both existed to skip zero-filling a freshly allocated output buffer per record. rustls/rustls#3127 reworks `MessageEncrypter::encrypt` to write into a caller-provided `&mut [u8]` fed from recycled record buffers, which are already initialised — so there is no per-record fill left to skip, and the saving arrives through buffer reuse instead. With that consumer gone, those variants would have introduced two patterns with no precedent in this crate's public API for no caller. If rustls lands on a `&mut Vec<u8>` parameter instead, they are easy to revisit as a follow-up.

Removing the `MaybeUninit` tier also removed the only `unsafe` this change added. `seal_separate_out_of_place` is now a direct `EVP_AEAD_CTX_seal_scatter` call over two slices, with each length check next to the buffer it guards.

### Call-outs:

**On the performance numbers, which have changed meaning.** An earlier revision of this description reported +24.6% on a rustls bulk-send benchmark. That figure belonged to the `_append` variant and does **not** carry over to what is being merged here, so I want to be explicit rather than leave a stale headline up.

What I measured, on a dedicated bare-metal `c7i.metal-24xl` (turbo disabled, `performance` governor, SMT sibling of the pinned core offlined, pinned to one core), three binaries on one box, `n=11` each, medians:

| build | bulk send | vs baseline |
|---|---|---|
| baseline: gather, then seal in place | 7528.7 MB/s | — |
| out-of-place seal into a **freshly zero-filled** buffer | 7559.4 MB/s | +0.4% |
| out-of-place seal appending into spare capacity | 9377.5 MB/s | +24.6% |

The middle row is the closest to this PR, but it is not the same configuration: it zero-filled the destination on every record, so it traded a gather `memcpy` for a destination `memset` and unsurprisingly came out flat. A recycled buffer, which is what rustls/rustls#3127 moves to, is filled once at setup and never again, so that per-record `memset` does not exist. I would therefore expect the recycled-buffer case to land much closer to the third row than the second — but **I have not measured that configuration**, and I would rather say so than quote a number that describes different code. Happy to run it if it would help the review; it needs a rustls build wired onto the new method, which is the same harness the table above came from.

The correctness argument for the change is independent of all this: out-of-place sealing is a capability the C library has always had, the copy is an artifact of the binding, and the opening direction already exposes it.

**FIPS service indicator.** The out-of-place path calls `EVP_AEAD_CTX_seal_scatter` without wrapping it in `indicator_check!`, which matches the existing `seal_in_place_separate_scatter` but differs from every other seal in `unbound_key.rs`. Since `indicator_check!` compiles to nothing outside `fips` + `debug_assertions`, this is service-indicator observability rather than a release-mode behaviour change, but it does mean these entry points cannot be covered by `assert_fips_status_indicator!`. I have kept the scatter family internally consistent rather than pick a direction unilaterally — happy to wrap all of them (including the pre-existing one) or to leave it as is, whichever you prefer.

**Deliberately not included:** `RandomizedNonceKey` and `SealingKey<N>` also seal in place and could take the same treatment — left out because the TLS use case doesn't exercise them, happy to add.

### Testing:

Unit tests in `aws-lc-rs/tests/aead_out_of_place.rs` (`LessSafeKey`) and `aws-lc-rs/tests/aead_out_of_place_tls.rs` (`TlsRecordSealingKey`). Both are `#![forbid(unsafe_code)]`.

- Output is byte-identical to `seal_in_place_separate_tag` for AES-128-GCM, AES-256-GCM and CHACHA20_POLY1305 at nine lengths from 0 to 16385, and for `TlsRecordSealingKey` across both algorithms and both protocol versions.
- Round-trips through `open_in_place`, and for the TLS type through `TlsRecordOpeningKey`. That is what verifies the `extra_in` inner content-type byte is encrypted and authenticated rather than copied; the tests also assert it does not appear in the clear.
- Sealing into a recycled buffer overwrites the previous record, leaving no stale bytes in the region handed back.
- The plaintext is provably unmodified.
- Mismatched buffer lengths are refused, over-long as well as short.
- On a `TlsRecordSealingKey`, a nonce already used in place is refused out of place too, so the two entry points share one counter and a caller cannot replay a nonce by switching between them.

One behaviour worth recording, because it surprised me and the docs now state it: the error path is not uniform. A length mismatch is rejected before the AEAD runs and leaves the caller's output untouched, whereas a failure inside AWS-LC reaches `EVP_AEAD_CTX_seal_scatter`, which zeroes both output buffers on the way out — deliberately, so a caller who ignores the return value cannot transmit a partial record. Both paths are asserted.


alex:
One random observation: I think if AWS-LC had the `EVP_AEAD_CTX_sealv` API that BoringSSL had, rustls might be able to save a copy when encrypting.

justsmth:
Thanks for the issue and PR! The problem is real and well-motivated, and we agree it's worth solving.

We have some concerns to work through for the API: this PR introduces a couple of patterns that don't currently exist in our public API (use of `MaybeUninit` and `Vec` parameters). We want to be deliberate about it since they set new precedents. We'll take a closer look.

A couple of non-technical issue: the clippy and rustfmt CI jobs are failing, and the two new test files are missing the license headers our CI checks for (see [`scripts/tools/copyright_check.sh`](https://github.com/aws/aws-lc-rs/blob/main/scripts/tools/copyright_check.sh)).

justsmth:
Following up on the API patterns I mentioned above -- I looked at recent rustls PRs, and I think it reinforces my inclination.

rustls/rustls#3127 (resolving rustls/rustls#3068) reworks `MessageEncrypter::encrypt` to write into a caller-provided `&mut [u8]` fed from recycled record buffers. For that implementation, `seal_separate_out_of_place` provides what's needed -- the "recycled buffers" would be already-initialized. I think the performance win would still be achieved for rustls -- it just gets delivered through rustls's buffer reuse rather than from using `Vec` spare capacity. (The `_append` and `_uninit` seem more tailored to the current fresh-`Vec`-per-record contract, which that PR removes.)

My preference would be to narrow this PR to just `seal_separate_out_of_place`. It closes the gap on its own and mirrors `open_separate_gather`, which already uses a caller-provided `&mut [u8]` for the open direction. If rustls/rustls#3127 stalls or rustls lands on a `&mut Vec<u8>` parameter instead, we can revisit the other variants as a follow-up.

iainmcgin:
Trimmed down to just `seal_separate_out_of_place` on `LessSafeKey` and `TlsRecordSealingKey`. I do think there's probably still some justification for a `MaybeUninit` variant, less opinionated about the Vec-form (which is primarily convenience). Happy to carry anthing else forward, I agree that just `seal_separate_out_of_place` covers the main thing I need. I've been engaging with https://github.com/rustls/rustls/pull/3127 as well as I conducted and end to end test of this change propagated all the way up to hyper to confirm it was worth it.

aws-lc-rs/src/aead.rs
```diff
@@ -901,6 +901,56 @@ impl LessSafeKey {
...
+    /// error the output buffers never hold usable ciphertext: a length mismatch is
+    /// rejected before the AEAD runs and leaves them untouched, while a failure inside
+    /// the AEAD zeroes them, so a caller that ignores the result cannot transmit a
+    /// partial or stale record.
+    // # FIPS
+    // This method must not be used.
+    //
+    #[inline]
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn seal_separate_out_of_place<A>(
```
justsmth: Since this will permanently be part of our 1.x public API, could we instead name this `seal_out_of_place_scatter`? (Same for the `TlsRecordSealingKey` method.)

It's the "out-of-place" analogue of the existing `LessSafeKey::seal_in_place_scatter` -- same parameters, so the `in_place` -> `out_of_place` substitution seems right.

aws-lc-rs/src/aead.rs
```diff
@@ -901,6 +901,56 @@ impl LessSafeKey {
...
+        &self,
+        nonce: Nonce,
+        aad: Aad<A>,
+        plaintext: &[u8],
+        ciphertext: &mut [u8],
```
justsmth: Nitpick:  [`open_separate_gather`](https://github.com/aws/aws-lc-rs/blob/a298fef48e857884c4e731a019a112bfb300a03e/aws-lc-rs/src/aead.rs#L812-L819) uses directional prefixes (`in_ciphertext`, `in_tag`, `out_plaintext`). Since this is documented as its sealing counterpart, `in_plaintext` / `out_ciphertext` would mirror that. (And parameter names show up in rustdoc.)

aws-lc-rs/src/aead/unbound_key.rs
```diff
@@ -194,6 +194,66 @@ impl UnboundKey {
...
+            )
+        } {
+            return Err(Unspecified);
+        }
+        // The tag length the C function reports is the only evidence available that it
+        // filled the tag buffer, so a mismatch has to be an error rather than a shorter
+        // successful write.
+        if out_tag_len != extra_out_and_tag.len() {
+            return Err(Unspecified);
+        }
```
justsmth: This should not be possible with the current implementation. Could we make this a `debug_assert_eq!` instead?

aws-lc-rs/src/aead/unbound_key.rs
```diff
@@ -194,6 +194,66 @@ impl UnboundKey {
...
+    ///
+    /// `ciphertext` must be exactly `plaintext.len()` bytes. `extra_in` is additional
+    /// plaintext (TLS 1.3's inner content-type byte, for instance) that is encrypted and
+    /// written to `extra_out_and_tag` ahead of the tag, so `extra_out_and_tag` must be
+    /// `extra_in.len() + algorithm().tag_len()` bytes. The output buffers may not
+    /// overlap each other or `plaintext`.
+    ///
+    /// # Errors
+    /// `error::Unspecified` if the lengths are wrong or the encryption operation fails.
+    #[inline]
```
justsmth: NP: We don't typically doc-comment `pub(crate)` methods.

justsmth (commented):
> FIPS service indicator. The out-of-place path calls EVP_AEAD_CTX_seal_scatter without wrapping it in indicator_check!, which matches the existing seal_in_place_separate_scatter but differs from every other seal in unbound_key.rs. ...

We can leave this as-is for this PR. It keeps the scatter/gather family internally consistent.

iainmcgin:
All four applied, thanks! One note on the `debug_assert_eq!` - `out_tag_len` was initialised to `extra_out_and_tag.len()`, so it could only ever catch a wrong write and never a missing one. To catch that, it is now set to zero, which will never match `extra_in.len() + tag_len()`. I trimmed and rearranged the docs to match the neighbouring methods. The old `# Errors` said "on success every byte of both output buffers has been written", which is true, but the check you asked me to remove was the only thing enforcing it, so that can't be stated as a guarantee anymore.

justsmth (approved):

torben-hansen (approved):

PR #1184, opened 2026-07-20
Support exporting build vars from -sys crates to build depending C code

glebpom:
## Description of changes:

Currently, the sys crates export include directories and linker directives, but downstream Cargo build scripts cannot reliably determine the exact native library artifacts produced by AWS-LC. This is especially problematic when compiling additional C code against AWS-LC because source builds use version-prefixed library names.

This change exports additional build metadata from `aws-lc-sys` and `aws-lc-fips-sys`:

- `libdir`
- `libcrypto_path`
- `libssl_path` when the `ssl` feature is enabled
- `link_kind`

The metadata is emitted consistently for CC builds, CMake builds, and system-library builds. Platform-aware artifact naming is included for static and dynamic libraries on Unix, Apple platforms, MinGW, and MSVC.

This enables downstream build scripts to pass AWS-LC’s exact headers and libraries to native build systems such as CMake, without rebuilding AWS-LC or guessing version-prefixed artifact names.

## Call-outs:

The new metadata becomes an interface consumed by direct Cargo dependents through versioned `DEP_AWS_LC_*` and `DEP_AWS_LC_FIPS_*` environment variables.

Exact absolute artifact paths are exported intentionally because the existing logical library names are insufficient for native build systems that require a concrete library file.

Special review attention would be helpful for:

- Cross-platform library filename mapping.
- Static versus dynamic `link_kind` reporting.
- Consistency between CC, CMake, and system-library builders.
- Conditional export of `libssl_path` when the `ssl` feature is enabled.

Existing Rust linker directives and linking behavior are unchanged.

## Testing:

The `links-testing` crate now verifies that:

- The exported library directory exists.
- The exported `libcrypto` path exists.
- The artifact belongs to the exported library directory.
- The linkage kind is either `static` or `dylib`.

The metadata was tested with:

- `aws-lc-sys` using the CC builder.
- `aws-lc-sys` using the CMake builder.
- `aws-lc-fips-sys` using its CMake build.
- A downstream native librdkafka build consuming the exported AWS-LC headers and exact library paths.

### DIFF
--- a/builder/main.rs
+++ b/builder/main.rs
@@ -342,6 +342,19 @@ impl OutputLibType {
             Self::Dynamic => "dynamic",
         }
     }
+
+    fn library_filename(self, name: &str) -> String {
+        match (target_os().as_str(), target_env().as_str(), self) {
+            ("windows", "msvc", _) => format!("{name}.lib"),
+            ("windows", _, Self::Dynamic) => format!("lib{name}.dll.a"),
+            ("windows", _, Self::Static) => format!("lib{name}.a"),
+            ("macos" | "ios" | "tvos" | "watchos" | "visionos", _, Self::Dynamic) => {
+                format!("lib{name}.dylib")
+            }
+            (_, _, Self::Dynamic) => format!("lib{name}.so"),
+            (_, _, Self::Static) => format!("lib{name}.a"),
+        }
+    }
 }
 
 impl OutputLib {
@@ -1397,6 +1410,37 @@ pub(crate) fn emit_source_build_metadata(manifest_dir: &Path) {
     }
 
     println!("cargo:rerun-if-changed=aws-lc/");
+}
+
+/// Exports stable, absolute native-library locations for downstream build
+/// scripts that need to compile additional C code against this AWS-LC build.
+pub(crate) fn emit_source_library_metadata(
+    lib_dir: &Path,
+    output_lib_type: OutputLibType,
+    build_prefix: &Option<String>,
+) {
+    let crypto_path =
+        lib_dir.join(output_lib_type.library_filename(&OutputLib::Crypto.libname(build_prefix)));
+    assert!(
+        crypto_path.is_file(),
+        "AWS-LC libcrypto artifact not found: {}",
+        crypto_path.display()
+    );
+
+    println!("cargo:libdir={}", lib_dir.display());
+    println!("cargo:libcrypto_path={}", crypto_path.display());
+    println!("cargo:link_kind={}", output_lib_type.rust_lib_type());
+
+    if cfg!(feature = "ssl") {
+        let ssl_path =
+            lib_dir.join(output_lib_type.library_filename(&OutputLib::Ssl.libname(build_prefix)));
+        assert!(
+            ssl_path.is_file(),
+            "AWS-LC libssl artifact not found: {}",
+            ssl_path.display()
+        );
+        println!("cargo:libssl_path={}", ssl_path.display());
+    }
 }
 
 fn setup_include_paths(out_dir: &Path, manifest_dir: &Path) -> PathBuf {
--- a/builder/main.rs
+++ b/builder/main.rs
@@ -347,7 +347,6 @@ impl OutputLibType {
         match (target_os().as_str(), target_env().as_str(), self) {
             ("windows", "msvc", _) => format!("{name}.lib"),
             ("windows", _, Self::Dynamic) => format!("lib{name}.dll.a"),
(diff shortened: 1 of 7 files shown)


builder/main.rs
```diff
@@ -1399,6 +1411,37 @@ pub(crate) fn emit_source_build_metadata(manifest_dir: &Path) {
...
+    println!("cargo:link_kind={}", output_lib_type.rust_lib_type());
+
+    if cfg!(feature = "ssl") {
+        let ssl_path =
+            lib_dir.join(output_lib_type.library_filename(&OutputLib::Ssl.libname(build_prefix)));
+        assert!(
+            ssl_path.is_file(),
+            "AWS-LC libssl artifact not found: {}",
+            ssl_path.display()
+        );
```
justsmth, line 1440: Since asserts fail the build for all consumers (not just ones reading these vars), could these be downgraded to `cargo:warning`?

links-testing/build.rs
```diff
@@ -31,19 +31,37 @@ fn main() {
 }
 
 fn build_and_link(links: &str, target_name: &str) {
```
justsmth, line 33: Our CI job for "links-testing" should be extended to cover a few more build configurations:
  - Windows dynamic -- both `*-windows-gnu` and `*-windows-mscv`
  - ssl consumer coverage -- an ssl-enabled feature in `links-testing` that asserts `DEP_*_LIBSSL_PATH`

builder/main.rs
```diff
@@ -342,6 +342,18 @@ impl OutputLibType {
             Self::Dynamic => "dynamic",
         }
     }
+
+    fn library_filename(self, name: &str) -> String {
+        match (target_os().as_str(), target_env().as_str(), self) {
+            ("windows", "msvc", _) => format!("{name}.lib"),
+            ("windows", _, Self::Dynamic) => format!("lib{name}.dll.a"),
```
justsmth, line 349: I'm not sure this is true for our prefixed source builds. The sys crates' CMakeLists.txt only sets `IMPORT_PREFIX` for MSVC; on MinGW the import library keeps CMake's default `lib` prefix, so we'd get `libcrypto.dll.a` while this predicts `libaws_lc_X_Y_Z_crypto.dll.a`.
- With the `assert!` below, this might break `AWS_LC_SYS_STATIC=0` builds on windows-gnu, which currently link fine because ld falls back to the DLL directly.

builder/main.rs
```diff
@@ -1399,6 +1411,37 @@ pub(crate) fn emit_source_build_metadata(manifest_dir: &Path) {
...
+        crypto_path.is_file(),
+        "AWS-LC libcrypto artifact not found: {}",
+        crypto_path.display()
+    );
+
+    println!("cargo:libdir={}", lib_dir.display());
+    println!("cargo:libcrypto_path={}", crypto_path.display());
+    println!("cargo:link_kind={}", output_lib_type.rust_lib_type());
+
+    if cfg!(feature = "ssl") {
```
justsmth, line 1433: We need to add an ssl-enabled feature to `links-testing` that asserts `DEP_*_LIBSSL_PATH` the same way as libcrypto, plus a matrix entry to the `links-crate-tests` job.

justsmth:
I've stacked five follow-up commits addressing my review feedback on top of your branch here: https://github.com/justsmth/aws-lc-rs/tree/pr-1184-review-fixes

Since this PR comes from an organization-owned fork, I can't push to the branch directly (GitHub doesn't support maintainer edits for org forks).

My changes cover: 
- setting `IMPORT_PREFIX` for non-MSVC toolchains so MinGW shared builds produce the prefixed `lib*.dll.a` the metadata predicts; 
- downgrading the artifact asserts to warn-and-skip (links-testing now enforces presence in CI instead); 
- libssl_path validation + ssl feature coverage in links-testing; 
- the expanded `links-crate-tests` CI matrix (dynamic linkage, ssl, windows-gnu); and 
- a README note about link-time vs runtime artifacts on Windows.

You can bring them into this PR with:
```
git pull https://github.com/justsmth/aws-lc-rs pr-1184-review-fixes
git push
```

I'd also be fine with closing this PR and creating a separate one with these changes, if you'd prefer. 

NOTE: my changes grow the CI matrix and exercise some combinations for the first time (ssl on macOS/Windows, windows-gnu dynamic), so we may see new failures that I can help triage.

glebpom:
@justsmth Thanks for the review! Since you've already fixed the remaining issues - please proceed with merging your PR, I'm totally fine with that

PR #1187, opened 2026-07-24
Export native library build metadata from the -sys crates

justsmth:
### Issues:
Supersedes #1184 (includes @glebpom's original commits).

### Description of changes:
Downstream build scripts compiling C code against AWS-LC can't reliably locate our native library artifacts, since source builds use version-prefixed names. This exports `libdir`, `libcrypto_path`, `link_kind`, and (with `ssl`) `libssl_path` from the sys crates as `DEP_AWS_LC_*` / `DEP_AWS_LC_FIPS_*` env vars, consistently across the CC, CMake, and system-library build paths. Existing linker directives are unchanged.

Follow-up commits on top of #1184:

- Apply `IMPORT_PREFIX` for all toolchains (previously MSVC-only) so MinGW shared builds produce the version-prefixed import library the metadata predicts.
- Skip the `*_path` exports with a `cargo:warning` when a predicted artifact is missing, instead of `assert!`-failing every consumer's build; `links-testing` enforces presence in CI instead.
- Add ssl-enabled features to `links-testing` that assert `DEP_*_LIBSSL_PATH`.
- Extend the `links-crate-tests` CI matrix: dynamic linkage, ssl, and windows-gnu.

### Call-outs:
The new keys become a versioned public interface for direct dependents. Exact paths are exported intentionally: logical names are insufficient for native build systems that need a concrete library file. On Windows, `*_path` values are the link-time artifact (import library), not the runtime DLL.

The `links-crate-tests` matrix grows from 16 to 50 jobs and exercises some configurations for the first time (ssl on macOS/Windows, windows-gnu dynamic); if these surface unrelated failures, we can trim with excludes.

### Testing:
- `links-testing` verified locally on macOS for: static (CC), dynamic (CMake), ssl static/dynamic, FIPS ssl, and the `aws-lc-rs` forwarding path.
- The expanded CI matrix validates the exports across all platforms, both linkages, and ssl; the windows-gnu entries are the first CI coverage for MinGW dynamic builds.

```

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -128,11 +128,29 @@ jobs:
     if: github.repository_owner == 'aws'
     name: sys crate tests
     runs-on: ${{ matrix.os }}
+    env:
+      # 'default' leaves the linkage choice to the builder; 'dynamic' forces a
+      # shared-library build so the exported artifact-path metadata is
+      # validated for both link kinds.
+      AWS_LC_SYS_STATIC: ${{ (matrix.linkage == 'dynamic' && '0') || '' }}
+      AWS_LC_FIPS_SYS_STATIC: ${{ (matrix.linkage == 'dynamic' && '0') || '' }}
     strategy:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-15-intel, macos-latest, windows-latest ]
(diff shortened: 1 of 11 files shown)


prasden (dismissed):

justsmth:
Submitted PR upstream to address the CI failure here: https://github.com/aws/aws-lc/pull/3398

dougch (approved):

samuel40791765 (approved):

PR #1189, opened 2026-07-24
fix: align PQDSA API with signature module conventions

justsmth:
### Issues:
Follow-up to the ML-DSA stabilization in #1188. No issue is being resolved.

### Description of changes:
A batch of consistency fixes bringing the PQDSA (ML-DSA) API in line with the rest of the `signature` module:

* Renames `PqdsaKeyPair::to_pkcs8` to `to_pkcs8v1`. Elsewhere in the module (e.g. `Ed25519KeyPair`) the version-unqualified name means PKCS#8 v2, so the old name was misleading. A deprecated `to_pkcs8` alias remains available under the `unstable` feature, so existing consumers get a deprecation warning rather than a compile error.

* `PqdsaVerificationAlgorithm::parsed_verify_digest_sig` now always returns `Unspecified`. Digest-then-verify is not an operation defined by FIPS 204: pure ML-DSA signs the message itself, and the pre-hash variant (HashML-DSA) uses a distinct domain separator that this API does not implement. Previously this path would happily verify the digest bytes as if they were a pure ML-DSA message. This aligns the parsed path with the unparsed path and with Ed25519.

* Declares the `signature::parsed_public_key_tests` module. The file was added in #863 but the `mod` declaration was never added, so its 27 tests have silently not been compiling or running.

* Smaller additions: `Debug` for `PqdsaPrivateKey`, `public_key_len`/`seed_len` accessors on `PqdsaSigningAlgorithm`, FIPS annotations on keygen/sign/verify, an ML-DSA subsection in the signature module's Algorithm Details docs, and assorted doc fixes.

### Call-outs:
* The `parsed_verify_digest_sig` change is behavioral: anything that was verifying an externally-computed digest against an ML-DSA signature via `ParsedPublicKey` will now fail. That prior behavior was arguably a bug (it accepted a construction FIPS 204 doesn't define).

* Since `parsed_public_key_tests` hasn't been compiling since #863, this PR is effectively (re)introducing 27 tests.

### Testing:
Existing unit and integration tests, plus the newly-enabled `parsed_public_key_tests` module (27 tests). `pqdsa_test.rs` updated for the `to_pkcs8v1` rename.

### DIFF
--- a/aws-lc-rs/src/pqdsa/signature.rs
+++ b/aws-lc-rs/src/pqdsa/signature.rs
@@ -16,7 +16,7 @@ use core::fmt::{Debug, Formatter};
 #[cfg(feature = "ring-sig-verify")]
 use untrusted::Input;
 
-/// An PQDSA verification algorithm.
+/// A PQDSA verification algorithm.
 #[derive(Debug, Eq, PartialEq)]
 pub struct PqdsaVerificationAlgorithm {
     pub(crate) id: &'static AlgorithmID,
@@ -24,7 +24,7 @@ pub struct PqdsaVerificationAlgorithm {
 
 impl sealed::Sealed for PqdsaVerificationAlgorithm {}
 
-/// An PQDSA signing algorithm.
+/// A PQDSA signing algorithm.
 #[derive(Debug, Eq, PartialEq)]
 pub struct PqdsaSigningAlgorithm(pub(crate) &'static PqdsaVerificationAlgorithm);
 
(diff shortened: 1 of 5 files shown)


aws-lc-rs/src/pqdsa/signature.rs
```diff
@@ -69,17 +83,20 @@ impl ParsedVerificationAlgorithm for PqdsaVerificationAlgorithm {
...
-        public_key: &ParsedPublicKey,
-        digest: &Digest,
-        signature: &[u8],
+        _public_key: &ParsedPublicKey,
+        _digest: &Digest,
+        _signature: &[u8],
     ) -> Result<(), Unspecified> {
-        let evp_pkey = public_key.key();
-        evp_pkey.verify_digest_sig(digest, No_EVP_PKEY_CTX_consumer, signature)
+        // ML-DSA does not support digest-then-verify. Verifying a signature over
```
jakemas: This is not strictly true. ML-DSA does support digest-then-verify, it's just very poorly worded in FIPS 204. "External Mu" mode is a mode in which messages can be digested first, externally, by preparing `Mu` externally. There are lots of details here: https://csrc.nist.gov/csrc/media/Projects/post-quantum-cryptography/documents/faq/fips204-sec6-03192025.pdf

Of course, no need to litter this comment with all this info, but may want to tighten up language here.

jakemas (dismissed):
Reviewed and built/tested locally (all pass). Small nit on a comment.

samuel40791765 (dismissed):

jakemas (approved):
Perfect, thank you!

samuel40791765 (approved):

PR #1191, opened 2026-07-27
fix: zeroize KeyEncryptionKey key material on drop

justsmth:
### Description of changes:
`KeyEncryptionKey` stored the raw key-encryption-key bytes in a plain `Box<[u8]>` that was dropped without scrubbing, leaving key material behind in freed heap memory. Other key types in this crate (`SymmetricCipherKey`, `ChaCha20Key`, HKDF keys, owned `Buffer`s) zeroize their key material on drop; this adds a `Drop` impl to bring the `key_wrap` module in line with that practice.

A second commit fixes doc/comment nits in the same module: the comment describing `AES_unwrap_key`'s input validation had both bounds wrong (AWS-LC accepts `in_len == 24` and `in_len == INT_MAX`), plus a handful of spelling errors in doc comments.

### Call-outs:
No functional or API changes -- the `Drop` impl only affects what's left in memory after the key is dropped.

### DIFF
--- a/aws-lc-rs/src/key_wrap.rs
+++ b/aws-lc-rs/src/key_wrap.rs
@@ -109,7 +109,7 @@ pub const AES_256: AesBlockCipher = AesBlockCipher {
 /// A Key Wrap (KW) algorithm implementation.
 #[allow(clippy::module_name_repetitions)]
 pub trait KeyWrap: Sealed {
-    /// Peforms the key wrap encryption algorithm using a block cipher.
+    /// Performs the key wrap encryption algorithm using a block cipher.
     /// It wraps `plaintext` and writes the corresponding ciphertext to `output`.
     ///
     /// # Errors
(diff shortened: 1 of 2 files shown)


prasden (approved):

aws-lc-rs/src/key_wrap.rs
```diff
@@ -408,6 +409,12 @@ impl KeyWrapPadded for KeyEncryptionKey<AesBlockCipher> {
     }
 }
 
+impl<Cipher: BlockCipher> Drop for KeyEncryptionKey<Cipher> {
```
sgmenda, line 412: I wonder if there's a cleaner way to do this, where all key objects inherit from `Key` that implements `Drop` like this?

sgmenda (approved):
nit: the commit message is misleading, it should also mention the typo fixes. 

lgtm

Issue #1193, opened 2026-07-28
aws-lc-sys build script fails to copy bindings from read only prebuild aws-lc on repeat runs


RobinMarchart:
### Problem:

When using a prebuild aws-lc, the build script at aws-lc-sys/builder/system_library.rs:345 copies the pregenerated bindings to the out dir. Because this is done through std::fs::copy, the new file has the same permissions as the original file. If the original file was marked as read only, this results in a failed build on the next run because it lacks permissions to modify the file. The build script should probably adjust file permissions to make the copied file writable.

#### Relevant details
version 0.42.0
aws-rs build from nixpkgs (and therefore with all files permissions set to read only)

System information:
 * CPU architecture: x86-64
 * OS: NixOS (Unstable), Linux 7.1.4

Build log:
```text
--- stderr

  thread 'main' (359765) panicked at ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/aws-lc-sys-0.42.0/builder/main.rs:1338:21:
  called `Result::unwrap()` on an `Err` value: "Failed to copy bindings from /nix/store/mnn8nxzpksw3nxb0249jfhr4a2b1nw72-aws-lc-5.0.0-dev/share/rust/aws_lc_bindings.rs: Permission denied (os error 13)"
```

justsmth:
Hello! [aws-lc-rs v1.18.0](https://github.com/aws/aws-lc-rs/releases/tag/v1.18.0) and aws-lc-sys v0.44.0 are now released with the fix for this (#1194). Copies into `OUT_DIR` now clear any stale destination first and leave the copy writable -- your diagnosis of `fs::copy` propagating permission bits was exactly right. Two other copies had the same problem (include headers, prebuilt NASM objects) and are fixed as well.

This issue will now be closed. Feel free to comment here or open a new issue if you have any problems after upgrading. Thanks for reporting, and for helping to improve our library! 😃

PR #1194, opened 2026-07-28
Fix bindings copy from read-only prebuilt AWS-LC install

justsmth:
### Issues:

Addresses: #1193

### Description of changes:

When linking against a prebuilt AWS-LC via `AWS_LC_SYS_SYSTEM_DIR`, the build script copies the pre-generated bindings into `OUT_DIR` with `fs::copy`, which propagates the source's permission bits. If the install is read-only (e.g. the Nix store), the copied `bindings.rs` is too, and the next build-script rerun fails with `Permission denied` trying to overwrite it.

Two other copies into `OUT_DIR` hit the same problem when the crate source is read-only, so all three now share a `copy_writable` helper that removes any stale destination before copying and leaves the fresh copy writable:

* `system_library.rs` -- the pre-generated bindings (the reported bug).
* `setup_include_paths` -- headers copied out of the crate source.
* `nasm_builder.rs` -- prebuilt NASM objects copied out of the crate source.

Removing first means an `OUT_DIR` already poisoned by an earlier build recovers without a `cargo clean`. Leaving the copy writable matters because other writers of these same paths don't clear read-only themselves -- `nasm -o` fails with `unable to open output file`, and `File::create` in the internal-bindgen path fails the same way.

### Call-outs:

* The error in the issue names the _source_ path, but the failing open is on the destination -- easy to misread when debugging.
* The include-header copy can fail within a single build, not just on rerun: unlike the sibling directory copy it has no `skip_exist`, so a header present in two include paths gets overwritten immediately.
* The read-only handling is asymmetric on purpose. `remove_file` rejects a read-only file on Windows through at least Rust 1.85 (std fixed it by 1.90, but our MSRV is 1.71), so the attribute is cleared first under `cfg(windows)`. On Unix removal depends on the parent directory's write bit, and `set_readonly(false)` there would widen the mode to `0o666` -- hence `mode | 0o200` instead.
* New `clippy-builder-windows` job. Clippy only runs on Linux today, so every `#[cfg(windows)]` block in the builder -- including the hand-rolled `GetShortPathNameW` FFI -- is cfg-stripped before lints run and has never been linted. Linting `-p builder-test --lib` covers it in a few seconds without building AWS-LC. It surfaced 3 pre-existing findings in `to_short_path`, fixed here. Scoped to `--lib` because `--all-targets` pulls in the builder's test modules, which have their own pre-existing findings; wiring those in is worth a follow-up.

### Testing:

New `test-readonly-install-{unix,windows}` CI jobs build `aws-lc-sys` twice against a read-only install, forcing a build-script rerun in between via the existing `rerun-if-changed=builder/` directive. The Windows job exercises the attribute-clearing branch, which is `cfg`'d out on Unix.

Verified locally that a source build against read-only include directories succeeds, succeeds again across a forced rerun, and recovers an `OUT_DIR` left read-only by a pre-fix build. Verified the Windows-specific behaviors on a Windows Server 2022 host across Rust 1.71/1.77/1.85/1.90: `fs::copy` propagates the read-only attribute, overwriting a read-only file fails, and `remove_file` behaves as described above.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -62,6 +62,22 @@ jobs:
         working-directory: ${{ matrix.crate_dir }}
         run: cargo clippy ${{ matrix.features }} --all-targets -- -W clippy::all -W clippy::pedantic -D warnings
 
+  # The clippy job above runs on Linux, so every #[cfg(windows)] block in the
+  # builder is cfg-stripped before lints run. builder-test compiles the builder
+  # as a plain lib, so this lints that code without building AWS-LC. `--lib`
+  # skips the builder's test modules, which have pre-existing findings.
+  clippy-builder-windows:
+    if: github.repository_owner == 'aws'
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v7
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install ${{ env.RUST_NIGHTLY_TOOLCHAIN }} --profile minimal --component clippy --no-self-update
+          rustup default ${{ env.RUST_NIGHTLY_TOOLCHAIN }}
+      - name: Run cargo clippy (builder)
+        run: cargo clippy -p builder-test --lib -- -W clippy::all -W clippy::pedantic -D warnings
+
   apidiff:
     if: github.repository_owner == 'aws' && github.event_name == 'pull_request'
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


prasden (dismissed):
Should the CI tests also be added to the FIPS tests in https://github.com/aws/aws-lc-rs/blob/main/.github/workflows/system-lib-tests-fips.yml

prasden (approved):

nhatnghiho (approved):

Issue #1205, opened 2026-08-09
`CFLAGS_<target>` also should be filtered


leleliu008:
aws-lc-sys version: 0.43.0

Although it usually can not be set via shell code like:

```bash
export CFLAGS_aarch64-apple-darwin=-O3
```

But it can be set via C code:

```C
setenv("CFLAGS_aarch64-apple-darwin", "-O3", 1);
```

related code : https://github.com/aws/aws-lc-rs/blob/main/builder/cc_builder.rs#L583-L601

reference: https://docs.rs/cc/latest/cc/#external-configuration-via-environment-variables

justsmth:
Thanks for the report -- you're right. Our jitterentropy guard filtered `CFLAGS_aarch64_apple_darwin`, `HOST_CFLAGS`, `TARGET_CFLAGS`, and `CFLAGS`, but not the raw-triple spelling `CFLAGS_aarch64-apple-darwin` -- so that one slipped through.

And I can reproduce this on `aarch64-apple-darwin` (`cc` appends env CFLAGS after the build script's own flags, so the trailing `-O3` wins):

```
"-pedantic" "-O0" "-fwrapv" "-Wconversion" ... "-D_DARWIN_C_SOURCE=1" "-O3" "-o" ... jitterentropy-base.o
jitterentropy-base.c:47:3: error: "The CPU Jitter random number generator must not be compiled with optimizations..."
```

The fix should be in our next release of aws-lc-sys.

justsmth:
We've now released [aws-lc-rs v1.18.1](https://github.com/aws/aws-lc-rs/releases/tag/v1.18.1) which has [the change](https://github.com/aws/aws-lc-rs/pull/1207) to address the concern above. 

This issue will now be closed. Feel free to comment here or open another issue if you continue to see the problem after updating. Thanks again for the report! 😃

PR #1207, opened 2026-08-10
fix(builder): filter the raw-triple CFLAGS spelling for jitterentropy

justsmth:
### Issues:
Addresses #1205

### Description of changes:
Filter every target-specific `CFLAGS` spelling read by `cc` before compiling jitterentropy so an inherited optimization flag cannot override its required `-O0`. Add hostile-environment CI coverage for the raw target-triple spelling.

### Call-outs:
`cc` accumulates flags from every matching `CFLAGS` variable rather than selecting one by precedence. Raw target-triple names cannot be exported by a POSIX shell, but they can be set by a parent process and inherited. Our supported `cc` range also spans both legacy normalization (`-` to `_`) and current normalization (`-` and `.` to `_`), so dotted targets require both spellings.

### Testing:
- `cargo test -p builder-test`
- `env 'CFLAGS_aarch64-apple-darwin=-O3' cargo test -p aws-lc-rs --all-targets`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -830,6 +830,9 @@ jobs:
             env_value: '-O2'
           - env_name: TARGET_CFLAGS
             env_value: '-O2'
+          # Not exportable by a POSIX shell, but inheritable from a parent process (setenv).
+          - env_name: CFLAGS_x86_64-unknown-linux-gnu
+            env_value: '-O3'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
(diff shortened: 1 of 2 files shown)


justsmth:
Fix for FreeBSD CI coming in this PR: https://github.com/aws/aws-lc-rs/pull/1213

WillChilds-Klein (dismissed):

geedo0 (dismissed):

geedo0 (approved):

WillChilds-Klein (approved):

PR #1212, opened 2026-08-14
Fix `-Wa,--debug-prefix-map` probe

### LINKED ISSUE #1211: aws-lc-sys failed to build with clang when CFLAGS includes -flto
`CFLAGS="-flto=thin" cargo build --release` fails to build with the error `clang-22: error: unsupported argument '--debug-prefix-map=/tmp/aws-lc-sys=' to option '-Wa,'`.

This is caused by `cc::Build.is_flag_supported()` returning a false positive because
clang doesn't emit unsupported argument to option '-Wa,' error when -flto is also specified.

justsmth:
### Issues:
Addresses: #1211

### Description of changes:
`-Wa,--debug-prefix-map` mirrors `-ffile-prefix-map` for `.S` sources, since GCC does not forward the C-side flag to `as`. We applied it whenever `cc::Build::is_flag_supported()` accepted it, relying on clang rejecting it -- but clang only diagnoses an unknown `-Wa,` value when a compile job reaches its integrated assembler. With `-flto` in `CFLAGS`, the C probe stops at bitcode and reports the flag as supported; the real `.S` sources do reach the assembler and fail the build.

The compiler family now decides: only GCC gets the flag. That loses nothing for clang, whose integrated assembler honors `-ffile-prefix-map` for `.S` inputs directly. The probe survives only to catch a GNU `as` predating the option (binutils 2.26), and now runs with user `CFLAGS` suppressed.

### Call-outs:
* Two mechanisms here each fix #1211 on their own: the family gate, and the probe no longer seeing `-flto` once `CFLAGS` is suppressed. That redundancy is deliberate -- the gate also covers clangs that enable LTO through default config files rather than `CFLAGS`.
* `cflags_ignore_guards` now also removes `TARGET_CFLAGS`: `cc` reads it instead of `HOST_CFLAGS` when cross-compiling and concatenates every variant that is set, and unlike the memcmp probe this one runs in cross builds.
* Not fixed here: CMake never seeds `CMAKE_ASM_FLAGS` with `-ffile-prefix-map`, so clang + CMake can still record build paths in `.S` DWARF. Predates this change, and is why the new path-stripping job is cc-builder-only. Can file separately.

### Testing:
* New `builder-tests` lane runs `cargo test -p builder-test` on Linux/macOS/Windows -- previously no CI job ran the builder's unit tests at all. The compiler-probe tests among them only execute on Linux, where the runners provide both clang and GCC.
* Unit: gating table test, a clang regression test that the flag is withheld under `-flto`, and a GCC companion test that the flag survives the same `CFLAGS`. The clang test pins end behavior (either mechanism above withholds the flag alone; it fails with both reverted), and the GCC test doubles as a suppression tripwire because GCC rejects `-flto=thin`.
* CI: `linux-clang-lto` reproduces the report end to end for both builders -- no lane combined clang, Linux, and a release opt level before, which is how this slipped through. `linux-clang-path-stripping` proves clang strips `.S` paths via `-ffile-prefix-map` alone, guarding against vacuous passes by requiring `.S` DWARF to be present.

### DIFF
--- a/builder/cc_builder.rs
+++ b/builder/cc_builder.rs
@@ -28,7 +28,7 @@ use crate::{
 use std::cell::Cell;
 use std::collections::HashMap;
 use std::env;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 #[non_exhaustive]
 #[derive(PartialEq, Eq)]
@@ -380,7 +380,7 @@ impl CcBuilder {
         }
 
         let mut cc_build = self.create_builder();
-        let (is_cl_like, build_options) = self.collect_universal_build_options(&cc_build, false);
+        let (_, build_options) = self.collect_universal_build_options(&cc_build, false);
         for option in build_options {
             option.apply_cc(&mut cc_build);
         }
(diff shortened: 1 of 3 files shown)


builder/cc_builder.rs
```diff
@@ -991,22 +980,65 @@ impl CcBuilder {
...
+
+    let flag = format!("-Wa,--debug-prefix-map={}=", manifest_dir.display());
+
+    let _cflags_guards = cflags_ignore_guards();
+    let probe_build = cc::Build::default();
+    // `try_get_compiler` so an unresolvable compiler cannot panic the build script.
+    let Ok(compiler) = probe_build.try_get_compiler() else {
+        return None;
+    };
+    if !compiler.is_like_gnu() {
```
dougch, line 1031: Claude flagged this as possible not doing what we expect (compiler wrappers/MSVC)- is `is_like_clang` more useful ?

dougch (approved):
minor question

builder/cc_builder.rs
justsmth: `is_like_gnu()` is here because this is specifically a GNU `as` option. Using `is_like_clang()` as the exclusion would also not exclude `clang-cl` (Microsoft's).

I do not think a code change is needed here.

justsmth:
Fix for FreeBSD CI coming in this PR: https://github.com/aws/aws-lc-rs/pull/1213

prasden (approved):

PR #461, opened 2024-07-17
ECDSA P-521 with SHA1, SHA224, SHA256, SHA384

skmcgrail:
### Issues:
Resolves #457

### Description of changes: 
Add support for ECDSA with P-521 curve with either SHA-1, SHA-224, SHA-256, and SHA-384. Note SHA-1 will only be allowed for signature verification only.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

PR #1214, opened 2026-08-22
Add ECDSA P-256 SHA-1 ASN.1 verification algorithm

assafvayner:
### Issues:
None open. Follows up on #461, which added SHA-1 ECDSA verification for P-521 (`ECDSA_P521_SHA1_ASN1`).

### Description of changes:
Adds two constants:

- `ECDSA_P256_SHA1_ASN1` — verification of ASN.1 DER-encoded ECDSA P-256 signatures over SHA-1, mirroring `ECDSA_P521_SHA1_ASN1`.
- `ECDSA_P256_SHA1_ASN1_SIGNING` — the corresponding signing algorithm, documented with a `# ⚠️ Warning` (backwards compatibility only).

No aws-lc / aws-lc-sys changes needed — `EVP_DigestSign`/`EVP_DigestVerify` already support SHA-1 with ECDSA.

**Motivation: Amazon CloudFront signed URLs.** With a [custom policy](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html), the policy statement is hashed with SHA-1 and signed with an RSA or ECDSA private key. [Supported key pairs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-trusted-signers.html) are RSA-2048 or ECDSA P-256. SHA-1 is the default; SHA-256 requires an explicit `Hash-Algorithm=SHA256` URL parameter. So P-256 + SHA-1 is needed both to issue CloudFront-compatible signed URLs and to verify them. Neither works with aws-lc-rs today, forcing pure-Rust fallbacks.

At Hugging Face we both sign and verify cloudfront style signed URLs in our production services.

In a benchmark with this new interface, aws-lc-rs is ~8.6× faster at signing and ~4.5× at verification than the `p256` crate for this operation (EC2 m6i, Ice Lake).

### Call-outs:
- In #461 review, SHA-1 signing constants were removed in favor of verification-only. This PR proposes signing again for P-256 because CloudFront's own signed-URL scheme requires producing these signatures. If verification-only is still preferred, I'm happy to split the PR.
- SHA-1 signing is not FIPS-approvable. I can gate the signing constant behind `#[cfg(not(feature = "fips"))]` if desired.

### Testing:
- New roundtrip test in `tests/ecdsa_tests.rs`: sign via `sign`/`sign_digest`, verify via `verify`/`verify_digest`, negative check under `ECDSA_P256_SHA256_ASN1`.
- Cross-verified against RustCrypto `p256` in both directions on x86_64 and aarch64.

This change was developed with AI assistance, and human-reviewed and tested before submission.


justsmth:
Thanks for the PR! 

Our CI requires contributions to include the following statement in the PR description:


If you agree, would you append this text to the description? (I'll then rerun the failing CI job for this.)

assafvayner:
@justsmth Thanks for unblocking the next step. I added the line to the PR description and fixed a formatting issue that the rustfmt step caught!

justsmth (dismissed):

justsmth:
I pushed a commit that adds several NIST CAVP test vectors, and fixed how our tests handle the FIPS service indicator when "verify" is approved, but not "sign".

WillChilds-Klein (dismissed):

justsmth:
> ... If verification-only is still preferred, I'm happy to split the PR.

Yeah, we'd like to narrow the scope of this for now to just the "verify" part. There's some concern that consumers might pickup the signing operation not realizing its deprecated. 

We can discuss signing further on another PR if you'd like.

justsmth (approved):

WillChilds-Klein (approved):

assafvayner:
Hi, thank you @justsmth and @WillChilds-Klein for the quick actions on this PR and for merging verification for now!

> At Hugging Face we both sign and verify Cloudfront style signed URLs in our production services.

For context, since we sign urls with ECDSA and SHA-1 for both Cloudfront and another service we own (we operate a service to replace our Cloudfront usage, and follow the same signature process for auth there). We only verify these signatures in production on our service, (Cloudfront verifies the signature for Cloudfront presigned urls).

So we also need signing capabilities, and we also get more speedup benefit for switching the signing implementation to AWS-LC. (~8.6× faster at signing and ~4.5× at verification than the p256 crate for this operation).

I understand your concerns for not including signing currently. Before I open another issue or new PR, I will explore if we can eventually switch to sha-256 instead of sha-1.

And again, thank you!

justsmth:
> Before I open another issue or new PR, I will explore if we can eventually switch to sha-256 instead of sha-1.

Thanks. If needed, I'm fine with adding the signing operation. Maybe we could append a `_FOR_LEGACY_USE_ONLY` to the constant to make it stand out?

PR #1228, opened 2026-09-04
feat: implement `Eq` for `RsaParameters`

dsykes16:
Implement `Eq` and `PartialEq` for `aws_lc_rs::signature::RsaParameters` to permit discerning `ParsedPublicKey` algorithm. These traits are already implemented on `EcdsaVerificationAlgorithm` and are not required on `EdDSAParamters` as `Ed448` is not supported (only `Ed25519`).

### Issues:
N/A

### Description of changes: 
It currently appears to be impossible to discern the specific variant of RSA via the `algorithm` method on a `ParsedPublicKey` without resorting to its `Debug` trait. Prior to 0c88590, it was possible to use `std::ptr::addr_eq`. While it is technically possible to use a small buffer on the stack to hold the `Debug` identifier, that feels like an anti-pattern and no less brittle or subject to breakage as `ptr::addr_eq`. This change permits discerning the algorithm of a `ParsedPublicKey` in a reliable and straightforward way, and provides tests to ensure there are no unexpected breaking changes. The real world application for this is in [OxiToken](https://github.com/dsykes16/oxiauth/blob/main/oxitoken/src/crypto/aws_lc/verify.rs) (JWT library), which was broken by the changes in 0c88590.

### Call-outs:
This only implements traits on existing types; no breaking changes. 

`Eq` and `PartialEq` are already implemented on `EcdsaVerificationAlgorithm` here https://github.com/aws/aws-lc-rs/blob/main/aws-lc-rs/src/ec/signature.rs#L30

### Testing:
Unit tests

### DIFF
--- a/aws-lc-rs/src/ed25519.rs
+++ b/aws-lc-rs/src/ed25519.rs
@@ -33,7 +33,7 @@ const ED25519_SIGNATURE_LEN: usize = crate::aws_lc::ED25519_SIGNATURE_LEN as usi
 const ED25519_SEED_LEN: usize = 32;
 
 /// Parameters for `EdDSA` signing and verification.
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 pub struct EdDSAParameters;
 
 impl sealed::Sealed for EdDSAParameters {}
--- a/aws-lc-rs/src/rsa/signature.rs
+++ b/aws-lc-rs/src/rsa/signature.rs
@@ -22,13 +22,18 @@ use untrusted::Input;
 
 #[allow(non_camel_case_types)]
 #[allow(clippy::module_name_repetitions)]
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 pub enum RsaPadding {
     RSA_PKCS1_PADDING,
     RSA_PKCS1_PSS_PADDING,
 }
 
 /// Parameters for RSA verification.
+///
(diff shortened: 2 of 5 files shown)


justsmth:
Thanks for the PR! I pushed a follow-up with a few small improvements:
* derives `Eq`/`PartialEq` on `EdDSAParameters` as well, so every `VerificationAlgorithm` implementor is comparable
* switches `RsaParameters` from identifier-only `PartialEq` to a structural derive (plus a short doc comment), so equality no longer depends on each constant carrying a unique `RsaVerificationAlgorithmId`
* adds `compile_time_assert_eq` for the ECDSA types and folds the Ed25519 check into the existing traits test

dsykes16:
Thanks, all the changes in your commit look good to me. I wasn't sure if an `Eq/PartialEq` derive on `EdDSAParameters` would be accepted since `Ed25519` is the only algo supported right now as far as I know, so I didn't really have a use-case I could point to there, but I appreciate all the `VerificationAlgorithm` implementors now being comparable and it's future-proofed for if `Ed448` is ever added.

I went the custom `impl PartialEq for RsaParameters` route since it satisfied comparisons for all possible variants that are publicly exposed with as little overhead as possible, but I doubt a couple extra comparisons have a measurable performance impact and a simple `derive` is definitely cleaner and more future proof, so I certainly have no objections to that change.

justsmth (approved):

prasden (approved):

PR #1238, opened 2026-09-22
fix(builder): use unresolved CARGO_MANIFEST_DIR for source paths

justsmth:
### Context and motivation

When `CARGO_MANIFEST_DIR` goes through a symlink, the cc-builder produces a static library whose archive member names depend on the absolute build path, even when every object is byte-identical. In environments that build behind a per-build symlink, the archive digest changes on every build. That defeats remote compile caches (e.g. sccache), and the `aws-lc-sys` miss cascades through every dependent crate.

### Description of changes

cc-rs names each object `<hash(dirname)>-<basename>.o`. Before hashing, it string-strips the raw `CARGO_MANIFEST_DIR` from the dirname. We were building source paths from the symlink-resolved (canonicalized) manifest dir, so the strip silently missed and the absolute path was hashed instead.

On non-Windows hosts, the builder now uses `CARGO_MANIFEST_DIR` as cargo reports it. Windows still canonicalizes and uses 8.3 short paths for `MAX_PATH`.

### Testing

- A new `symlinked-manifest-dir-reproducibility` job in `tests.yml` builds `aws-lc-sys` twice: once from the checkout and once through a symlink to it via `--manifest-path`. It requires identical `ar t` member lists and no leaked build paths. Run locally, it fails on `main` and passes with this fix.
- Manual check before the fix: the two builds produced different member names (`0d86c3eef58cbbbf-asn1_lib.o` vs `bc1b343226bc6a17-asn1_lib.o`) while the object bytes were identical.
- The CMake builder through the symlink shows no leaked paths.
- `cargo test -p builder-test` passes.

### Review considerations

- The canonicalization dates to #85, and the history gives no stated reason for it.
- The CI job uses `--manifest-path` rather than `cd`-ing into the symlink. Cargo derives `CARGO_MANIFEST_DIR` from the resolved working directory, so the `cd` version would always pass.
- Windows is unchanged. Member names there can still depend on the build path, because the 8.3 short path differs from `CARGO_MANIFEST_DIR`. That's out of scope here.


.github/workflows/tests.yml
```diff
@@ -1157,6 +1157,45 @@ jobs:
...
+        # CARGO_MANIFEST_DIR from the resolved cwd, which would make the check vacuous.
+        run: |
+          ln -s "${PWD}" /tmp/aws-lc-rs-link
+          cargo build -p aws-lc-sys --release --target-dir /tmp/target-real
+          cargo build -p aws-lc-sys --release --target-dir /tmp/target-link \
+            --manifest-path /tmp/aws-lc-rs-link/Cargo.toml
+      - name: Verify archive member names match
+        run: |
+          REAL=$(find /tmp/target-real/release -type f -name "libaws_lc_*_crypto.a")
+          LINK=$(find /tmp/target-link/release -type f -name "libaws_lc_*_crypto.a")
```
prasden, line 1197: Should we check if they exist first? 

```
REAL=$(find /tmp/target-real/release -type f -name "libaws_lc_*_crypto.a") && test -f "$REAL" || { echo "FAIL: real archive not found"; exit 1; }
LINK=$(find /tmp/target-link/release -type f -name "libaws_lc_*_crypto.a") && test -f "$LINK" || { echo "FAIL: link archive not found"; exit 1; }
```

prasden (dismissed):

.github/workflows/tests.yml
justsmth: Good catch -- with an empty find result, both calls fail identically, so diff compares two empty outputs and passes, and strings "" matches nothing. Added a guard requiring exactly one archive per target dir (same pattern as `linux-clang-path-stripping`) in d1e08ab.

prasden (approved):

PR #1239, opened 2026-09-22
Add wasip2/p3 support

ricochet:
Add support for wasm32-wasip2 and wasm32-wasip3 targets.

- Add pre-generated bindings (these are compatible for both wasip2 and wasip3).
- Add cross-builds in CI using wasi-sdk 34.
- Test in CI with wasmtime.
- Update docs for new toolchain.

## Testing

- Linked the aws-lc-rs digest example for WASIp3 and ran it under Wasmtime.
- Ran the aws-lc-sys WASIp3 sanity tests under Wasmtime


justsmth:
Thanks for the PR!

As for bindings -- now that we have "[universal bindings](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-sys/src/universal_crypto.rs)", we're not interested in adding any more platform-specific ones.

I'll try to get this reviewed soon.

